PR #138, opened 2015-01-08
Make finder#show more agnostic

tommyp:
As we move Finders to being published by multiple Apps we want to make the Frontend more agnostic to what information it receives. This commit does that by specifying that if a value isn't returned in the response it is false. This happens for the beta flag, beta message and email signup flag.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -1,6 +1,6 @@
 class FinderPresenter
 
-  attr_reader :name, :slug, :document_noun, :document_type, :organisations, :keywords, :beta_message
+  attr_reader :name, :slug, :document_noun, :document_type, :organisations, :keywords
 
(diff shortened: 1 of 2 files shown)


jamiecobbett:
I'll leave it to someone familiar with the app to review, but :+1:

app/presenters/finder_presenter.rb
```diff
@@ -11,15 +11,18 @@ def initialize(content_item, values = {}, keywords = nil)
     @organisations = content_item.links.organisations
     facets.values = values
     @keywords = keywords
-    @beta_message = content_item.details.beta_message
   end
 
   def beta?
-    content_item.details.beta
+    content_item.details.beta || false
```
tommyp: Nothing, but I thought it would be better to have `beta?` to explicitly return false than `nil`.

tommyp:
Closing this because it seems these changes were not needed.

PR #142, opened 2015-01-13
Contacts finder

tommyp:
This PR contains the first load of work to render a Contacts Finder. The PR contains the following:

- **Accept routes not at root**: Other Finders will live at other URLs apart from root. This commit adds
a catchall glob and moves it after any that match email-signup.

- **Add Contact Document format**: Add a Contact document type. The only metadata this format has is ContactGroups which is labelled as Topics. These will be fetched from Rummager (when they are in there).

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -1,0 +1,13 @@
+class Contact < AbstractDocument
+  def self.tag_metadata_keys
+    %w(
+      contact_group
+    )
+  end
+
+  def self.metadata_name_mappings
+    {
+      "contact_group" => "Topic",
+    }
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,8 +1,8 @@
 FinderFrontend::Application.routes.draw do
   mount JasmineRails::Engine => '/specs' if defined?(JasmineRails)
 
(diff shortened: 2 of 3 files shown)


app/models/contact.rb
```diff
@@ -0,0 +1,13 @@
+class Contact < AbstractDocument
+  def self.tag_metadata_keys
+    %w(
+      contact_group
+    )
+  end
+
+  def self.metadata_name_mappings
+    {
+      "contact_group" => "Topic",
```
rboulton: Do we really want to overload the word "Topic" even further?  I ... don't.

tommyp: I don't either, but this is what it's called in the existing Contacts App index page. I'm open to suggestions though. @jamiecobbett?

bradleywright: [Here's an example of real usage](https://www.gov.uk/government/organisations/hm-revenue-customs/contact) - topic is the "area of business you want to contact the org about" (or thereabouts). So topic is a correct word here, if an overloaded one. We'd need an equivalent phrase.

evilstreak: I'm going to merge this – we can revisit the language later.

PR #144, opened 2015-01-16
Add validation for email signup checkboxes

james:
Provides an error to the user if there are checkboxes and they select
none of them on email signup

Testable by going to http://finder-frontend.dev.gov.uk/aaib-reports/email-signup and hitting submit with no checkboxes checked.

screenshot 2015-01-16 16 14 25

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -7,17 +7,31 @@ class EmailAlertSubscriptionsController < ApplicationController
 
   def new
     if content
-      @signup = SignupPresenter.new(content)
+      @signup = signup_presenter
     else
       error_not_found
     end
   end
 
   def create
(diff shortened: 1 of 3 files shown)


app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -7,18 +7,36 @@ class EmailAlertSubscriptionsController < ApplicationController
...
+      @signup = signup_presenter
+      @error_message = "Please choose an email alert"
+      render action: :new
+    end
   end
 
 private
 
+  def valid_choices?
+    if available_choices.present? && chosen_options.blank?
```
evilstreak: Would this method body be better as `available_choices.blank? || chosen_options.present?`

evilstreak:
:+1:

james:
@evilstreak good idea. Done.

PR #148, opened 2015-01-21
Use facets for metadata

tommyp:
This PR contains the work necessary to reduce specificity in FinderFrontend. To do this we need to remove each Document's Model and use the Finder to figure out what fields we are asking to be returned from Rummager and what metadata values should we show for each document. The way I've done this is adding the concept of a `FilterableFacet` and changing `Facet` to represent information about a set of metadata key/value pairs and if this should be filterable we then suggest a type of `{multi-select|date}` and it becomes a `FilterableFacet`.

The commits roughly go as follows:
- Use the Finder's facets to tell Rummager what values we want returned
- Remove document Models and Parser
- Refactor the facets to become `Facet` and `FilterableFacet` and expose `Finder#metadata` and `Finder#filters`
- Allow metadata labels for when the name of the facet isn't what we want to label the metadata with
- Set a summaries flag on the Finder to decide if we should show them
- Refactor `AbstractDocument` to `Document` and use the facets to convert metadata values to labels
- Pass the `FinderPresenter` object down to the `Document`
- Cleanup the features by updating the fixture content item with the metadata flags and the `closed_date` facet

After some discussion with @evilstreak I took a bit of a different approach and added extra commits which do the following:

- Set if the facet should be filterable and if it should be displayed as result metadata by having every type of facet be a `Facet` class or one of it's child classes.
- Remove filterable being hardcoded into the Filterable Facet class, but we still use this class so we can set the values of the facets.
- `FacetParser` now uses this new way of initialising the Facets.
- Use these new type of facets to specify tag and date metadata for the document
- Add the updated schema attrs to the CMA Cases content item fixture.

The additional commits were made to handle edge cases in:

1. [International Development Funding](http://gov.uk/international-development-funding) where we have metadata which is filterable but not displayed under the results.
2. [AAIB Reports](https://www.gov.uk/aaib-reports) which has free text fields which aren't filterable, though these are only displayed in SpecialistFrontend.


app/presenters/finder_presenter.rb
```diff
@@ -64,8 +76,7 @@ def related
 
   def results
     @results ||= ResultSet.get(
-      slug,
-      document_type,
+      self,
```
KushalP: I'm really unsure about this. It makes the owning object unclear.

My gut says to start by removing calls to the `some_finder_presenter.results` method here with direct uses of `ResultSet`.

tommyp:
Hold off merging this while I do some investigation. Think I've found a small issue with DSU.

tommyp:
This is good to go now. I noticed some weirdness with the metadata labels not being changed to `short_name` so I fixed that and cleaned up one of the commits which was changed later.

evilstreak:
Review still in progress! TBC

app/models/document.rb
```diff
@@ -33,18 +19,38 @@ def path
...
+
+  def tag_metadata_keys
+    finder.metadata.select{ |f| f.type == "text" }.map(&:key)
+  end
+
+  def metadata_keys
+    tag_metadata_keys + date_metadata_keys
+  end
+
+  def metadata_name_mappings
```
tommyp: Actually, we only want this to use facets which have `short_name`s, but I get what you're saying.

app/models/document.rb
```diff
@@ -33,18 +19,34 @@ def path
   end
 
   def summary
-    nil
+    description = attrs.fetch(:description, nil)
+
+    # This truncates the description at the end of the first sentence
+    description = description.gsub(/\.\s[A-Z].*/, '.') if description.present? && finder.show_summaries?
+  end
```
evilstreak: What about something like this to break it down a bit?

```ruby
def summary
  if finder.show_summaries? && description.present?
    description.gsub(/\.\s[A-Z].*/, '.')
  end
end

private # or in the initializer
def description
  attrs.fetch(:description, nil)
end
```

app/models/document.rb
```diff
@@ -1,27 +1,14 @@
...
-    tag_metadata_keys + date_metadata_keys
-  end
-
-  def self.metadata_name_mappings
-    {}
-  end
-
-  def initialize(attrs)
+  def initialize(attrs, finder)
+    attrs = attrs.with_indifferent_access
```
evilstreak: What prompted the change to turn this into a `HashWithIndifferentAccess`?

app/models/document.rb
```diff
@@ -91,19 +89,11 @@ def metadata_value_present?(metadata_hash)
...
-  end
-
-  def truncated_summary_or_nil
-    description = attrs.fetch(:description, nil)
-
-    # This truncates the description at the end of the first sentence
-    description = description.gsub(/\.\s[A-Z].*/, '.') if description.present?
-    description
+  def metadata_name_mappings
+    finder.metadata.each_with_object({}) { |facet, hash| hash[facet.key] = finder.label_for_metadata_key(facet.key) }
```
evilstreak: What does producing this lookup table give us that we can't get by just calling the finder directly? Ie:

```ruby
  def humanize_metadata_name(metadata_hash)
    metadata_hash.merge(
      name: finder.label_for_metadata_key(metadata_hash.fetch(:name))
    )
  end
```

app/models/result_set.rb
```diff
@@ -1,8 +1,8 @@
 class ResultSet
   attr_reader :documents, :total
 
-  def self.get(document_type, params)
-    ResultSetParser.parse(FinderFrontend.get_documents(document_type, params))
+  def self.get(finder, params)
+    ResultSetParser.parse(FinderFrontend.get_documents(finder, params), finder)
```
evilstreak: Given that you're passing `finder` into both calls, it might be clearer to break this over multiple lines:

```ruby
ResultSetParser.parse(
  FinderFrontend.get_documents(finder, params),
  finder,
)
```

app/models/document.rb
tommyp: All the test data is JSON, but the hash returned in the real world is symbols so all these setters in the initialize method fail in the tests. Should I change all the data?

app/parsers/facet_parser.rb
```diff
@@ -1,12 +1,16 @@
...
+      when 'text'
+        SelectFacet.new(facet)
+      when 'date'
+        DateFacet.new(facet)
+      else
+        raise ArgumentError.new("Unknown filterable facet type: #{facet.type}")
+      end
     else
-      raise ArgumentError.new("Unknown facet type: #{facet.type}")
+      Facet.new(facet)
```
evilstreak: Does it matter than non-filterable facets will just be an instance of `Facet`? Is there a difference in behaviour between a text facet and a date facet when it comes to displaying the metadata on results?

lib/finder_frontend.rb
```diff
@@ -1,14 +1,15 @@
...
-  def self.get_documents(document_type, params)
-    FindDocuments.new(document_type, params).call
+  def self.get_documents(finder, params)
+    FindDocuments.new(finder, params).call
   end
 
   class FindDocuments
-    def initialize(document_type, params)
+    def initialize(finder, params)
       # TODO Get `document_type` from `schema`
```
evilstreak: Is this `TODO` satisfied now?

app/models/document.rb
evilstreak: We should make the test data reflect the real world data – otherwise someone could remove this line and replace all the `fetch` arguments with strings, and the tests would indicate everything is fine

evilstreak:
Sorry for adding another round of comments. This really is good work.

I'll come over and work with you to get it merged today.

lib/finder_frontend.rb
tommyp: It is and I definitely removed that. Must've crept back in with a rebase.

app/models/document.rb
tommyp: For reference for anyone else this was discussed offline. This line was originally in https://github.com/alphagov/finder-frontend/blob/master/app/parsers/document_parser.rb#L3 but because we don't have a `DocumentParser` anymore I moved this line to the initialize method of `Document`.

evilstreak:
:fireworks:

tommyp:
🎉🎉

PR #149, opened 2015-01-22
Use "or" for two options

tommyp:
When selecting 2 options, the search description was "a and b". When selecting 3 options, it was "a, b, or c".

This commit adds the two_word_connector option to the fragment_values_to_s method to correctly make the 2 option description "a or b".


james:
urgh, that's not a very nice API.

tommyp:
Yeah. Who knew `and` in "this and that" is different from `and` in "this, this and that".

james:
It's to handle the oxford comma

tommyp:
Oh yeah, cause I suppose that then "this, this, and that". Does that mean it should be "this, this, or that"?

PR #143, opened 2015-01-16
Minor layout pedantry

futurefabric:
A few little style changes to even out spacing on checkbox labels and above the result set.

PR #150, opened 2015-01-22
Result count spacing

futurefabric:
A couple of tiny presentational changes:

* Remove margin at top of result count so it's positioned in the same way an unfiltered result set is.
* Bring result count and type closer together.

Like this:

finder

I should really have included them in this [previously merged PR](https://github.com/alphagov/finder-frontend/pull/143) but hey ¯\_(ツ)_/¯

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -397,7 +397,7 @@
       @include core-16;
       vertical-align: baseline;
       border-bottom: solid 1px $border-colour;
-      margin: $gutter-half 0;
+      margin: 0 0 $gutter-half 0;
       padding-bottom: $gutter-half;
 
       .result-count {
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -402,7 +402,7 @@
 
       .result-count {
         @include bold-48;
-        margin-right: $gutter-one-third;
+        margin-right: $gutter-one-third / 2;
       }
     }
 


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -397,12 +397,12 @@
...
       vertical-align: baseline;
       border-bottom: solid 1px $border-colour;
-      margin: $gutter-half 0;
+      margin: 0 0 $gutter-half 0 ;
       padding-bottom: $gutter-half;
 
       .result-count {
         @include bold-48;
-        margin-right: $gutter-one-third;
+        margin-right: $gutter-one-third / 2;
```
evilstreak: Would it be better to use `$gutter-one-sixth` here?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -397,12 +397,12 @@
       @include core-16;
       vertical-align: baseline;
       border-bottom: solid 1px $border-colour;
-      margin: $gutter-half 0;
+      margin: 0 0 $gutter-half 0 ;
```
tommyp: Extra space after the last 0 here.

app/assets/stylesheets/finder_frontend.scss
futurefabric: $gutter-one-sixth doesn't exist as a thing.

edds: It does in whitehall but never got moved over to the frontend toolkit.

futurefabric: Intentionally not moved across or not _yet_ moved across?
What shall I do here then?

evilstreak: This isn't the only place we use `$gutter-one-third / 2` in this app's CSS, so let's merge it and tidy it up later if and when it's moved to the toolkit.

futurefabric: OK cool, this should be good to merge now then :)

tommyp:
@futurefabric Good work. This was noticed ages ago and it got continually put on the backburner/forgot about so it's nice to get it sorted.

PR #152, opened 2015-01-29
Make facet parser accept text facets

tommyp:
With the changes in 04fcb52d we need to make this App work with both `text` and `multi-select` while we deploy them.

### DIFF
--- a/app/models/facet.rb
+++ b/app/models/facet.rb
@@ -7,6 +7,7 @@ class Facet
     @name = facet.name
     self.value = facet.value.presence
     @preposition = facet.preposition
+    @filterable = facet.filterable.nil? ? true : facet.filterable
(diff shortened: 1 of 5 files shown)


tommyp:
Updated the commit to fix some broken tests.

evilstreak:
If we deployed just this it would stop working – it needs the schema to be updated to include `filterable: true` to maintain functionality.

We need to default `filterable` to true to get it backwards compatible.

tommyp:
I thought we were deploying https://github.com/alphagov/specialist-publisher/pull/392 first? Or should I make them as independent from each other as I can?

tommyp:
Added some more code to make this more resilient against the _old_ versions of the schemas.

PR #154, opened 2015-01-30
Support finders with deep paths

jamiecobbett:
This is something of a grab bag of tidying up combined with a change to handling finders with deep paths. The former made the latter easier.

Context:
Ultimately, I am trying to get Contact finders working, and this makes progress towards that. I think we should definitely have an acceptance test for Contact finders, but trying to write that now proved difficult when there are significant changes needed in this app and in Rummager to make that happen.


evilstreak:
:+1:

jamiecobbett:
:smile: :dancer:

PR #159, opened 2015-02-03
Explain facets and links in Content Item

tommyp:
This commit adds an explanation for what keys should be in the facets and what they do. It also explains what this App does with the `links` provided by the Content Item.

### DIFF
--- a/docs/finder-content-item.md
+++ b/docs/finder-content-item.md
@@ -56,6 +56,82 @@ A boolean. Required.
 
 Used to decide if the summaries for Documents should be displayed in the results list. It will truncate the summary at the end of the first sentence.
 
-## facets
+## `facets`
 
-__TO DO__
+An array of hashes. Optional.
+
+Facets describe the metadata that the Finder deals with. They can contain several keys:
+
+### `key`
+
+A string. Required.
+
+`snake_case` string which matches to the field being searched in Rummager.
+
+### `filterable`
+
+A boolean. Required.
+
+Specifies if the facet should have a matching Filter for the results.
+
+### `display_as_result_metadata`
+
+A boolean. Required.
+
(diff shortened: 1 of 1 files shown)


docs/finder-content-item.md
```diff
@@ -56,6 +56,80 @@ A boolean. Required.
...
+
+### `display_as_result_metadata`
+
+A boolean. Required.
+
+Specifies if the facet should be returned as metadata underneath each result.
+
+### `name`
+
+A string. Required if `filterable` is set to `true`.
```
evilstreak: Would it be better to just say this is required? It doesn't seem like a difficult value to provide for all facets

docs/finder-content-item.md
```diff
@@ -56,6 +56,80 @@ A boolean. Required.
...
+
+A label for the Filter on the left hand side.
```
evilstreak: Perhaps something like:

Used to label the filter panel for the facet.

For date facets, it may be used to label the metadata shown for each result. (See also `short_name`)

docs/finder-content-item.md
```diff
@@ -56,6 +56,80 @@ A boolean. Required.
...
+
+### `proposition`
```
evilstreak: Typo: preposition

evilstreak:
A few comments, but this looks good otherwise

tommyp:
Fixed the errors pointed out by @evilstreak. 

@dsingleton, @edds and @elliotcm - Would you's like to read over this and see if it all makes sense?

dsingleton:
Looks OK to me.

edds:
Looks good. Thanks @tommyp. We can iterate if anyone else has any suggestions.

PR #160, opened 2015-02-03
Use new grid layout

dsingleton:
There a significant visual difference, so i'm not sure this is ready to merge, but I think theres enough to be worth discussion. I'd recommend looking at the commits (and their messages) in isolation, rather than the whole diff.

Mostly this converted fine, there were some sticking points around:
- Needing an element to wrap the filters and results columns, acting as a grid row - basically not the `header`. I chose `.filtering` as a not-awful start, but I think we could do better.
- Continuing to enforce the max-width of the `header`, to ensure the line length doesn't go beyond `2/3`rds, is done in a fiddly way. It requires floating 
 - One option would be to it in a row/column, but that would add quite a bit of markup just for CSS hooks.
 - Another would be using max-width 
- Finally, the show stopper; the gutters between columns is significantly smaller, `30px` in the new vs `60px` in the old. As the screenshots show, it's a hefty change.

The first point is semantic flddling, while the other two are a class of problem we'll see on other applications when trying to use modern grids and so are worth proper discussion.

Screenshots!

| Before | After |
| ---------| ------- |
| finder-before | finder-after | 
| finder-desktop-before | screenshot 2015-02-04 10 18 56 | 
| finder-email-before | finder-email-after | 
| finder-email-desktop-before | finder-email-desktop-after |

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -57,26 +57,11 @@
   }
 
   header {
-    width: $full-width;
-    margin-bottom: $gutter;
-    clear: both;
+    margin: $gutter 0;
 
-    h1 {
-      @include heading-48;
-      font-weight: bold;
-      margin: $gutter-half 0;
-
-      .aside {
-        @include heading-27;
-        display: block;
-        color: $grey-1;
-      }
-
-      @include media(tablet) {
-        margin: $gutter 0;
-        width: $two-thirds;
-        float:left;
-      }
+    @include media(tablet) {
+      width: $two-thirds;
+      float: left;
     }
 
     .email-link {
@@ -115,6 +100,7 @@
   }
 
   .filtering {
+    clear: both; // Clear `header` float for max-width
     @extend %grid-row;
   }
 
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -52,16 +52,12 @@
     @include bold-24;
   }
 
-  .govuk-beta-label {
-    margin: 0 $gutter;
-  }
-
   header {
-    margin: $gutter 0;
+    @extend %grid-row;
+    margin-bottom: $gutter;
 
-    @include media(tablet) {
-      width: $two-thirds;
-      float: left;
+    .title-and-metadata {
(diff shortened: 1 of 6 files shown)


app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -4,14 +4,16 @@
   <% if @signup.beta? %>
     <%= render partial: 'govuk_component/beta_label' %>
   <% end %>
-  <h1>
-    <span class="aside">Email alert subscription</span>
-    <%= @signup.name %>
-  </h1>
+  <div class="inner-block">
+  <%= render partial: 'govuk_component/title', locals: {
```
tommyp: Minor thing: indentation is off here.

tommyp:
Good PR :+1:. Not sure if the gutter spacing is a show stopper or we should hack the spacing back in. Thoughts @futurefabric?

app/views/email_alert_subscriptions/new.html.erb
dsingleton: Good catch, fixed.

futurefabric:
There's less visual separation with that tighter gutter, but I don't think this should prevent merging. 
We can always take a look at it separately later if needs be.

dsingleton:
Consensus seems to be that the smaller gutter is fine. It's the standard way the new grid works, which is used on `frontend`. 

I've just spotted a bug with the beta label layout, which i'll fix, and then this should be good to merge.

dsingleton:
Updated to handle cases with; related links, beta and email links:

screenshot 2015-02-04 16 08 47

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -133,17 +89,16 @@
...
   }
 
-  .filter-form,
-  .filtered-results {
-    @include media(tablet) {
-      float: left;
-      width: $one-third;
-    }
+  .filtering {
+    clear: both; // Clear `header` float for max-width
```
edds: now that `header` is a grid row I don't this this is needed. Grid rows include our containing floats shim.

edds:
One totally minor feedback point. Other than that this is a good change :+1: Lots of red.

dsingleton:
Fixed the `clear`, good catch!

PR #161, opened 2015-02-06
Get the filter restriction from the Content Item

tommyp:
As we now want Finders not restricted by the document type we can move this default restriction into the Content Item and call it from the Finder Presenter. We also need to use the document_type for the email alert subs controller as it passes this to the EmailAlertAPI.

### DIFF
--- a/lib/finder_frontend.rb
+++ b/lib/finder_frontend.rb
@@ -8,7 +8,6 @@ module FinderFrontend
   class FindDocuments
     def initialize(finder, params)
       @finder = finder
-      @document_type = finder.document_type
       @params = params
     end
 
@@ -19,7 +18,7 @@ module FinderFrontend
 
   private
 
-    attr_reader :document_type, :params, :finder
+    attr_reader :params, :finder
 
     def default_params
       {
@@ -45,7 +44,7 @@ module FinderFrontend
     end
 
     def massaged_params
-      ParamsMassager.new(params, document_type).to_h
+      ParamsMassager.new(params, finder).to_h
     end
 
     def rummager_api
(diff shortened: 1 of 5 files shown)


lib/finder_frontend.rb
```diff
@@ -93,10 +92,8 @@ def filter_params
         }
     end
 
-    def document_type_param
-      {
-        "filter_document_type" => document_type,
-      }
+    def default_filter
```
tommyp: Not sure about the name of this. Maybe `restrictive_filter` is better?

jamiecobbett:
As per my comment https://github.com/alphagov/specialist-publisher/pull/409#issuecomment-73078870 I think we might want to make this an array, not a single value, before it is cemented in. What do you think?

lib/finder_frontend.rb
```diff
@@ -93,10 +92,8 @@ def filter_params
...
+      finder.filter.to_h
```
tommyp: @evilstreak About your point the other day, that `filter_` is prepended to the `filter_params` - should I modify this to do that here too and remove `filter_` from the content item, or should we be explicit there and let it do it's thing here?

lib/finder_frontend.rb
jamiecobbett: `restrictive_filter` seems like a [tautology](http://en.wiktionary.org/wiki/tautology). `base_filter`?

tommyp:
@jamiecobbett Ah yeah. I meant to reply but couldn't remember where you'd made that comment. This is just a hash, that is merged in, so I don't think we need to make it an array.

jamiecobbett:
@tommyp Ah, yes you're right. I think I saw "filter" and thought singular.

lib/finder_frontend.rb
tommyp: I like `base_filter`. I'll go with that.

dsingleton:
LGTM, good explanation in the README :+1:

evilstreak:
I'm not sure about this handles the filters for email alert subscriptions.

This change means that although you can have a finder that filters by something other than document type, the email signup will only work properly for finders that filter by a document type and nothing else.

Can we talk about this in person?

PR #162, opened 2015-02-09
Allow Finders to have a summary

tommyp:
For the Policies as a Finder, we want to be able to have a bit of text to explain what the Finder is. This commit adds the ability for the Content Item to have a summary and for it to be rendered.

screen shot 2015-02-09 at 17 35 47

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -60,6 +60,10 @@
       @include grid-column(2/3);
     }
 
+    .summary {
+      margin-top: $gutter;
+    }
+
     .email-link {
       margin-top: $gutter;
       a {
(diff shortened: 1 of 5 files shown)


bradleywright:
I feel as though the addition of `delegate` should have been a separate refactoring commit (good change, by the way).

Not a blocker, just a slight deviation from the normal "add functionality, pass tests, refactor".

Also, no tests.

tommyp:
Ah yeah. I meant to add that to the commit message. I'll do it tomorrow as a separate commit.

edds:
On some of our mockups the summary is many paragraphs long. I was envisaging that the summary would be sent through in pre-pressed html to support that (probably by using govspeak so the editors could also add in links and such if they wanted to). Would be worth chatting this through with Ben to find out how he was expecting it to work.

tommyp:
Just realised Content Items have a description field. Just we be using this for the summary?

tommyp:
Added a test and moved the delegation to a separate commit based on @bradleywright's comment. Also made it Govspeak using the component after checking with @bmwelby and I double checked with Publishing that we shouldn't use the description field for this purpose.

evilstreak:
:+1:

PR #163, opened 2015-02-10
Use namespaced filter keys

tommyp:
As we don't want to have to prefix all filter keys with `filter_`, this commit makes this happening programmatically in `lib/finderfrontend.rb`. This also changes the readme to reflect the changes and the `email_alert_subscriptions_controller.rb` to call the document_type.

### DIFF
--- a/lib/finder_frontend.rb
+++ b/lib/finder_frontend.rb
@@ -61,7 +61,6 @@ module FinderFrontend
     def to_h
       keyword_param
         .merge(filter_params)
-        .merge(base_filter)
         .merge(order_param)
     end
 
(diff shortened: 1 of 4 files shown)


lib/finder_frontend.rb
```diff
@@ -93,7 +93,9 @@ def filter_params
     end
 
     def base_filter
-      finder.filter.to_h
+      finder.filter.to_h.reduce({}) { |memo, (k,v)|
+        memo.merge("filter_#{k}" => v)
```
evilstreak: Rather than do this `reduce` dance twice, what about making `filter_params` do something like:

```ruby
def filter_params
  params
    .expect("keywords")
    .merge(base_filter)
    .reduce({}) …
end
```

tommyp: Fair point. Amended the commit.

PR #165, opened 2015-02-13
Pass a context to the title component

tommyp:
As we now have Finders which may act as a format, we want to be able to display the context in the title. This is passed as a `human_readable_finder_format` in the Content Item.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -4,6 +4,7 @@ class FinderPresenter
 
   delegate :beta_message,
            :document_noun,
+           :human_readable_finder_format,
            :filter,
            :summary,
            to: :"content_item.details"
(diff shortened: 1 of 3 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -4,6 +4,7 @@ class FinderPresenter
 
   delegate :beta_message,
            :document_noun,
+           :context,
```
evilstreak: Is `context` the best name for the this? I think it might be too broad. How about something like `title_context`?

tommyp: Agreed that it might be too broad but I thought I'd go with what the component uses. @edds?

evilstreak: Its use in the component is implicitly scoped to being about the title because it's the `govuk_component/title` partial.

What sort of content are you intending this to be used for? It feels like there should either be a more descriptive name to use for this, or we should be using something like `title_context` and update `name` to be `title`.

tommyp: Good point about the scoping. It's going to be used for the Policy Areas and Programmes to say what type of page it is. I'll change it now.

PR #167, opened 2015-02-17
Atom feeds

tommyp:
This PR adds Atom feeds to Finders. The commits in this PR fit into 2 parts:

1. **Show a link Atom feed** - Adding an atom view, linking to it and updating the link with the query params when the filter is changed
2. **Add an autodiscovery link** - Adding an autodiscovery link for the Atom feed and updating it to match the one returned in the results

One thing to note is that all Filters being applied to the results will be passed through to the Atom link as it works as another view. This includes keyword based searches.

Screenshot showing placement:
screen shot 2015-02-17 at 16 56 51

The design and placement of the link was discussed with @markhurrell and it was decided that having it within the results block would make it clearer that the user was getting updates about _that_ list. The design may be iterated on in future.


app/views/finders/show.atom.builder
```diff
@@ -0,0 +1,21 @@
+atom_feed do |feed|
+  feed.title(@finder.name)
+
+  dates = []
+  @results.results.map(&:metadata).each do |r|
```
dsingleton: The update logic is a bit heavy for a template(ish), might be worth moving to the presenter? Would have the bonus of making it more easily testable too.

app/assets/javascripts/application.js
```diff
@@ -30,9 +30,14 @@ jQuery(function($) {
...
   var $form = $('.js-live-search-form'),
       $results = $('.js-live-search-results-block');
+      $atom_autodiscovery_link = $("link[type='application/atom+xml']").eq('0');
 
   if($form.length && $results.length){
-    new GOVUK.LiveSearch({$form:$form, $results:$results});
+    new GOVUK.LiveSearch({
+      $form:$form,
+      $results:$results,
+      $atom_autodiscovery_link:$atom_autodiscovery_link
```
dsingleton: Sorry, pedantry ahoy:

    Use camelCase when naming objects, functions, and instances. 

https://github.com/alphagov/styleguides/blob/master/js.md#naming-conventions

app/helpers/application_helper.rb
```diff
@@ -12,4 +12,8 @@ def input_checked(key, value)
   def facet_has_any_selected_values?(key)
     params.has_key?(key)
   end
+
+  def absolute_url_for(path)
+    URI.join("https://www.gov.uk", path)
```
dsingleton: Is there a `Plek` env variable for this? It'd be preferable if this respected host environment, eg dev, preview, prod

KushalP: > Is there a `Plek` env variable for this?

There is: https://github.com/alphagov/plek/blob/1e538d6cf825fa91cbaf07ba82de80854705edbc/lib/plek.rb#L88-93

app/views/finders/show.atom.builder
tommyp: Agreed that this is really heavy, but this is the only way I could figure out of getting the latest date for each document as not all documents expose `published_at` or something similar and some expose more than one date. CMA for example has Opened and Closed date. Unless anyone has a similar solution I can move this to a Presenter.

app/helpers/application_helper.rb
dsingleton: Maybe [`GOVUK_WEBSITE_ROOT`](https://github.com/search?q=GOVUK_WEBSITE_ROOT+%40alphagov&type=Code)?

bradleywright: Doesn't Rails already have absolute URL helpers? 

http://apidock.com/rails/v3.2.13/ActionView/Helpers/UrlHelper/url_for (see `:only_path` option)

tommyp: For the sake of documentation, it looks like `:only_path` works when `url_for` is constructing the URL from a hash or object. Giving it just the path throws an error.

app/controllers/finders_controller.rb
```diff
@@ -37,4 +38,8 @@ def facet_params
   def keywords
     params[:keywords] unless params[:keywords].blank?
   end
+
+  def atom_url
+    request.original_url.gsub('.json', '').gsub(finder_slug, "#{finder_slug}.atom")
```
dsingleton: Could this have a test? It would help show what the intended behaviour is, eg, does this preserve filtering arguments?

evilstreak: The `atom` URL is just the regular URL with `.atom` added before the query string. Since we update the URL each time the filters are updated, what's the benefit of doing this on the server and sending it in the response rather than doing it on the client?

app/views/finders/_results.mustache
```diff
@@ -1,3 +1,8 @@
+<div class="feed">
+  <span>Get updates to this list</span>
+  <a href="{{atom_url}}">feed</a>
```
evilstreak: I know this is the markup we use in other places (eg Whitehall), but is this OK for screen readers? Should we have more descriptive link text and visually hide it?

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -105,6 +107,7 @@
       presenter.to_hash[:documents].present?.should == true
       presenter.to_hash[:applied_filters].present?.should == true
       presenter.to_hash[:any_filters_applied].present?.should == true
+      presenter.to_hash[:atom_url].present?.should == true
```
evilstreak: For the sake of consistency this line is fine, but as an aside, I think you can use `presenter.to_hash[:atom_url].should be_present` here for a slightly nicer phrasing

evilstreak:
It's great to see a feature added like this which all existing finders can benefit from. :+1:

app/views/finders/show.atom.builder
dsingleton: I can't think of anything better, definitely worth moving into a Presenter regardless. Having some tests around it would be really good, and if we do find a better approach it's masked from the template anyway.

app/views/finders/_results.mustache
edds: We should only use visually hidden text as a last resort as it is so often forgotten about when refactoring. If you are concerned that the link needs more text for users of screen readers then you really should consider adding that text for visual users as well. Low skilled users often need the same extra clues that a screen reader users are helped by.

app/controllers/finders_controller.rb
tommyp: Won't that mean that Users without JS won't get the atom URL with matching query params?

evilstreak: Yes, good point. I still feel like it's a bit weird to be returning the atom URL as part of the JSON response. Maybe it's fine!

If we do want to keep it in the JSON, is there any prior art for this on GOV.UK? Perhaps something along the lines of HTML's `<link rel="alternate" />`? Since we're basically saying, "here's a place you can get this same content in a different format".

edds: @evilstreak: from [this json](https://www.gov.uk/government/announcements.json?keywords=&announcement_filter_option=government-responses&topics%5B%5D=all&departments%5B%5D=all&world_locations%5B%5D=all&from_date=&to_date=&_=1424274252289) which is used to power the annoncements filter:

```json
{
  ...
  "next_page":2,
  "next_page_url":"...",
  "atom_feed_url":"https://www.gov.uk/government/announcements.atom?announcement_filter_option=government-responses",
  "email_signup_url":"..."
}
```

app/views/finders/show.atom.builder
dsingleton: Actually, after chatting to @edds, he suggested the results are already in date order, and could just pick the date of the first? Using the same date field that rummager is ordered by.

tommyp: While that is correct, it has just me realise that the date we get back attached to the document isn't the same one that Rummager is sorting them by.

dsingleton:
Aside from the updated at date issue this all LGTM :+1:

app/controllers/finders_controller.rb
dsingleton: JSON issues aside, i think the controller method could be a lot cleaner, but i'd settle for some tests around it so it can be cleaned up later.

tommyp:
This PR has been updated based on @dsingleton's comments. Summary of the updates:
- add an `AtomPresenter`
- move the `atom_url` method into the `ResultSetPresenter` and construct this from `Finder.slug` and  `params.to_query`.
- get the `last_update` field from Rummager in order to use the correct dates in the feed
- add more test cases to the `FinderController` spec

app/presenters/atom_presenter.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(finder)
+    @finder = finder
+  end
+
+  def title
+    finder.name
+  end
+
+  def entries
+    @finder.results.documents
```
jamiecobbett: This can just be `finder` as you've gone to the trouble of setting `attr_reader :finder`.

app/presenters/result_set_presenter.rb
```diff
@@ -64,4 +65,8 @@ def documents
       SearchResultPresenter.new(result).to_hash
     end
   end
+
+  def atom_url
+    params.empty? ? "#{finder.slug}.atom" : "#{finder.slug}.atom?#{params.to_query}"
```
jamiecobbett: I'm now much more confident that there won't be unintended side effects from this :+1:

app/presenters/atom_presenter.rb
tommyp: Ah yeah. Didn't mean to leave that when I copied it from the view. Updated the commit.

PR #169, opened 2015-02-20
Better Atom entry id

tommyp:
This commit adds an EntryPresenter class and updates the `id` of the entry to follow the tag paradigm. It also uses the `schema_date` from the Builder to specify the year of the spec which it follows.

The work in this commit comes from the information [here](http://web.archive.org/web/20110514113830/http://diveintomark.org/archives/2004/05/28/howto-atom-id).

### DIFF
--- a/app/presenters/entry_presenter.rb
+++ b/app/presenters/entry_presenter.rb
@@ -1,0 +1,31 @@
+class EntryPresenter
+
+  delegate :title,
+           :summary,
+           :path,
+           to: :entry
+
+  def initialize(entry)
+    @entry = entry
+  end
+
+  def tag(schema)
+    "tag:#{website_root},#{schema_date(schema)}:#{entry.path}"
+  end
+
+  def updated_at
+    DateTime.parse(entry.last_update)
+  end
+
+private
+  attr_reader :entry
+
+  def website_root
+    Plek.current.website_root.gsub(/https?:\/\//, '')
+  end
+
+  def schema_date(schema)
+    schema.instance_variable_get(:@feed_options)[:schema_date]
+  end
+
+end
--- a/app/views/finders/show.atom.builder
+++ b/app/views/finders/show.atom.builder
@@ -8,7 +8,7 @@ atom_feed do |feed|
   end
 
   @feed.entries.each do |result|
-    feed.entry(result, id: result.path, url: absolute_url_for(result.path), updated: DateTime.parse(result.last_update)) do |entry|
+    feed.entry(result, id: result.tag(feed), url: absolute_url_for(result.path), updated: result.updated_at) do |entry|
       entry.title(result.title)
       entry.summary(result.summary, type: 'html')
     end
--- a/app/presenters/atom_presenter.rb
+++ b/app/presenters/atom_presenter.rb
@@ -9,11 +9,13 @@ class AtomPresenter
   end
 
   def entries
-    finder.results.documents
+    finder.results.documents.map { |d|
+      EntryPresenter.new(d)
+    }
   end
 
   def updated_at
-    DateTime.parse(entries.first.last_update)
+    entries.first.updated_at
   end
 
 private


bradleywright:
Something worth noting is that any software that uses the ID as a canonical reference will treat this like an entirely new document. It might be better to use something stable to the lifecycle of the document, like `content_id`.

tommyp:
From chatting with @edds I understood that we should be changing this ID when a document is updated?

bradleywright:
According to [the spec](https://tools.ietf.org/html/rfc4287#section-4.2.6):

>    When an Atom Document is relocated, migrated, syndicated, republished, exported, or imported, the content of its atom:id element MUST NOT change.  Put another way, an atom:id element pertains to all instantiations of a particular Atom entry or feed; revisions retain the same content in their atom:id elements.  It is suggested that the atom:id element be stored along with the associated resource.

which reads to me like the ID remains stable across all revisions. Atom already has a mechanism for flagging updates, which is the [`atom:updated` element](https://tools.ietf.org/html/rfc4287#section-4.2.15).

tommyp:
After some initial confusion I've updated this PR to follow the spec for the `atom:id`.

app/views/finders/show.atom.builder
```diff
@@ -8,7 +8,7 @@ atom_feed do |feed|
   end
 
   @feed.entries.each do |result|
-    feed.entry(result, id: result.path, url: absolute_url_for(result.path), updated: DateTime.parse(result.last_update)) do |entry|
+    feed.entry(result, id: "tag:#{Plek.current.website_root.gsub('http://www.', '')},#{feed.instance_variable_get(:@feed_options)[:schema_date]}:#{result.path}", url: absolute_url_for(result.path), updated: DateTime.parse(result.last_update)) do |entry|
```
jamiecobbett: This line is now really long and the tag logic seems complex enough to belong in a presenter.

rboulton:
When does the ":schema_date" value change?

Also, if the same item were to occur in more than one atom feed (eg, the feed for the same finder, but with different filters applied), should it have the same ID?  I believe it's desirable for it to do so, so that readers can tell if it's already been read.

tommyp:
`:schema_date` is the date of the Atom spec. Currently it's `2005`. The `id` of the feed entry will be consistent between differently filtered versions of the feed.

tommyp:
Added an EntryPresenter based on @jamiecobbett's comment.

app/presenters/entry_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+           :summary,
+           :path,
+           to: :entry
+
+  def initialize(entry)
+    @entry = entry
+  end
+
+  def tag(schema)
+    "tag:#{Plek.current.website_root.gsub('http://', '')},#{schema.instance_variable_get(:@feed_options)[:schema_date]}:#{entry.path}"
```
jamiecobbett: `Plek.current.website_root.gsub('http://', '')` isn't going to have the intended effect in environments where we use https - preview, staging and production. I think the right thing to do would be to use a regex: `gsub(%{https?://}, '')`.

Additionally, I think it's worth breaking the `Plek.current.website_root.gsub('http://', '')` and the `schema.instance_variable_get(:@feed_options)[:schema_date]` into private methods so that the line is shorter and it is easier to understand what the tag is made up of.

app/presenters/entry_presenter.rb
```diff
@@ -0,0 +1,31 @@
...
+           :summary,
+           :path,
+           to: :entry
+
+  def initialize(entry)
+    @entry = entry
+  end
+
+  def tag(schema)
+    "tag:#{website_root},#{schema_date(schema)}:#{entry.path}"
```
jamiecobbett: This is much more readable :+1:

bradleywright:
This is better. My only issue is that if the URL changes the ID will change, but I'm not sure that matters at this point.

tommyp:
I noticed that Whitehall uses a `document_id`, but we don't have anything like that available. since these results come from Rummager. This is using the `id` from Rummager as the path. Is there a better solution to follow what Whitehall does?

bradleywright:
Only if `content_id` is in Rummager, which it's not. Therefore this is fine :+1:

PR #172, opened 2015-03-02
Test content schema

tommyp:
This PR changes Finder Frontend to test against the example Content Item added to `govuk-content-schemas` in https://github.com/alphagov/govuk-content-schemas/pull/15. The commits are as follows

- removing the content item fixture for the Finder
- add a support file for govuk-content-schemas
- test specs using schema
- test features using schema and move away from talking about `cases` to the more generic `documents`
- update the jenkins.sh to use the schemas
- add a new jenkins-schema.sh to test Finder Frontend when a change is made to `govuk-content-schemas`


heathd:
:+1: looks good to me

features/support/env.rb
```diff
@@ -14,7 +14,6 @@
 require 'slimmer/test_helpers/shared_templates'
 World(Slimmer::TestHelpers::SharedTemplates)
 
-
```
dsingleton: Worth rebasing out the whitespace only changes in this file?

PR #175, opened 2015-03-03
Government navigation for government finders

tommyp:
This PR adds a render call for the `government_navigation` component in static which is added in https://github.com/alphagov/static/pull/553. This is only called if the Finder's `base_path` begins with `/government`. If it does, it also then extracts the second part of the `base_path` and uses this to decide which link should get a class of active added.

It also addresses some require weirdness with the `govuk_content_schemas_example` file.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -54,6 +54,14 @@ class FinderPresenter
     facets.filters
   end
 
+  def government?
+    slug.starts_with?("/government")
+  end
+
+  def government_content_section
+    slug.split('/')[2]
+  end
+
   def metadata
     facets.metadata
   end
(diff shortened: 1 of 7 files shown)


edds:
Is it worth adding a test to make sure that the component gets called from the appropriate finder?

tommyp:
What's the latest reckons about testing components?

jamiecobbett:
I think this is all that we have right now: https://github.com/alphagov/slimmer#testing-shared-components

tommyp:
Added a feature test which makes sure the component is on the page as per @edds' comment and removed some duplication in step definitions which I noticed.

jamiecobbett:
@tommyp I think you've got it down, but I've fleshed out the documentation for testing components: https://github.com/alphagov/slimmer/pull/116

PR #178, opened 2015-03-19
Refactor metadata

tommyp:
This PR cleans up some of the code around displaying metadata. Specially, it adds a helper to turn an array of links in a content item into an array of appropriate `<a>` tags. It also removes `#primary_organisation` and just uses the `#organisations` method.


edds:
Going to add to these commits to also add people as I was halfway through that.

PR #179, opened 2015-03-20
Add people to metadata

edds:
This includes the work from #178 which had the following description:

This PR cleans up some of the code around displaying metadata. Specially, it adds a helper to turn an array of links in a content item into an array of appropriate <a> tags. It also removes #primary_organisation and just uses the #organisations method.

It also adds in people into the metadata list.

Relies upon https://github.com/alphagov/govuk-content-schemas/pull/36 being merged for the tests to pass.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -17,6 +17,12 @@ module ApplicationHelper
     URI.join(Plek.current.website_root, path)
   end
 
+  def page_metadata_links(metadata)
+    metadata.inject({}) do |memo, (type, links)|
+      memo.merge(type => arr_to_links(links))
+    end
+  end
+
   def arr_to_links(arr)
     arr.map { |link|
       link_to(link.title, link.web_url)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,4 +16,10 @@ module ApplicationHelper
   def absolute_url_for(path)
     URI.join(Plek.current.website_root, path)
   end
+
+  def arr_to_links(arr)
+    arr.map { |link|
+      link_to(link.title, link.web_url)
+    }
+  end
 end
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -87,15 +87,11 @@ class FinderPresenter
   end
 
   def organisations
-    content_item.links.organisations
+    content_item.links.organisations || []
   end
 
   def part_of
-    @part_of = content_item.links.part_of || []
-  end
-
-  def primary_organisation
-    organisations.first
+    content_item.links.part_of || []
   end
 
   def related
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -86,12 +86,13 @@ class FinderPresenter
     content_item.details.show_summaries
   end
 
-  def organisations
-    content_item.links.organisations || []
-  end
+  def page_metadata
+    metadata = {
+      part_of: part_of,
+      from: from,
+    }
 
-  def part_of
-    content_item.links.part_of || []
+    metadata.reject { |_, links| links.empty? }
   end
 
   def related
@@ -113,6 +114,22 @@ class FinderPresenter
 
 private
   attr_reader :content_item, :values
+
+  def part_of
+    content_item.links.part_of || []
+  end
+
+  def organisations
+    content_item.links.organisations || []
+  end
+
+  def people
+    content_item.links.people || []
+  end
+
+  def from
+    organisations + people
+  end
 
   def facet_search_params
     facets.values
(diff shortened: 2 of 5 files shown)


edds:
Gonna clean this up a little.

app/helpers/application_helper.rb
```diff
@@ -16,4 +16,17 @@ def facet_has_any_selected_values?(key)
   def absolute_url_for(path)
     URI.join(Plek.current.website_root, path)
   end
+
+  def page_metadata_links(metadata)
+    metadata.inject({}) do |memo, (type, links)|
+      memo[type] = arr_to_links(links)
+      memo
```
evilstreak: It's more idiomatic to use `merge` rather than mutation when using `inject`/`reduce`:

```ruby
def page_metadata_links(metadata)
  metadata.reduce({}) { |memo, (type, links)|
    memo.merge(type => arr_to_links(links))
  }
end
```

If you think mutation would be a better approach (I don't!), then `with_object` lets you use mutation rather than return values:

```ruby
def page_metadata_links(metadata)
  metadata.each.with_object({}) do |(type, links), memo|
    memo[type] = arr_to_links(links)
  end
end
```

app/presenters/finder_presenter.rb
```diff
@@ -86,16 +86,15 @@ def show_summaries?
...
   end
 
-  def organisations
-    content_item.links.organisations
-  end
-
-  def part_of
-    @part_of = content_item.links.part_of || []
+  def any_page_metadata?
+    page_metadata.any? { |type, links| links.any? }
```
evilstreak: Would it be better to bake this into the return value of `page_metadata` instead? Eg

```ruby
def page_metadata
  metadata = {
    part_of: part_of,
    from: from,
  }

  metadata.reject { |_, links| links.empty? }
end
```

That way, you could just use `page_metadata.any?` for this method, and other consumers of `page_metadata` don't have to worry about some of the values being empty arrays.

Relatedly, if you have a variable you aren't using (`type` in this case) it's good practice to use `_` as the variable, or prefix it like `_type` if you want to give context on what it is but indicate it's unused.

app/presenters/finder_presenter.rb
```diff
@@ -118,6 +117,22 @@ def label_for_metadata_key(key)
...
+  def organisations
+    content_item.links.organisations || []
+  end
+
+  def people
+    content_item.links.people || []
+  end
+
+  def from
+    (organisations + people).compact
```
evilstreak: You shouldn't need to `compact` this, as neither `organisations` nor `people` should be returning arrays including `nil`.

lib/govuk_content_schema_examples.rb
```diff
@@ -21,8 +21,8 @@ module GovukContentSchemaExamples
     include GdsApi::TestHelpers::ContentStore
 
     # Returns a hash representing an finder content item from govuk-content-schemas
-    def govuk_content_schema_example(name)
-      string = GovukContentSchemaTestHelpers::Examples.new.get('finder', name)
+    def govuk_content_schema_example(name, format=nil)
```
evilstreak: Rather than default this parameter to `nil` and provide a default value where it's used, it would be better to just add the default value here:

```ruby
def govuk_content_schema_example(name, format="finder")
  string = GovukContentSchemaTestHelpers::Examples.new.get(format, name)
  JSON.parse(string)
end
```

spec/presenters/finder_presenter_spec.rb
```diff
@@ -60,6 +71,21 @@
     it "exposes the government_content_section" do
       government_presenter.government_content_section.should == "policies"
     end
+
+    it "has metadata" do
+      government_presenter.any_page_metadata?.should be_true
```
evilstreak: There are a couple of RSpec matchers which can make testing booleans a bit smoother: https://github.com/rspec/rspec-expectations#predicate-matchers

To use it in this case would probably require renaming the method to `has_page_metadata?`, which would then allow this test to be:

```ruby
it "has metadata" do
  government_presenter.should have_page_metadata
end
```

For the negative case swap the `should` for `should_not`.

Whether you think that's an improvement worth having I'll leave up to you to decide.

One other thing I think it's worth noting: the `should` syntax is deprecated in RSpec 3 in favour of the `expect` syntax, in order to avoid monkey patching `Object`. This code base mixes styles already, so again I'll leave it to you to decide what to do with that information here.

```
it "has metadata" do
  expect(government_presenter).to have_page_metadata
end
```

evilstreak:
I've added lots of comments here, but they really all relate to Ruby style.

The substance of this change is great, and very easy to follow. :+1:

edds:
Many thanks for those suggestions @evilstreak, they were really clear and made a lot of sense. I have updated the commit accordingly.

PR #181, opened 2015-03-26
Update finder to order with public_timestamp

tekin:
All existing specialist content should now include a "public_timestamp" in the search index, which is semantically identical to last_update. This change will make it possible for finders to work with non-specialist content, which sets a public_timestamp, but not last_update.

Trello: https://trello.com/c/oMo0JVPx/159-update-finder-frontend-to-order-by-public-timestamp-and-not-last-update

/cc @tommyp 

- [x] dependent rummager changes deployed
- [x] rummager index rebuilt
- [x] code review
- [ ] product owner signoff

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,14 +1,14 @@
 class Document
-  attr_reader :title, :last_update
+  attr_reader :title, :public_timestamp
 
   def initialize(attrs, finder)
     attrs = attrs.with_indifferent_access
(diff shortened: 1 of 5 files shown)


edds:
LGTM

tommyp:
I'm happy with the code but would feel better testing this locally against the existing Finders however I'm having trouble with my VM and Rummager so once I get that sorted and checked I'll merge :+1:

PR #182, opened 2015-03-31
Applicable nations

tommyp:
This PR allows policies to display which nations they apply to and link to the alternative policies for the inapplicable nations. To do this I've modified the existing helpers for displaying page metadata to handle strings and renamed them to reflect this.

The Finder Presenter generates the sentences. This will contain the links to the alternative policies if there are any alternative policies, which should only be there if there isn't all 4 nations in the `applies_to` array. Note that these don't need to be present and a Policy can apply to certain nations without having alternative policies. [Ticket](https://trello.com/c/01iC4gHk/82-add-ability-to-set-inapplicable-nations-on-policies).

Screenshot:

screen shot 2015-04-07 at 16 12 57

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -1,4 +1,5 @@
 class FinderPresenter
+  include ActionView::Helpers::UrlHelper
 
   attr_reader :name, :slug, :organisations, :keywords
 
@@ -90,9 +91,10 @@ class FinderPresenter
     metadata = {
       part_of: part_of,
       from: from,
+      other: other,
     }
 
-    metadata.reject { |_, links| links.empty? }
+    metadata.reject { |_, links| links.blank? }
   end
 
   def related
@@ -129,6 +131,33 @@ private
 
   def from
     organisations + people
+  end
+
+  def other
+    if applicable_nations_html_fragment
+      { "Applies to" => applicable_nations_html_fragment }
+    end
+  end
+
+  def applicable_nations_html_fragment
+    nation_applicability = content_item.details.nation_applicability
+    if nation_applicability
+      applies_to = nation_applicability.applies_to.map(&:titlecase)
+      alternative_policies = nation_applicability.alternative_policies.map do |alternative|
+        link_to(alternative.nation.titlecase, alternative.alt_policy_url, ({rel: 'external'} if is_external?(alternative.alt_policy_url)))
+      end
+      if alternative_policies.any?
+        "#{applies_to.to_sentence} (see policy for #{alternative_policies.to_sentence})".html_safe
+      else
+        applies_to.to_sentence
+      end
+    end
+  end
+
+  def is_external?(href)
+    if host = URI.parse(href).host
+      "www.gov.uk" != host
+    end
   end
 
   def facet_search_params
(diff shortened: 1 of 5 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -90,9 +91,10 @@ def page_metadata
     metadata = {
       part_of: part_of,
       from: from,
+      other: other,
     }
 
-    metadata.reject { |_, links| links.empty? }
+    metadata.reject { |_, links| !links.present? }
```
tekin: The  opposite of `present?` is `blank?`, which might read better.

app/presenters/finder_presenter.rb
```diff
@@ -131,6 +133,33 @@ def from
...
 
+  def other
+    if applies_to
+      {
+        "Applies to" => applies_to.html_safe,
+      }
+    end
+  end
+
+  def applies_to
```
tekin: It feels like this method would be better as two separate methods: one boolean method that determines if the document applies to all nations, and another to construct the meta data fragment. Then the boolean method could be used in the `other` method above. Something like this:

```ruby

def other
  unless applies_to_all_nations?
    { "Applies to" => applicable_nations_html_fragment }
  end
end

def applies_to_all_nations?
  content_item.details.nation_applicability && content_item.details.nation_applicability.any?
end

def applicable_nations_html_fragment
  applies_to = content_item.details.nation_applicability.map(&:titlecase)
  alternative_policies = applicability.alternative_policies.map do |alternative|
    if alternative.alt_policy_url
      link_to(alternative.nation.titlecase, n.alt_policy_url, ({rel: 'external'} if is_external?(n.alt_policy_url)))
    else
      alternative
    end
  end
  "#{applies_to.to_sentence} (see policy for #{alt_policies.to_sentence})".html_safe
end
```

Also, is that if/else is required? Looking at the [policy-publisher code](https://github.com/alphagov/policy-publisher/pull/34/files#diff-00c608fb3b501b4f92d04bcf03760afeR6), it isn't possible to set a nation as being inapplicable without supplying an alternative policy URL. And if it was, would the array contain strings and not the hash that it contains when a URL is provided?

tekin:
Is the CSS for the external link elsewhere? The external icon looks a bit skewiff.

spec/presenters/finder_presenter_spec.rb
```diff
@@ -82,6 +93,28 @@
...
+    end
+
+    it "has no applicable nations in the metadata if it applies to all nations" do
+      applies_to = government_presenter.page_metadata
+      expect(applies_to).not_to have_key(:other)
+    end
+
+    it "sets rel='external' for an external link" do
+      href = "http://www.externalwebsite.com"
+      expect(national_applicability_presenter.send(:is_external?, href)).to be(true)
```
tekin: Testing private methods is a bit of a smell. I think a test that examines the output fragment and checks that the link contains the `rel="external"` attribute would be more useful and valuable.

tommyp:
@tekin Sorted out all those comments and found some CSS to do the correct positioning for the icon.

tekin:
Is that CSS tweak still there? It still looks wonky when I run it locally:

screen shot 2015-04-09 at 12 29 51

Note the colour as well.

tommyp:
Try clearing your cache. It looked a bit wonky for me too until I did that.

tekin:
Ah, yep, straightened out now. The colour is still off though. Otherwise, looks good! :+1:

tekin:
Just noticed the colour is the same as on other pages, I guess it's a thing.

PR #183, opened 2015-04-08
Handle nil metadata gracefully

tommyp:
Finder frontend currently has a bug which means that if it gets a metadata field returned which is nil then the app will 500. This commit makes the document model handle this by not including the nil metadata and silently triggering an Airbrake notification.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -74,7 +74,7 @@ private
 
   def tag_labels_for(key)
     Array(attrs.fetch(key, []))
-      .map { |tag| tag.fetch("label") }
+      .map(&method(:get_metadata_label))
      .select(&:present?)
   end
 
@@ -102,5 +102,12 @@ private
     metadata_hash.merge(
       name: finder.label_for_metadata_key(metadata_hash.fetch(:name))
     )
+  end
+
+  def get_metadata_label(tag)
+    tag.fetch("label")
+  rescue => error
+    Airbrake.notify(error)
+    nil
   end
 end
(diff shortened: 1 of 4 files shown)


tekin:
I think it would be helpful if there was a test to cover this.

tommyp:
Yeah I didn't get a chance to do it yesterday and other PRs are more urgent to fix so I'll get on it after that.

app/models/document.rb
```diff
@@ -64,7 +64,7 @@ def tag_metadata
 
   def tag_labels_for(key)
     Array(attrs.fetch(key, []))
-      .map { |tag| tag.fetch("label") }
+      .map(&method(:get_metadata_label))
```
tekin: Is the `method` call necessary or would `.map(&:get_metadata_label)` work?

tommyp: `get_metadata_label` isn't a method on the result of `tag.fetch("label") so no.

evilstreak: To elaborate a little:

`Symbol#to_proc` returns a proc that sends the symbol as a message to the object given to the proc. `&:foo` is equivalent to the block `{ |x| x.foo }`.

`Method#to_proc` returns a proc that calls the method with the arguments given to the proc. `&method(:foo)` is equivalent to the block `{ |x| foo(x) }` (although it supports multiple arguments as well).

edds: Given this is causing some confusion is it worth expanding this out into the more explicit `{ |tag| get_metadata_link(tag) }` style?

tekin: Of course. Agree with @edds comment, better to call the method directly rather than pass it in as a `Method` object.

tommyp: Even though `Method#to_proc` is used elsewhere in this model?

PR #189, opened 2015-04-22
Add the possibility to add a logo on a finder

jackscotti:
European Structural and Investment funds must comply with EU legislations and they therefore need to display the EU logo.

Quote from the legislation: "the Union emblem and the reference to the Union shall be visible, when landing on the website, inside the viewing area of a digital device, without requiring a user to scroll down the page".

- Why is this not a GOV.UK Component? Because we might receive different logos, with different text and with different styles.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -7,6 +7,7 @@ class FinderPresenter
            :document_noun,
            :human_readable_finder_format,
            :filter,
+           :logo_path,
            :summary,
            to: :"content_item.details"
 
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -85,6 +85,15 @@
       }
     }
 
+    .logo {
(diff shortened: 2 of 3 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -23,6 +23,10 @@ def beta?
     content_item.details.beta
   end
 
+  def logo
+    content_item.details.logo
```
tommyp: Since you're just calling `content_item.details.logo` here, you can add it to the `delegate` section at the [top of the class](https://github.com/alphagov/finder-frontend/blob/add_eu_logo_to_esif_funds/app/presenters/finder_presenter.rb#L6).

jackscotti: Uhm... will do, thanks Tommy.

tommyp: Delegation is good :+1:

issyl0: Done. I quite like that...

PR #190, opened 2015-04-24
Remove email_alert_signup_enabled flag

tommyp:
As all Finders which now want a signup page have it live, we no longer
need to hide the link to the page. This commit removes the code for this
and changes it to be displayed based on the presence of the link in the
`email_alert_signup` array.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -23,10 +23,6 @@ class FinderPresenter
     content_item.details.beta
   end
 
-  def email_alert_signup_enabled?
-    content_item.details.email_signup_enabled
-  end
-
   def email_alert_signup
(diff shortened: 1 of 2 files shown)


evilstreak:
:+1: This is good, but it's not backwards compatible due to AAIB still have this flag set to false. This ought to be blocked by a PR updating the AAIB Finder config in Specialist Publisher

tommyp:
I was about to open that PR, but had to get on a ✈️

I'll open it when I get somewhere with internet.

tommyp:
Don't deploy before https://github.com/alphagov/specialist-publisher/pull/477 is deployed.

jackscotti:
@tommyp alphagov/specialist-publisher#477 is deployed now.

PR #192, opened 2015-04-28
Short-term hack to set default policy date filter

dsingleton:
> As a user looking at a policy finder
> I need to initially see information from the current government only
> So that I am not confused by old content that won't be as relevant

- https://trello.com/c/sRi6RXKd/118-default-date-view-on-finder-frontend-policy-frontend

This this will be used for a few weeks post-election and should be
completely removed afterwards. It only applies to a policy finders, (eg
`/government/policies/benefits-reform`, but not the finder of policies, eg
`/government/policies` nor any other finders, eg, `/cma-cases`)

This will not show documents-related-to-policy published under the previous
government, though they can been seen by removing/changing the published
after date in the finder UI.

This can be removed once there are documents-related-to-policy published
by the new government. It's intended to be reverted and entirely removed
at that point, including the test that needed to be modified (stubbing rummager
request URLs is a bit manual, doubly so for non-MOSW finders - improving
that is on our tech debt list)

`date_new_government_formed` will need updating to be the day the new
government forms, if it is not the day after polling day.

Longer term it would be nice to support this properly through the finder schema, allowing a filter facet to set a default value. That required more time than we had to do/test properly, so we're going with this hacky-but-isolated short term approach.

CC @evilstreak and @tommyp 

screenshot 2015-04-28 15 31 22

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -2,6 +2,7 @@ require 'gds_api/helpers'
 
 class FindersController < ApplicationController
   include GdsApi::Helpers
+  before_filter :apply_policy_finder_default_date
 
   def show
     @results = ResultSetPresenter.new(finder, facet_params)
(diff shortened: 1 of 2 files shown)


evilstreak:
:+1:

tommyp:
Also :+1:

rboulton:
:+1: ... But what's a MOSW finder?

tommyp:
@rboulton MOSW = Ministry of Silly Walks. It's what I used for the example in the schema.

dsingleton:
fyi:
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/finder/frontend/examples/finder.json

On 29 April 2015 at 08:46, Tommy Palmer <notifications@github.com> wrote:

> @rboulton <https://github.com/rboulton> MOSW = Ministry of Silly Walks.
> It's what I used for the example in the schema.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/finder-frontend/pull/192#issuecomment-97336156>
> .
>



-- 
*David Singleton*

Cabinet Office, Government Digital Service
Developer, GOV.UK
david.singleton@digital.cabinet-office.gov.uk

PR #195, opened 2015-05-06
Increase summary font size

futurefabric:
The size of the intro text on the policy finders is currently too small.
This change uses the standard GOV.UK body copy size.


alicebartlett:
Extra commendation for the branch name. ⭐️

futurefabric:
Cool, thanks! 
It's a good font, it needs more big ups :zap:

PR #196, opened 2015-05-06
Add keyline beneath finder summary

futurefabric:
For finders that feature a summary a keyline is needed to add visual structure to the page. 
This clearly marks where the summary ends and where the finder begins.

# Before

without

# After

with

### DIFF
--- a/app/views/finders/show.html.erb
+++ b/app/views/finders/show.html.erb
@@ -58,6 +58,10 @@
   <% end %>
 </header>
 
+<% if finder.summary %>
+  <div class="keyline"><!-- --></div>
+<% end %>
+
 <div class="filtering">
   <%= render finder.facets %>
   <div class='js-live-search-results-block'>
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -118,6 +118,13 @@
     }
   }
 
+  .keyline {
+    height:1px;
+    line-height:1px;
+    background-color:$grey-3;
+    margin: 0 0 $gutter 0;
+  }
+
   .filtering {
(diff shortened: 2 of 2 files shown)


app/views/finders/show.html.erb
```diff
@@ -58,6 +58,10 @@
   <% end %>
 </header>
 
+<% if finder.summary %>
+  <div class="keyline"><!-- --></div>
```
fofr: Would an `<hr>` tag be more appropriate semantically? "a paragraph-level thematic break", which is what the commit suggests this is for.

futurefabric: I did start with an `<hr>` but settled on a div instead. Mainly because this isn't at a paragraph level so much as separating `header` from `.filtering`

I think the cleanest way would be to use a `border-bottom` declaration on `header` but I can't do this as it will overhang into the grid gutters left and right.

tommyp: If you wanted to do the `border-bottom` on the header, you could update Finder Frontend to use the `core_layout` from Slimmer.

futurefabric: @tommyp This sounds interesting. Got any time to take a look with me?

dsingleton: I can run you through that too, if @tommyp's busy.

edds:
Going to close this as there doesn't seem to be any activity on it. If this is still wanted then it can be re-opened.

futurefabric:
@edds This is still a presentational feature that the policy finders need, but I'm not likely to pick this up now I'm not on the team :disappointed:. I added it to the wiki as an outstanding thing that needs to happen. Is it likely to still get picked up?

PR #197, opened 2015-05-07
Handle nil input in DateParser.parse

evilstreak:
When a request is made with parameters like `?date[from]&date[to]` Rails
converts that into

    "date" => {
      "from" => nil,
      "to" => nil,
    }

That ends up getting passed down to `DateParser.parse` which blows up
when calling `#strip` on the `nil` input.

This fixes a recurring error reported in errbit, eg https://errbit.production.alphagov.co.uk/apps/54088bad0da1152a9f001417/problems/5542817b65786364ce010100

### DIFF
--- a/app/parsers/date_parser.rb
+++ b/app/parsers/date_parser.rb
@@ -1,7 +1,7 @@
 class DateParser
   def self.parse(date_string)
 
-    date_string = date_string.strip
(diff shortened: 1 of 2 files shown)


tommyp:
My inbox says thanks.

PR #203, opened 2015-05-18
Allow disabling of atom feed

tommyp:
For Finders which aren't sorted by a date, it doesn't make sense to have
the atom feed still be present. This commit allows the feed to be
disabled through a boolean in the content item.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -55,6 +55,10 @@ class FinderPresenter
     facets.filters
   end
 
+  def feed_disabled?
+    content_item.details.feed_disabled || false
+  end
+
   def government?
     slug.starts_with?("/government")
   end
--- a/app/presenters/result_set_presenter.rb
+++ b/app/presenters/result_set_presenter.rb
@@ -18,7 +18,7 @@ class ResultSetPresenter
       applied_filters: describe_filters_in_sentence,
       documents: documents,
       any_filters_applied: any_filters_applied?,
-      atom_url: atom_url
+      atom_url: finder.feed_disabled? ? nil : atom_url,
     }
   end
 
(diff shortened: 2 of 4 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -55,6 +55,10 @@ def filters
     facets.filters
   end
 
+  def feed_disabled?
```
evilstreak: What about calling it `feed_enabled?` and defaulting it to `true`? Avoiding double negatives is generally desirable

evilstreak:
Since the rationale behind this is to let us have feeds only for date-sorted lists, what about linking this to sorting instead?

Is it desirable to be more declarative, using the content item to describe what the finder is, rather than using it to configure how it should be shown?

app/presenters/finder_presenter.rb
tommyp: I had that originally but after chatting with @elliotcm his advice was to use the method to modify what is the default for Finders.

tommyp:
The other semi-related PR that I have was to allow a default order but I think the idea is that at some point users will be able to decide their own way of ordering these. I think these 2 pieces of work are only related in that they were both needed for the Policies finder because we can't order it by the content changed within each policy.

tommyp:
@evilstreak  - How would you feel if we disabled atom feeds if the default order is title and show it in every other case?

evilstreak:
What about enabling them if ordering by `-public_timestamp` (ie the default) and not in any other case? It seems like we wouldn't want a custom ordered finder to have an atom feed more often than we would.

tommyp:
I was thinking that we may want it for content that is sorted by other date metadata and the other key that you can sort it by for hiding it would be `title`, but you're point also makes sense. I'll go with that for now and we can always refactor it later.

PR #207, opened 2015-05-21
Some refactoring

evilstreak:
This is mostly preparatory work to allow us to introduce a `FinderAPI` object that the controller can use to fetch a finder and its results from. That will let us have a clean split between logic for fetching data (ie parsing user parameters, making calls to APIs and combining responses into a sensible data structure) and logic for presenting the data to the user (ie all the presenters).

/cc @dsingleton


jackscotti:
LGTM :+1:

dsingleton:
Going in the right direction :+1:

PR #208, opened 2015-05-21
Disable atom feed if finder has a default order

tommyp:
When a Finder has a default_order we want to disable the atom feed from
appearing. The current behaviour of a Finder having an atom feed is
unchanged unless the Finder has a default_order. This commit changes the
atom_url method to return nil if the Finder has a default order, hides
the atom_url link if that is nil in the hash and only responds to the
atom format if the atom_url method isn't nil.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -124,8 +124,12 @@ class FinderPresenter
     end
   end
 
+  def atom_feed_enabled?
+    !default_order.present?
+  end
+
   def atom_url
-    ["#{slug}.atom", values.to_query].reject(&:blank?).join("?")
+    ["#{slug}.atom", values.to_query].reject(&:blank?).join("?") if atom_feed_enabled?
   end
 
 private
--- a/spec/presenters/finder_presenter_spec.rb
+++ b/spec/presenters/finder_presenter_spec.rb
@@ -13,6 +13,8 @@ RSpec.describe FinderPresenter do
 
   let(:national_applicability_with_internal_policies_presenter) { described_class.new(national_applicability_with_internal_policies_content_item) }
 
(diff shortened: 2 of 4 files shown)


spec/presenters/finder_presenter_spec.rb
```diff
@@ -134,6 +134,13 @@
         presenter.atom_url.should == "/mosw-reports.atom?format=publication&keyword=legal&state=open"
       end
     end
+
+    context "when the finder is ordered by title" do
+      it "doesn't return an atom_url" do
+        presenter.stub(:default_order).and_return('title')
```
evilstreak: I'd prefer to avoid stubbing a method on the object we're testing. Can you make the content item respond with the correct value instead?

app/presenters/finder_presenter.rb
```diff
@@ -125,7 +125,7 @@ def display_key_for_metadata_key(key)
   end
 
   def atom_url
-    ["#{slug}.atom", values.to_query].reject(&:blank?).join("?")
+    ["#{slug}.atom", values.to_query].reject(&:blank?).join("?") unless default_order
```
evilstreak: Since the condition is non-intuitive, I think it would be better to pull this into a private `#atom_feed_enabled?` method. That way the `#atom_url` only has to worry about how to build the URL, instead of whether it should build it at all.

tommyp: I originally went that way, but the frontend part of it is checking if the `atom_feed` is nil or not, so I went down this route instead. It seemed like an extra step that wasn't needed but I guess it makes this clearer.

tommyp:
Updated this PR to use the Policies Finder for testing this behaviour and refactored the methods involved.

PR #209, opened 2015-05-27
Separate data fetching from presentation

evilstreak:
## Aim of the change

This PR separates the application logic responsible for fetching data from APIs from the logic for presenting that data to the user.

The flow of data goes like:

- `FindersController` makes a call to a `FinderApi` object with the base path and user parameters
- The API returns a single data structure representing the finder and its results
- The controller passes this data into a `FinderPresenter`
- The controller passes the finder presenter into a `ResultSetPresenter`
- The controller renders the view with both presenters

Ideally the controller would only instantiate a `FinderPresenter`, and not also have to instantiate a `ResultSetPresenter`, but that's outside the scope of this PR.

## Reviewing the changes

This is best reviewed by commit by commit.

There are four commits with the major changes:

- 703db93082355116134976535ac1db1cf50b9816 adds `FinderApi`
- 57d3469a604ac750f9d59834b6d580bcaa7b1f22 uses the new `FinderApi`
- 6dd3e4357efe3a65007fc360c075938553cca06d extracts query building logic from `FinderPresenter`
- b3c91f7c0ac133051f7b66ee241516a79a72bac3 adds a `FilterQueryBuilder`

The rest of the commits are more contained refactoring and deletion of unused code.

## What's missing

Unit tests. I'd like to add several, for the parts of the system which I think aren't going to be ripped to pieces by near-future work. Before I do, I'd like to get some feedback from people about whether the approach I've taken seems good.

**Note:** this pull request does little the improve the nomenclature problem we have. It's still no clearer what is meant by all the words. Facet, filter, search, query, keywords, params, metadata. A lot of these terms are vague, and many are used in more than one way in this code base.

I'm happy to make some naming changes as part of this work, especially to newly created classes and methods, but I think renaming is something that will continue for a while until we have something easy to follow.


app/lib/finder_api.rb
```diff
@@ -0,0 +1,43 @@
+require 'delegate'
+
+class FinderApi
+  def initialize(content_api:, search_api:)
+    @content_api = content_api
```
tommyp: Are we happy with calling this `content_api` when it should be the `content_store`?

app/controllers/application_controller.rb
```diff
@@ -14,6 +14,10 @@ def set_slimmer_headers
     response.headers[Slimmer::Headers::TEMPLATE_HEADER] = "header_footer_only"
   end
 
+  def finder_base_path
+    "/#{finder_slug}"
```
tommyp: Is it a bit weird that we take the slug from the `params` and then add a slash back in front of it for the `base_path`?

evilstreak: Yes, it's a bit weird. We could look at changing it, but until we move the email signup stuff into email-alert-frontend we have to handle paths with multiple components anyway.

tommyp:
Few minor things but this is good. Definitely a step in the right direction :+1:

dsingleton:
Broadly :+1:

app/lib/search_query_builder.rb
```diff
@@ -0,0 +1,91 @@
...
+  def order_by_relevance_query
+    {}
+  end
+
+  def default_order_query
+    {"order" => default_order}
+  end
+
+  def keyword_query
+    keywords ? {"q" => keywords} : {}
```
rboulton: I think you want `"q" => params["keywords"]` here.

tommyp: It's calling the method below.

evilstreak:
Closing while I get the testing sorted. I'll open a new PR when ready

PR #210, opened 2015-05-28
Replace select filter with govuk component

alicebartlett:
This pull request replaces the `select_filter` with the `option-select-component` from http://github.com/alphagov/static, bringing full circle some work begun in January to componentise the `select_filter` to allow it to be used across GOV.UK without replication of code.

This PR:
- Deletes all of the code that has now been moved to static (JavaScript, CSS, HTML)
- Adds a method to the select_filter model to provide lists of options for the `option-select`
- Adds the call to the partial
- Adds a small amount of CSS to correct the margins for the component.


app/models/select_facet.rb
```diff
@@ -6,6 +6,17 @@ def initialize(facet)
     @allowed_values = facet.allowed_values
   end
 
+  def options
+    allowed_values.map do | allowed_value |
+      {
+        "value" => allowed_value['value'],
```
tommyp: `allowed_value` should be an object here so you can call `allowed_value.value`, etc.

edds:
\o/ w00t!  

Really glad this change is happening. Also "+24 −498", pretty rad diff stats.

tommyp:
This is ace :+1: 

I have one minor bit of feedback and those diff stats are unreal, but I think this is the first bit of work that ports _functionality_ to be powered by a component. Obviously we aren't able to test this with the component being rendered (which is what that test failure is about) but maybe there needs to be a discussion about how we ensure functionality being powered by components doesn't break?

dsingleton:
:balloon: :tada:  :balloon: :tada:  :balloon: :tada:  :balloon: 

@tommyp Thats a good question, and not one we have a solid answer to yet with components and the contracts static has with clients. Worth more discussion.

Pulling in components/static during tests is possible (not dissimilar to content-schemas), but needs some thought about how that would work in practice, and any performance impact on tests. 

In the mean time, the tests could be re-written, so that:
 - Anything that is asserting a value is present in a filter (eg, an option rendered/selected) should be able to do so from the JSON dump of arguments passed to a component.
- Interactions with component UI (like `select_filters`) could construct/visit URLs directly, rather than being tied to the component.

Where we're using complex components it might be worth adding `Smokey` tests to give us some full integration testing.

tommyp:
@dsingleton I think the latter option is probably the easiest in the short term, but it may be better to look into rendering the actual component as I'm sure that as we port more functionality to be components this kind of stuff could become more brittle.

app/models/select_facet.rb
alicebartlett: thanks tommy!

alicebartlett:
@tommyp @dsingleton OK, I've fixed the failing test with an equivalent action.

tommyp:
🎉🎉🎉

bradleywright:
:+1: :dancer: :8ball:

PR #212, opened 2015-06-02
Facet options from Rummager

tommyp:
This PR allows Finder Frontend to pull the facet options from Rummager when they are no `allowed_values` for a facet. It also modifies some code in the `SelectFacet` to use hashes as we are now using the component for the option_select which expects a hash instead of an object.

Work mostly completed by @dsingleton. I only swooped in at the end in an attempt to steal all the credit.

[Ticket](https://trello.com/c/dw31ZV0Y/231-add-filter-options-to-finder-of-finders).

Relies on https://github.com/alphagov/govuk-content-schemas/pull/78 being merged first to pass.

### DIFF
--- a/app/models/facet_collection.rb
+++ b/app/models/facet_collection.rb
@@ -20,6 +20,12 @@ class FacetCollection
     end
   end
 
+  def options=(options_hash)
+    filters.each do |facet|
+      facet.options = options_hash[facet.key] if options_hash[facet.key]
+    end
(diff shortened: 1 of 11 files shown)


tommyp:
As discussed offline with @evilstreak we're going to close this and do the same work by branching off master now that https://github.com/alphagov/finder-frontend/pull/215 has merged.

PR #215, opened 2015-06-03
Separate data fetching from presentation

evilstreak:
Follow up to #209, now with added specs.

## Aim of the change

This PR separates the application logic responsible for fetching data from APIs from the logic for presenting that data to the user.

The flow of data goes like:

- `FindersController` makes a call to a `FinderApi` object with the base path and user parameters
- The API returns a single data structure representing the finder and its results
- The controller passes this data into a `FinderPresenter`
- The controller passes the finder presenter into a `ResultSetPresenter`
- The controller renders the view with both presenters

Ideally the controller would only instantiate a `FinderPresenter`, and not also have to instantiate a `ResultSetPresenter`, but that's outside the scope of this PR.

## Reviewing the changes

This is best reviewed by commit by commit.

There are four commits with the major changes:

- d271df911266ab60e22b07642ea64a445e95185f adds `FinderApi`
- ad93de75e654c289c9670ecb84f3b623633a280c uses the new `FinderApi`
- b9b619aa79874bc7dd45d51f8903f67a053d98dc extracts query building logic from `FinderPresenter`
- 82395fe2a8c5ffbb6f5fd734f44cd3522d3b356c adds a `FilterQueryBuilder`

The rest of the commits are more contained refactoring and deletion of unused code.


app/lib/filter_query_builder.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+private
+  attr_reader :facets, :user_params
+
+  def filters
+    @filters ||= facets.select(&:filterable).map { |f| build_filter(f) }
+  end
+
+  def build_filter(facet)
```
tommyp: Is it wise to pass in `facet` here if we're still referring to the metadata as facets, right?

app/lib/filter_query_builder.rb
```diff
@@ -0,0 +1,96 @@
...
+      'text' => TextFilter,
+    }.fetch(facet.type)
+
+    params = user_params.fetch(facet.key, nil)
+
+    filter_class.new(facet, params)
+  end
+
+  class Filter
+    def initialize(facet, params)
```
tommyp: Same as above.

app/lib/finder_api.rb
```diff
@@ -0,0 +1,48 @@
+require 'delegate'
+
+class FinderApi
+  def initialize(content_api:, search_api:)
+    @content_api = content_api
```
tommyp: Shouldn't this be `content_store`, not `content_api`?

evilstreak: Updated.

app/lib/filter_query_builder.rb
evilstreak: As discussed IRL, the data structure being passed in here is one of the "facets" from the content item, so I think `facet` is fine.

PR #216, opened 2015-06-04
Get facet options from rummager

tommyp:
If a facet of type `text` has no `allowed_values` and is `filterable` then we want to query Rummager for the options for that facet. 

This PR adds a `FacetQueryBuiler` which takes a list of facets and decides if they need to be dynamic and returns a hash of the keys with the value of `"1000,order:value.title"`. This is then used in the `SearchQueryBuilder` with all the other params to build the query for Rummager.

The result which is returned then modifies the Content Item before passing it to the `FinderPresenter` which then constructs all of the facets through `FacetCollection`.

This PR also does a little refactoring with how the total is passed around. Interesting stuff is mostly 35bfce9 and 966e362. [Ticket](https://trello.com/c/dw31ZV0Y/231-add-filter-options-to-finder-of-finders).

Spiritual successor to https://github.com/alphagov/finder-frontend/pull/212 so shout out to @dsingleton for the first commit in this and doing most of the work on that previous PR.

### DIFF
--- a/app/lib/facet_query_builder.rb
+++ b/app/lib/facet_query_builder.rb
@@ -1,0 +1,32 @@
+class FacetQueryBuilder
+  def initialize(facets:)
+    @facets = facets
+  end
+
+  def call
+    dynamic_facets.reduce({}) { |query, facet|
+      # TODO title will only work for Orgs, this key will need changed for
+      # other dynamic facets
+      #
+      # "1000,order:value.title" is specifying that we want 1000 results back
+      # which are ordered by the title attribute of each value (option)
+      # that is returned
+      query.merge(facet.key => "1000,order:value.title")
+    }
+  end
+
+private
+  attr_reader :facets
+
+  def dynamic_facets
+    text_facets.select { |f| f.allowed_values.blank? }
+  end
+
+  def text_facets
+    filterable_facets.select { |f| f.type == "text" }
+  end
+
+  def filterable_facets
+    facets.select(&:filterable)
+  end
+end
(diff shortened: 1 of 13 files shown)


tommyp:
Paired with @evilstreak on this.

evilstreak:
Looks good to me, but since I paired on it I think @dsingleton should review this.

edds:
"1000,order:value.title" <- what does that mean? What is 1000?

tommyp:
Added a comment explaining the magic number of `1000,order:value.title`.

app/lib/facet_query_builder.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def call
+    dynamic_facets.reduce({}) { |query, facet|
+      # TODO title will only work for Orgs, this key will need changed for
+      # other dynamic facets
+      #
+      # "1000,order:value.title" is specifying that we want 1000 results back
+      # which are ordered by the title attribute of each value (option)
+      # that is returned
+      query.merge(facet.key => "1000,order:value.title")
```
dsingleton: (not a blocker, pedantry ahoy)

This is a good comment :+1:, but a lot of what it explains could be expressed in code too, something like:
```
num_facets = 1000
order_field = 'value.title'
query.merge(facet.key => "#{num_facets},order:#{order_field}")
```

(if we had more facet key/val pairs it might be worth making that a hash, but we just have the one for now, so it didn't seem worth overcomplicating it)

Personally I find it more readable, and it reduces the chance a comment would go out of sync with the code it's describing.

dsingleton:
Otherwise, LGTM. I'd merge, but I wrote quite a bit of the original, so i'm not sure thats kosher. ¯\_(ツ)_/¯

dsingleton:
Agreed with @tommyp it's been seen by enough eyes, so merged.

PR #219, opened 2015-06-25
Add print styles so finders look better printed

edds:
Hide the sidebar navigation as it isn't any use when printed, replicate
the visual presentation for the metadata and for the document rows.

Before/After:
screen shot 2015-06-25 at 12 08 35

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -1,0 +1,50 @@
+@import "colours";
+@import "shims";
+
+.filter-form,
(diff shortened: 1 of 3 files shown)


jamiecobbett:
Not qualified to review the code, but the results are :heart_eyes:

alicebartlett:
Yeesh! What an improvement. Good job @edds.

dsingleton:
:heart_eyes: :heavy_multiplication_x: :100:

PR #220, opened 2015-06-26
Ensure date-help-text id is unique

robmckinnon:
* Ensure unique date-help-text id is created when `_date_facet` is used more than once in a finder.
* Without this `slimmer` raises exception that id has been used more than once.


tommyp:
Thanks for this PR. Good catch! :+1:

PR #222, opened 2015-07-01
Track filtering

edds:
So that we know how people use the filters on finder pages add a page
track event after we have updated the URL.


benilovj:
Should this have a jasmine spec?

edds:
Now with added tests.

benilovj:
:+1:

fofr:
Can the virtual pageviews be added to https://gov-uk.atlassian.net/wiki/display/GOVUK/Analytics+on+GOV.UK?

PR #224, opened 2015-07-13
Pagination

tommyp:
Within certain Finders, we don't want the documents per page to be all of them (or 1000). This PR allows us to specify a `default_documents_per_page` as part of the content item and have Finder Frontend build the pagination query within the `SearchQueryBuilder` and render the Previous and Next component. [Ticket](https://trello.com/c/NfxTcDci/273-paginate-finder-frontend).

Now with AJAX! This happens by returning the partial as part of the hash and having mustache render it which means it respects the query being added.

Screenshot:
screen shot 2015-06-30 at 12 42 15

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -46,6 +46,12 @@ private
     content_item.details.results = search_response.fetch("results")
     content_item.details.total_result_count = search_response.fetch("total")
 
+    content_item.details.pagination = build_pagination(
+      content_item.details.documents_per_page,
+      search_response.fetch('start'),
+      search_response.fetch('total')
+    )
+
     search_response.fetch("facets", {}).each do |facet_key, facet_details|
       facet = content_item.details.facets.find { |f| f.key == facet_key }
       facet.allowed_values = allowed_values_for_facet_details(facet_details) if facet
@@ -63,5 +69,14 @@ private
         value: value.fetch("slug", ""),
       )
     }
+  end
+
+  def build_pagination(documents_per_page, start_offset, total_results)
+    if documents_per_page
+      OpenStruct.new(
+        current_page: (start_offset / documents_per_page) + 1,
+        total_pages: (total_results / documents_per_page.to_f).ceil,
+      )
+    end
   end
 end
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -47,7 +47,7 @@ private
     content_item.details.total_result_count = search_response.fetch("total")
 
     content_item.details.pagination = build_pagination(
-      content_item.details.documents_per_page,
+      content_item.details.default_documents_per_page,
       search_response.fetch('start'),
       search_response.fetch('total')
     )
--- a/spec/presenters/result_set_presenter_spec.rb
+++ b/spec/presenters/result_set_presenter_spec.rb
@@ -2,19 +2,29 @@ require 'spec_helper'
 
 RSpec.describe ResultSetPresenter do
 
-  subject(:presenter) { ResultSetPresenter.new(finder)}
+  subject(:presenter) { ResultSetPresenter.new(finder, view_context)}
 
   let(:finder) do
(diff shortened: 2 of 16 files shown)


spec/presenters/result_set_presenter_spec.rb
```diff
@@ -2,20 +2,28 @@
...
       results: results,
       document_noun: document_noun,
-      total: 2,
+      total: 20,
       facets: [ a_facet, another_facet ],
       keywords: keywords,
       atom_url: "/a-finder.atom",
+      default_documents_per_page: 10,
+      values: {},
+      pagination: OpenStruct.new(
```
mikejustdoit: I realise this is already a common thing in this file, but instead of an `OpenStruct` a lot of the time we can use a `double`.

And either way, it's fine to break these nested ones out into their own variables, too.

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -73,10 +81,10 @@
 
   let(:keywords){ '' }
   let(:document_noun){ 'case' }
-  let(:total) { 2 }
+  let(:total) { 20 }
 
   let(:results) do
-    OpenStruct.new({ total: total, documents: [ document ] })
+    OpenStruct.new({ total: total, documents: (1..20).map {|n| document }})
```
mikejustdoit: Maybe this range could be more understandable (and robust) as `1..total`?

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -97,6 +105,21 @@
...
+              <span class="pagination-label">1 of 2</span>
+            </a>
+          </li>
+          <li class="next-page">
+            <a href="next-page" rel="next">
+              <span class="pagination-part-title">Next page</span>
+              <span class="pagination-label">2 of 2</span>
+            </a>
+          </li>
+      </ul></nav>' }
```
mikejustdoit: I'm not sure if actually having this markup here affects the spec? If not, it might be clearer to just return a smaller string, like `"<nav></nav>"`?

app/lib/finder_api.rb
```diff
@@ -46,6 +46,12 @@ def augment_content_item_with_results(content_item, search_response)
     content_item.details.results = search_response.fetch("results")
     content_item.details.total_result_count = search_response.fetch("total")
 
+    content_item.details.pagination = build_pagination(
+      content_item.details.default_documents_per_page,
+      search_response.fetch('start'),
+      search_response.fetch('total')
+    )
```
mikejustdoit: I was wondering if instead of adding the pagination to the content_item's details, the FinderApi could return the content_item and the pagination data as a kinda `return [content_item, pagination]`... But I haven't worked with finders much, so maybe the pagination in the `details` already makes enough sense.

tommyp: Previously this was the case, we were passing about a bunch of different things in order to build a Finder but it got hella confusing. It was decided a while ago (https://github.com/alphagov/finder-frontend/pull/215) that instead we should fetch the Content Item, use that to build a search query, use the results of the search query to modify the Content Item and then passed this modified Content Item about.

mikejustdoit:
The commit message for "Feature for Pagination" sounds like it's describing 2 changes, if I've read that right and splitting them up is possible, that's a nice to have.

tommyp:
Sort of, but the other change without the Pagination test doesn't really have much context.

dsingleton:
Overall this looks good :+1: 

Still unhappy with the hackyness of rendering a component from the view for use in the client side update, but can't think of a better way. Hopefully we'll be able to revisit it when we have a bit more experience with client-side use of components.

tommyp:
Yeah. I agree that it's not nice and moves behaviour that would be expected in the view to a weird place but I don't know of a better way of doing it.

dsingleton:
:page_facing_up: :page_facing_up: :page_facing_up: :tada:

PR #228, opened 2015-07-31
Allow Finders to display Alpha Banner

tommyp:
This commit allows `finders#show` to be displayed with the Alpha banner
in the same way we can currently display the Beta banner.


dsingleton:
LGTM. 

Slight tangent, but I wonder if phase state is something that might become (an optional) part of the standard content item definition, as it's not really format specific.

PR #230, opened 2015-08-12
Modify alpha beta boolean methods

tommyp:
With the new phase attribute on the base content item we can modify the
`alpha?` and `beta?` methods to check the contents of this and use that
to change the frontend.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -25,11 +25,11 @@ class FinderPresenter
   end
 
   def alpha?
-    content_item.details.alpha
+    content_item.phase == 'alpha'
   end
 
   def beta?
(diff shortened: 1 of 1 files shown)


jackscotti:
For what else would you `phase` for?

app/presenters/finder_presenter.rb
```diff
@@ -184,4 +184,8 @@ def is_external?(href)
       "www.gov.uk" != host
     end
   end
+
+  def phase
```
jackscotti: For what else would you use this method? I don't think this is necessary

tommyp: Ah yeah. I forgot to delete that before committing.

PR #231, opened 2015-08-17
Add meta description and rel="alternate" tags to finders

jackscotti:
To improve the search result snippets show on external search engines.

Trello: https://trello.com/c/zFUIJFaJ/287-add-meta-description-tags-to-finders


edds:
I wonder if that is an appropriate `rel="alternate"` for a finder. In my head the alternate version would be the contents of the finder rather than the structure of how the finder is built as that is the main part of the page. However, I don't hold that belief strongly so happy to accept that this is appropriate.

jackscotti:
@edds thanks for the feedback. Before opening the PR I've asked @alext whether this was a reasonable thing to add.

What you says makes sense though. Happy to delete the last commit.

(I got most of this code from this other PR in collections: https://github.com/alphagov/collections/pull/133)

bradleywright:
The emerging pattern is to show the content-store JSON for the document you're looking at, so I'm okay with this PR as-is.

app/views/finders/_finder_meta.html.erb
```diff
@@ -0,0 +1,4 @@
+<% if finder.description %>
+  <meta name="description" content="<%= finder.description %>">
+<% end %>
+<link rel="alternate" type="application/json" href="/api/content<%= finder.slug %>">
```
alext: Does the slug include the leading `/`?

tijmenb: This is the `base_path` of the content_item, that always has the leading slash right?

https://github.com/alphagov/finder-frontend/blob/master/app/presenters/finder_presenter.rb#L20

jackscotti: It should always include the leading slash.

alext: Gah!  This inconsistent naming isn't great.  Out of scope for this PR though.

tijmenb:
:+1:

PR #240, opened 2015-09-28
Use standard Airbrake configuration

tijmenb:
This is the default GOV.UK Airbrake config now, which reads from ENV instead of relying on overwriting files. It's taken from

https://github.com/alphagov/govuk-rails-app-template

It's safe to deploy because until the corresponding change has been made in alphagov-deployment, this file still gets overwritten.


jamiecobbett:
:+1:

tijmenb:
@jamiecobbett If you like it can you merge it? :)

PR #249, opened 2016-03-07
Update gds-api-adapters

tijmenb:
This will give us better error reporting and will start passing down the `govuk_original_url` header, which will give us more visibility into the data flow on GOV.UK.

https://trello.com/c/JVZPH4S9


jackscotti:
:+1:  :ship:  it when green

PR #256, opened 2016-10-21
Default public_timestamp to nil when not present

carvil:
We are moving old "finder-like" pages over to `finder-frontend`. One of
these finders is the people finder from Whitehall.

The Person records in Whitehall don't have a `public_timestamp` field
because they are not an edition. This means that we don't send that
information to Rummager. When `finder-frontend` fetches those records
from rummager and tries to fetch the `public_timestamp`, the
initialization of a `Document` fails.

This commit makes sure we default to `nil` when fetching the
`public_timestamp`. The reason `nil` is fine is because the default
ordering on people is by `title` (i.e. the minister's names) rather than
by `public_timestamp`.

Example person record: https://www.gov.uk/api/search.json?filter_link=/government/people/theresa-may

Part of https://trello.com/c/reCVyzpg/259-migrate-the-people-page-to-a-finder

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -1,0 +1,18 @@
+require "spec_helper"
+
+describe Document do
+  describe "initialization" do
+    it 'defaults to nil without a public timestamp' do
+      rummager_document = {
+        title: 'A title',
+        link: 'link.com'
+      }
+      finder = double(
+        'finder', date_metadata_keys: [], text_metadata_keys: []
+      )
+      document = described_class.new(rummager_document, finder)
+
+      expect(document.public_timestamp).to be_nil
+    end
+  end
+end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,17 +1,17 @@
 class Document
   attr_reader :title, :public_timestamp, :is_historic, :government_name
 
-  def initialize(attrs, finder)
-    attrs = attrs.with_indifferent_access
-    @title = attrs.fetch(:title)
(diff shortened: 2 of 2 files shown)


spec/models/document_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      }
+    end
+    let(:finder) do
+      double('finder', date_metadata_keys: [], text_metadata_keys: [])
+    end
+    let(:document) { described_class.new(attrs, finder) }
+
+    context 'without a public timestamp' do
+      it 'defaults to nil' do
+        expect(document.public_timestamp).to be_nil
```
tijmenb: Can we inline some of the `let`s here? I think it would be good to show that the rummager document doesn't have a `public_timestamp` inside the `it` block.

spec/models/document_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    let(:attrs) do
+      {
+        title: 'A title',
+        link: 'link.com'
+      }
+    end
+    let(:finder) do
+      double('finder', date_metadata_keys: [], text_metadata_keys: [])
+    end
+    let(:document) { described_class.new(attrs, finder) }
```
tijmenb: Would renaming `attrs` to `rummager_document` make sense?

carvil: @tijmenb I was following the same naming as in the main class so we could follow things along. I will rename in both places.

tijmenb: 👍 great improvement

PR #259, opened 2016-11-07
Change breadcrumbs to use navigation helpers

rubenarakelyan:
This commit adds the govuk_navigation_helpers and changes the finder view to use its output for breadcrumb generation. This allows finders to have breadcrumbs rather than just hardcoding in “Home”.

Paired with @klssmith

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -7,7 +7,9 @@ class FindersController < ApplicationController
     @results = ResultSetPresenter.new(finder, view_context)
 
     respond_to do |format|
-      format.html
+      format.html do
+        @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(raw_finder)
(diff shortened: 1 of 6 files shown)


app/controllers/finders_controller.rb
```diff
@@ -7,7 +7,10 @@ def show
     @results = ResultSetPresenter.new(finder, view_context)
 
     respond_to do |format|
-      format.html
+      format.html do
+        @content_item = Services.content_store.content_item(finder_base_path).to_hash
```
tijmenb: We're already doing a request to get the content item (somewhere in `FinderApi`). We should reuse that so we don't do two requests.

rubenarakelyan: Yup, found it. It's `raw_finder`.

tijmenb:
Nice!

PR #264, opened 2016-11-25
Add topicalness facet type

MatMoore:
This is a new kind of facet to dynamically split topical events into open or closed based on the end date.

This will look something like this when the finder itself is published:
screen shot 2016-11-24 at 15 05 14

I've also removed what seems like an unused field from the presenters, and fixed some broken tests. This should not affect existing finders (I hope).

Trello: https://trello.com/c/oVfSAO9W/311-migrate-policy-areas-topical-events-index-to-a-finder
Matching schema change: https://github.com/alphagov/govuk-content-schemas/pull/441
Finder: https://github.com/alphagov/whitehall/pull/2855

### DIFF
--- a/app/lib/filter_query_builder.rb
+++ b/app/lib/filter_query_builder.rb
@@ -23,6 +23,7 @@ private
     filter_class = {
       'date' => DateFilter,
       'text' => TextFilter,
+      'topical' => TopicalFilter,
     }.fetch(facet.type)
 
     params = user_params.fetch(facet.key, nil)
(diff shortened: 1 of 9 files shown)


app/lib/filter_query_builder.rb
```diff
@@ -23,6 +23,7 @@ def build_filter(facet)
     filter_class = {
       'date' => DateFilter,
       'text' => TextFilter,
+      'topical' => TopicalFilter,
```
tijmenb: https://github.com/alphagov/finder-frontend/blob/master/docs/finder-content-item.md#type needs updating for this I think.

app/lib/filter_query_builder.rb
```diff
@@ -86,4 +87,41 @@ def value
...
+
+      open = params.include?(facet.open_value.value)
+      closed = params.include?(facet.closed_value.value)
+
+      if open && !closed
+        open_value
+      elsif closed && !open
+        closed_value
+      else
+        # with both selected, the filter is not used
```
tijmenb: This should probably read "with both or neither selected".

MatMoore:
@tijmenb Have addressed your comments, except for the docs, which I'll add in a separate PR.

If this and https://github.com/alphagov/whitehall/pull/2855 are good to merge I can deploy the topical events finder this morning.

Issue #137, opened 2014-12-16
Invalid date feedback


tommyp:
Right now, if a User enters an invalid date, nothing happens at all. It would be good to introduce a validation error to tell the user that the value entered isn't a date.

tijmenb:
Now tracked here: https://trello.com/c/oC0Y60Uz/46-finders-provide-feedback-for-invalid-input

PR #266, opened 2016-11-28
[DISCUSS] Hide and report blank facet values

rubenarakelyan:
Currently, visiting a finder with a URL that filters on a facet value that is not synced between rummager and the publishing-api results in a blank facet value being displayed and checked.

This commit excludes facet values with no title from being displayed, and also raises an Airbrake error so that these problems are logged to be fixed.

Trello: https://trello.com/c/JMFQcEj6/337-finder-frontend-displays-blank-values-in-facets

Before:

screen shot 2016-11-28 at 15 33 27

After:

screen shot 2016-11-28 at 15 33 47

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -20,6 +20,9 @@ private
 
   attr_reader :base_path, :filter_params
 
+  class EmptyFacetValueError < StandardError
+  end
+
   def fetch_content_item
     Services.content_store.content_item!(base_path)
   end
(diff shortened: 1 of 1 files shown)


rubenarakelyan:
Initial ideas - there are a few ways of fixing this:

*Hide the empty items and raise an error*

Pros:

1. Users don't see one or more blank facet values, which can be confusing since it's not obvious what is being filtered
2. The finder still works if others filters are used

Cons:

1. The total number of found documents is not displayed even if documents are returned
2. If no documents are returned, the initial page is empty of all content which may be confusing

* Return a 404 page *

Pros:

1. Makes it obvious that we can't find what the user is looking for

Cons:

1. Users see a 404 page and can't use the other finder filters

* Keep the facet values, but replace their labels with something *

Pros:

1. If the user visits the URL directly then the existing filtering still works as intended, but there are no blanks
2. These values could be hidden if they are not directly in the URL (so users visiting the finder don't see them unless they are visiting with a URL that already filters by them)

Cons:

1. Multiple "unknown" values may be confusing as to what is being searched for

MatMoore:
Me and @rubenarakelyan talked about a very similar use case where the user filters on a value that doesn't exist in the `allowed_values`. Despite the name, this is actually allowed, but you get 0 results. I'm not sure if it's possible for this to mangle the facet labels or the sentence above the results, like with the inconsistent data scenario (I thought it was but I couldn't reproduce it).

Example: https://www.gov.uk/cma-cases?keywords=auction&case_type%5B%5D=ca99-and-civil-cartels&closed_date%5Bfrom%5D=&closed_date%5Bto%5D=

This could happen to real users if they bookmark a search and we then change the allowed_values for a facet, eg combining a bunch of document types together.

tijmenb:
There's a longstanding issue that is related: https://github.com/alphagov/finder-frontend/issues/137. We have no way to let the user know that their search query / URL is incorrect.

For example:

https://www-origin.publishing.service.gov.uk/aaib-reports?date_of_occurrence=what

Gives us this error:

https://errbit.publishing.service.gov.uk/apps/54088bad0da1152a9f001417/problems/583c75696578630b4f430200

rubenarakelyan:
I'm going to close this PR following a discussion with @tijmenb. As per https://trello.com/c/JMFQcEj6/337-finder-frontend-displays-blank-values-in-facets, I'll create a rake task in rummager to catch un-synced data and report it on a regular basis so that we're catching this closer to the source.

PR #268, opened 2016-12-15
Clean invalid UTF-8 characters

rubenarakelyan:
This commit adds a class that removes invalid UTF-8 characters from strings. The class is used when parsing keywords passed to finder-frontend in order to fix `ArgumentError: invalid byte sequence in UTF-8` exceptions.

### DIFF
--- a/app/lib/utf8_cleaner.rb
+++ b/app/lib/utf8_cleaner.rb
@@ -1,0 +1,24 @@
+# In some cases requests come in to this app with invalid UTF-8 characters
+# in the query string, for example:
+#
+#   keywords=%c0%2e%c0%2e
+#
+# We want to clean up these strings to remove the invalid characters
+class UTF8Cleaner
+  def initialize(string_to_be_cleaned)
+    @string_to_be_cleaned = string_to_be_cleaned
+  end
+
+  def cleaned
+    return unless @string_to_be_cleaned
+
+    # Doesn't actually do any conversion but removes invalid characters
+    @string_to_be_cleaned.encode(
+      'UTF-8',
+      'UTF-8',
+      invalid: :replace,
+      undef: :replace,
+      replace: ''
+    ).presence
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,2 +1,6 @@
 AllCops:
   TargetRubyVersion: 2.3
+  Exclude:
+    # Exclude the UTF8Cleaner test because it includes invalid UTF-8 characters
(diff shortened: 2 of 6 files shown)


app/lib/utf8_cleaner.rb
```diff
@@ -0,0 +1,26 @@
...
+    if !@string_to_be_cleaned.nil?
+      # Returns nil if the cleaned string is empty
+      @string_to_be_cleaned.encode(
+        'UTF-8',
+        'binary',
+        invalid: :replace,
+        undef: :replace,
+        replace: ''
+      ).presence
+    else
```
tijmenb: Is govuk-lint running on this project? Couple of things it should've caught.

MatMoore:
Can you write a couple of tests that use multibyte unicode characters?

How does rails set the encoding of it's parameter strings? It seems like if the string is not currently binary you would strip out extra characters here:

```
irb(main):001:0> "£".encoding
=> #<Encoding:UTF-8>
irb(main):002:0> "£".encode("UTF-8", "binary")
Encoding::UndefinedConversionError: "\xC2" from ASCII-8BIT to UTF-8
```

app/lib/utf8_cleaner.rb
rubenarakelyan: https://github.com/alphagov/finder-frontend/pull/269 :)

rubenarakelyan: Lint is now in and I've fixed up what it caught

rubenarakelyan:
I've added a test with an emoji character to test handling of multi-byte UTF-8 characters.

I also changed the encode method to go from UTF-8 to UTF-8 (i.e. no change of encoding), since it's the stripping out of invalid characters rather than the change of encoding that makes the real difference here.

rubenarakelyan:
I've also added the UTF8Cleaner in a couple of other places where keywords are taken from the query string to prevent more "invalid UTF-8" errors.

MatMoore:
No idea what it's used for, but there is also this rack middleware on alphagov, which rejects requests that aren't encoded in UTF-8: https://github.com/alphagov/invalid_utf8_rejector

rubenarakelyan:
Ah, I did not know about that gem! I'll close this PR and open a new one that just installs it for finder-frontend. The difference is that this PR strips out invalid characters but keeps the rest, whereas the gem rejects the entire request if it contains any such characters. I don't think this should matter since users won't normally be accessing pages like this - these requests are coming from automated bots.

PR #271, opened 2016-12-16
Render a 406 when an invalid format is requested

rubenarakelyan:
Currently, if the HTTP `accept` header sent with a request asks for a format other than HTML, JSON or Atom, an exception is raised. This commit catches the exception and serves an error with a 406 status code.


cbaines:
Would `406 Not Acceptable` be more appropriate?

rubenarakelyan:
Yes, good point - forgot about that error!

PR #272, opened 2016-12-16
Reject requests with invalid UTF-8 characters

rubenarakelyan:
This commit adds a gem that rejects requests that contain invalid UTF-8 characters. This prevents `ArgumentError: invalid byte sequence in UTF-8` exceptions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'shared_mustache', '~> 0.1.3'
 gem 'airbrake', '~> 4.0.0'
 gem 'chronic', '~> 0.10.2'
 gem 'govuk_navigation_helpers', '~> 2.0.0'
+gem 'invalid_utf8_rejector', '~> 0.0.3'
 
 group :assets do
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -9,6 +9,7 @@ gem 'shared_mustache', '~> 0.1.3'
 gem 'airbrake', '~> 4.0.0'
 gem 'chronic', '~> 0.10.2'
 gem 'govuk_navigation_helpers', '~> 2.0.0'
+gem 'invalid_utf8_rejector', '~> 0.0.1'
```
tijmenb: Any reason not to use the latest version?

tijmenb: https://rubygems.org/gems/invalid_utf8_rejector

rubenarakelyan: I looked in https://github.com/alphagov/invalid_utf8_rejector/blob/master/lib/invalid_utf8_rejector/version.rb and saw `0.0.1` - I'll update it in the Gemfile.

PR #273, opened 2016-12-16
Gracefully handle empty query string values

rubenarakelyan:
Currently, empty query string values that are defined to be dates (such as `public_timestamp[from]` and `public_timestamp[to]`) raise exceptions if they are requested with just the base name (as in, `public_timestamp` without `[from]` or `[to]`) and no value.

This commit:

1. Strips out empty query string values
2. Changes filtering facets to use filtered user input rather than taking directly from `params`

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -4,7 +4,7 @@ class FindersController < ApplicationController
   include GdsApi::Helpers
 
   def show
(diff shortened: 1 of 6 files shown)


rubenarakelyan:
I've moved `user_values` to be a method in `ResultSetPresenter` which is then called in the partials. Is this what you meant?

PR #276, opened 2016-12-19
Fix lint errors

rubenarakelyan:
This commit fixes lint errors in finder-frontend.


app/lib/filter_query_builder.rb
```diff
@@ -99,13 +102,11 @@ def value
         open_value
       elsif user_has_selected_closed && !user_has_selected_open
         closed_value
-      else
-        # with both or neither selected, the filter is not used
```
tijmenb: Can you move this explanation to the top of the method?

PR #278, opened 2016-12-20
Ensure Jenkins build parameters are set

suzannehamilton:
Jenkins build parameters do not seem to be set on the very first build of a new branch. env.PARAM_NAME returns null, and referring to them directly by PARAM_NAME throws groovy.lang.MissingPropertyException. This means that the first build of a new Jenkins branch always fails.

The parameters are set correctly on every subsequent build, whether it is triggered automatically by a branch push or manually in by a Jenkins user.

This is an extremely hacky solution - it checks for null parameters and makes sure that they are set to the defaults if so - but I haven't found anything better.

If anyone has any other ideas or can spot what I'm doing wrong with the parameters, please let me know! (@afda16?)

If not, I'll report this to Jenkins in case it's a broader issue with parameterized builds in multibranch pipelines.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,8 @@ REPOSITORY = 'finder-frontend'
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
 
+  DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+
   properties([
     buildDiscarder(
       logRotator(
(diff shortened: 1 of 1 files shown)


suzannehamilton:
I've raised this as a Jenkins pipeline issue since [other people have seen the same error](http://stackoverflow.com/questions/40782302/jenkinsfile-parameter-properties-not-configured-in-jenkins-server-at-initial-bra):

https://issues.jenkins-ci.org/browse/JENKINS-40574

PR #284, opened 2017-01-18
[DO NOT MERGE] Change handling of email alert titles

rubenarakelyan:
This commit splits the `title`, `short_name` and `description` attributes so that they can be separately generated, whereas before `title` was also used for `short_name` and `description` was empty when passed to the `email-alert-api`.

Since `title` and `short_name` have a 255 character limit enforced by GovDelivery, this commit adds logic to keep them under this limit: only the name of the finder and the number of topics chosen are shown, followed by an MD5 hash of all the topic names (for `title` only to keep it unique). `description` stores all the names of the chosen topics, which was previously stored in `title` and `short
_name`.

https://github.com/alphagov/email-alert-api/pull/136 makes changes to `email-alert-api` so that the separate `short_name` field is sent to GovDelivery.

Trello: https://trello.com/c/KTHRa4a0/422-fix-email-alert-api-errors-when-the-email-alert-short-name-is-longer-than-255-characters

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -50,6 +50,7 @@ private
       email_alert_api: email_alert_api,
       attributes: email_signup_attributes,
       subscription_list_title_prefix: content.details.subscription_list_title_prefix,
+      email_filter_name: content.details.email_filter_name,
       available_choices: available_choices,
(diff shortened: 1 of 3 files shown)


rubenarakelyan:
I'm closing this because it's becoming too complicated and the email team are looking at changing this all anyway. I'll put in a smaller change to get the CMA ones working.

PR #287, opened 2017-03-06
Add analytics meta tags to finder pages

suzannehamilton:
Add meta tag component to finder pages. This is in preparation for the education navigation release, which includes analysis of users' journeys through navigation and content pages. Finder pages (including policies) are part of the navigation, and using the analytics meta tags will let us track them as such.

I haven't added tests for the meta tags because I haven't seen any other projects do it, presumably because the tests would be fragile because they would have to make assumptions about what meta tags would be added. Is that right, or is it worth adding a test here?

https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -1,7 +1,7 @@
 class FinderPresenter
   include ActionView::Helpers::UrlHelper
 
-  attr_reader :name, :slug, :organisations, :keywords, :values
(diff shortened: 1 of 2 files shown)


suzannehamilton:
The meta tag that we're adding (for `finding` vs `thing` pages) has a default value, so we'd notice that these pages were showing up with the wrong tag.

PR #293, opened 2017-04-12
Disable cacheing in development env

davidbasalla:
Rummager request responses were being cached and caused surprising behaviour
in that newly added specialist documents would not show up in search results
after cacheing on the dev vm. Also disable cacheing for content store to be
consistent.

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -3,10 +3,16 @@ require 'gds_api/rummager'
 
 module Services
   def self.content_store
-    @content_store ||= GdsApi::ContentStore.new(Plek.find("content-store"))
+    @content_store ||= GdsApi::ContentStore.new(
+      Plek.find("content-store"),
+      disable_cache: Rails.env.development?
+    )
   end
 
   def self.rummager
-    @rummager ||= GdsApi::Rummager.new(Plek.find("search"))
+    @rummager ||= GdsApi::Rummager.new(
+      Plek.find("search"),
+      disable_cache: Rails.env.development?
+    )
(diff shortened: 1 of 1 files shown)


cbaines:
> Rummager request responses were being cached and caused surprising behaviour
> in that newly added specialist documents would not show up in search results
> after cacheing on the dev vm. Also disable cacheing for content store to be
> consistent.

I assume finder-frontend when running in the Rails production environment also exhibits this behaviour where newly added specialist documents do not appear, if so, I'm not sure I see sufficient justification to make the behaviour that most people see when developing the application different? If the behaviour is surprising to users regardless of the Rails environment, then just making it less surprising for those developing the application doesn't seem to me to be a very useful approach to improving the application itself, or the process of developing it?

davidbasalla:
@cbaines that's a fair point! Would it be feasible to turn off gds-api-adapter caching for Rummager in this app for _all_ environments (which would remove the surprising behaviour of documents not showing up immediately altogether)? Will discuss more with the team tomorrow as we came across this as part of our Blitz work...

cbaines:
> @cbaines that's a fair point! Would it be feasible to turn off gds-api-adapter caching for Rummager in this app for all environments (which would remove the surprising behaviour of documents not showing up immediately altogether)? Will discuss more with the team tomorrow as we came across this as part of our Blitz work...

The immediate trade-off is the number of requests to rummager vs the latency in responses. Unfortunately, rummager is an app I've never touched, so I'm not a very useful person to ask about this.

PR #294, opened 2017-05-10
Upgrade rails

mcgoooo:
General upgrade to rails 5, includes a lot of linting fixes
& syntax changes that don't warrant exmplaining. Also removed
the comments from the production.rb, these are standard rails
comments when you make a new repo, and make it hard to read.

assets

Rails was recommending a seperate initialiser for compiled assets

Unsafe hash

We have moved to using to_unsafe_hash due to rails strong parameters, no
longer inheriting from hash. Rails strong parameters as far as i can
tell are somewhat moot here, as we dont use it to serialize to activemodel
anywhere. Any app further down the line should have it's own checking in, but
this particular app is read only

Rspec in the development group & rspec/autotum

You need it in the dev group so it can ovverride the default task, otherwise
rspec tests do not run unless you specifially set rails_env=test.
also added rails controller testing, this is needed for rspec.
rspec autorun is also no longer needed.

Invalid utf-8 characters

pretty sure this is already taken care of by action dispatch
https://github.com/rails/rails/blob/34fe2a4fc778d18b7fe6bdf3629c1481bee789b9/actionpack/lib/action_dispatch/request/utils.rb#L36

this seems to only be on legacy repos as well


boffbowsh:
This looks ok, can you deploy it to integration to test it please?

mcgoooo:
@boffbowsh it's on integration, all looks ok

config/environments/production.rb
```diff
@@ -1,89 +1,19 @@
...
-
-  # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.action_controller.asset_host = "http://assets.example.com"
-
-  # Precompile additional assets.
-  # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
-  config.assets.precompile += %w(
-    application-ie6.css
-    application-ie7.css
-    application-ie8.css
```
fofr: Did this have any affect in IE6-8?

PR #299, opened 2017-08-10
Set up Policies finder A/B test

leenagupte:
Trello card: https://trello.com/c/tztMiJoy

Related PRs:
* https://github.com/alphagov/fastly-configure/pull/37
* https://github.digital.cabinet-office.gov.uk/gds/cdn-configs/pull/159

Related Zendesk ticket: https://govuk.zendesk.com/agent/tickets/2303129

## Motivation

There are two policy finders on GOV.UK. The [normal policies finder](https://www.gov.uk/government/policies) and the [all policy content finder](https://www.gov.uk/government/policies/all).

The all policy content finder is hidden to regular users. You have to know the URL to be able to find it. 

This A/B test will provide evidence to help determine which finder provides users with a better journey to find the policy content they are looking for.


spec/controllers/finders_controller_spec.rb
```diff
@@ -90,5 +91,87 @@
...
+        setup_ab_variant("PolicyFinderTest", "B")
+
+        get :show, params: { slug: 'government/policies' }
+        finder_slug = subject.instance_variable_get(:@results).finder.slug
+
+        expect(response.status).to eq(200)
+        expect(finder_slug).to eq("/government/policies/all")
+      end
+
+      it "directs users in both groups to a child policy finder" do
```
brenetic: This might be better worded to describe the case when a user goes directly to a finder that is not part of the a/b test. We had initially thought that you were trying to do an a/b/c test (which if you'd like to write the fastly VCL to do that please do!! :)  )

leenagupte: Hahaha! No, just an A/B test! I'll reword the test.

PR #312, opened 2017-09-26
Add component guide and create date filter component

fofr:
* Add first component, styles, partial and tests
* Move date facet into date filter component
* Enable aria-controls dynamically for component flexibility

## In finder

screen shot 2017-09-26 at 18 22 32

## Component guide

date filter component - finder frontend component guide 20170926


fofr:
https://finder-frontend-pr-312.herokuapp.com/component-guide/date-filter 😀

fofr:
@MatMoore Yes. If the JS is disabled or fails for whatever reason, the inputs won't be controlling another element, so the attributes would be invalid.

PR #314, opened 2017-10-16
[DO NOT MERGE] Split site search mustache up

dwhenry:
Extra the organisation filter into its own file in preparation of replacing with the select component.
Replace the search bar with the search component.


app/controllers/search_controller.rb
```diff
@@ -0,0 +1,60 @@
+require 'gds_api/helpers'
+
+class SearchController < ApplicationController
+  include GdsApi::Helpers
+  # before_filter :set_expiry
```
brenetic: ^^

dwhenry:
This is unlikely to be completed as we have pivoted to other stories. 

The effort to get to this point was low, however, with the failing tests, it would require too much to try and complete now.

fofr:
I've cherry picked 3c5522c into https://github.com/alphagov/finder-frontend/tree/search-frontend-tweaks

PR #317, opened 2017-10-26
Add format weighting ab test

brenetic:
We are going to be running an A/B test to test changes to format
weighting.

Merge after https://github.com/alphagov/finder-frontend/pull/318

Paired with @suzannehamilton 

https://trello.com/c/JJ3mIOP1

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -5,8 +5,13 @@ class SearchController < ApplicationController
   before_action :set_expiry
   before_action :remove_search_box
 
+  helper_method :search_ab_test_variant
+  after_action :set_search_ab_test_response_header
+
   rescue_from GdsApi::BaseError, with: :error_503
   layout 'search-application'
+
+  SEARCH_AB_TEST_DIMENSION = 42
 
   def index
     search_params = SearchParameters.new(params)
@@ -24,7 +29,9 @@ class SearchController < ApplicationController
     if search_params.no_search? && params[:format] != "json"
       render(action: 'no_search_term') && return
     end
-    search_response = SearchAPI.new(search_params).search
+
+    variant = search_ab_test_variant.variant_name
+    search_response = SearchAPI.new(search_params, format_weighting: variant).search
 
     @search_term = search_params.search_term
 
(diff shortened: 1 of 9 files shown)


suzannehamilton:
@MatMoore: That makes sense. I've updated the names. 👍

suzannehamilton:
I've just moved the A/B meta tag from `<body>` to `<head>`. The tests will fail until #318 is merged, because it fixes the header rendering.

app/controllers/search_controller.rb
```diff
@@ -24,7 +29,9 @@ def index
     if search_params.no_search? && params[:format] != "json"
       render(action: 'no_search_term') && return
     end
-    search_response = SearchAPI.new(search_params).search
+
+    variant = search_ab_test_variant.variant_name
+    search_response = SearchAPI.new(search_params, format_boosting: variant).search
```
dwhenry: we have called the variant `format_boosting` and the test `FormatWeighting` can we get these names to match?

spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        "suggested_queries": []
+      }|
+
+    stub_request(:get, /search.json/).to_return(status: 200, body: rummager_response, headers: {})
+  end
+
+  context "Format weighting A/B test" do
+    it "should set the correct A variant tags" do
+      with_variant FormatWeighting: "A" do
+        get :index, params: { q: "cheese" }
```
dwhenry: This test doesn't appear to have a test? or at least not one that relates to the test name `should set the correct A variant tags`

spec/spec_helper.rb
```diff
@@ -17,6 +17,10 @@
 # If you are not using ActiveRecord, you can remove this line.
 ActiveRecord::Migration.check_pending! if defined?(ActiveRecord::Migration)
 
+GovukAbTesting.configure do |config|
+  config.acceptance_test_framework = :active_support
+end
```
dwhenry: this might make more sense in a `spec/support/govuk_ab_testing.rb` file rather than here.

dwhenry:
Also, the description doesn't reflect the code which is confusing and we should mention the dependancy on https://github.com/alphagov/finder-frontend/pull/318/files being merged

spec/controllers/search_controller_spec.rb
brenetic: `with_variant` asserts meta tags have been set correctly amongst other things

PR #319, opened 2017-11-07
Initial Dockerfile

stephenharker:
Docker image should build but may need some tweaks.


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+ADD .ruby-version $APP_HOME/
+RUN bundle install
+
+ADD . $APP_HOME
+
+RUN GOVUK_APP_DOMAIN=www.gov.uk RAILS_ENV=production bundle exec rails assets:precompile
+
+HEALTHCHECK CMD curl --silent --fail localhost:$PORT || exit 1
```
RyanMacG: Not sure this will work as there doesn't seem to be a root route. `/cma-cases` is used in the puppet setup https://github.com/alphagov/govuk-puppet/blob/560c2f4f739f4d369eacb66a5a9b1e344527969a/modules/govuk/manifests/apps/finder_frontend.pp#L40

kevindew: ah yes! I didn't think of that. I'll add a healthcheck route in.

kevindew:
@RyanMacG I've updated this now as per your comment.

PR #342, opened 2017-12-12
Fix dep warnings

h-lame:
When doing #331 I spotted loads of deprecation warnings in the build output so I took it upon myself to fix them.  I thought it would be easy, but it was actually fairly painful, mostly because I had to handle the gds-api-adapters upgrade that switched the responses from being openstructs (boo) to hashes (yay).  As a result I've touched loads of files, so it's probably best looking at this commit-by-commit.

### DIFF
--- a/spec/lib/email_alert_signup_api_spec.rb
+++ b/spec/lib/email_alert_signup_api_spec.rb
@@ -11,18 +11,18 @@ describe EmailAlertSignupAPI do
   }
   let(:available_choices) {
     [
-      OpenStruct.new(
+      {
         "key" => "first",
         "radio_button_name" => "First thing",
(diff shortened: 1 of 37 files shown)


spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -11,18 +11,19 @@
   }
   let(:available_choices) {
     [
-      OpenStruct.new(
+      {
+
```
emmabeynon: Looks like there's an errant line here

h-lame: Oops!  I've fixed this in a rebase.  Thanks for the eagle-eyed check!

PR #351, opened 2017-12-15
Strip whitespace from search query

brenetic:
The 'Product Search Term' custom dimension sent to Google Analytics Enhanced Ecommerce may contain spaces at the beginning or end, or double spaces between words. This means that search queries with extra spaces are tracked separately from those with single spaces.

https://trello.com/c/nZ2U7b6P

### DIFF
--- a/app/models/search_parameters.rb
+++ b/app/models/search_parameters.rb
@@ -13,7 +13,7 @@ class SearchParameters
   end
 
   def search_term
-    params[:q]
+    params[:q]&.strip&.gsub(/\s{2,}/, ' ')
   end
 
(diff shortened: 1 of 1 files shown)


app/models/search_parameters.rb
```diff
@@ -13,7 +13,7 @@ def initialize(params)
   end
 
   def search_term
-    params[:q]
+    params[:q]&.strip&.gsub(/\s{2,}/, ' ')
```
dwhenry: Not sure if this might be nicer?
```
params[:q] && params[:q].strip.gsub(/\s{2,}/, ' ')
```

PR #354, opened 2017-12-15
Configure email-alert-api with bearer token

gpeng:
Email-alert-api will soon require a bearer token. This PR adds it in eager anticipation.

[Trello](https://trello.com/c/75sVuXhA/455-require-all-apps-to-authenticate-with-email-alert-api)

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -19,7 +19,8 @@ module Services
 
   def self.email_alert_api
     @email_alert_api ||= GdsApi::EmailAlertApi.new(
-      Plek.find("email-alert-api")
+      Plek.find("email-alert-api"),
+      bearer_token: ENV.fetch("EMAIL_ALERT_API_BEARER_TOKEN", "wubbalubbadubdub")
     )
   end
 end


app/lib/services.rb
```diff
@@ -19,7 +19,8 @@ def self.rummager
 
   def self.email_alert_api
     @email_alert_api ||= GdsApi::EmailAlertApi.new(
-      Plek.find("email-alert-api")
+      Plek.current.find("email-alert-api"),
+      bearer_token: ENV["EMAIL_ALERT_API_BEARER_TOKEN"] || "wubbalubbadubdub"
```
kevindew: ha love the wubbalubbadubdub 😀 

May just be a personal preference but I tend to find `ENV.fetch("EMAIL_ALERT_API_BEARER_TOKEN", "wubbalubbadubdub")` is a bit easier on the eyes, what do you think?

gpeng: Updated in c0f8f41. Your eye ease is always paramount 😆

kevindew:
There's a second approval for you 👍

PR #372, opened 2018-01-08
[DO NOT MERGE] Fix keyboard navigation on search page filters

andysellick:
Attempt to fix a problem on https://www.gov.uk/search?q=test where if navigating with the keyboard the user is forced to tab through all of the organisation checkboxes (100 of them) before reaching the search results.

Tested in Voiceover on Safari, Jaws 17 and 18 on IE.


fofr:
My suggestion to fix this would be to switch to using the option select component if possible.
Otherwise to try and employ the same technique that component uses for tabbing through (which I think is just using display:none and not opening the box when tabbing to it)

PR #374, opened 2018-01-10
Use option select component for search filters

andysellick:
Replace the 'filter by' section of the search results page with the option select component from static. This is to fix several accessibility issues (already fixed in the component), namely:

- keyboard users had to navigate through all 100 checkboxes to get to the search results
- expand/collapse ('Organisations' header) had no button or indication for screen reader users that it could be interacted with.

Before:

screen shot 2018-01-10 at 12 11 45

After:

screen shot 2018-01-10 at 12 13 08


nickcolley:
Great to see this happening 🙌

nickcolley:
@andysellick I feel comfortable reviewing the `govuk_publishing_component` gem changes but not anything else since I'm not going to be working on it, sorry

dwhenry:
It all looks fine, my only question is related to the filter options ordering.

This looks to have changed from alphabetical to document count desc ordering. 

I assume we are happy with this change/have A/B tested it with users to see if it affects usage?

PR #393, opened 2018-02-15
Use feedback component

tijmenb:
This replaces the "Is there anything wrong?" link with the new feedback component. This component asks the user if the page is useful and reports the results to Google Analytics.

See https://github.com/alphagov/govuk_publishing_components/pull/163 for the component.

https://trello.com/c/eKtXGtBD

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -139,7 +139,7 @@ GEM
       sass (>= 3.2.0)
     govuk_navigation_helpers (7.3.0)
       gds-api-adapters (>= 43.0)
-    govuk_publishing_components (4.1.1)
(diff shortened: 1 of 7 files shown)


Gemfile.lock
```diff
@@ -382,12 +382,12 @@ DEPENDENCIES
...
+  slimmer (~> 12.0.0)
   uglifier (~> 4.1)
   webmock (~> 3.3.0)
 
 RUBY VERSION
    ruby 2.4.2p198
 
 BUNDLED WITH
-   1.15.1
+   1.15.4
```
sihugh: We should maybe bump our bundler at some point

PR #404, opened 2018-02-20
Allow easier development of new finders

tijmenb:
To play around with building a new finder you currently have to go into one of the publishing apps (specialist publisher, policy publisher or whitehall), develop your finder and publish it to the publishing-api. This will make it available to this application. The feedback cycle for this is quite long and you'll have to have the development VM running.

This introduces a env var to override the finder. If set the app will use some JSON on your file system instead of talking to the content store.

This will allow you to quickly prototype new finders by creating a new content item in a file and running the app against it. You can copy paste `aaib_reports_example.json` or edit it.


boffbowsh:
😍

PR #406, opened 2018-02-20
[DON'T MERGE] Prototype finder replacement for the whitehall finders

tijmenb:
A prototype of a finder for everything on GOV.UK.

👉https://finder-frontend-pr-406.herokuapp.com/find-all-the-things!11!

Scoped to the education taxonomy:

https://finder-frontend-pr-406.herokuapp.com/find-all-the-things!11!?taxons=c58fdadd-7743-46d6-9629-90bb3ccc4ef0

screen shot 2018-02-20 at 15 21 38

PR #411, opened 2018-02-23
Add taxonomy and content purpose supergroup finder

steventux:
Taxonomy finder based on [@tijmenb's proof of concept](https://github.com/alphagov/finder-frontend/pull/406).
This should give us an idea of how a user journey from topic pages to content pages might look, giving the user the ability to filter topics down by content type. 

screenshot from 2018-03-13 17-04-53

### Example:
https://finder-frontend-pr-411.herokuapp.com/search/advanced?taxons=/education/funding-and-finance-for-students&content_purpose_supergroup=news_and_communications

TODO:

- [x] Present filtering criteria above results eg. _News stories about Education_
- [x] Present content_purpose_subgroups as filtering facets
- [x] Use `content_purpose_supergroup` and `content_purpose_subgroup` once available.


app/lib/advanced_search_finder_api.rb
```diff
@@ -0,0 +1,45 @@
...
+    content_item
+  end
+
+  def augment_facets_with_dynamic_values(content_item, _)
+    augment_facets_with_dynamic_subgroups(content_item) if supergroups.any?
+  end
+
+  def augment_facets_with_dynamic_subgroups(content_item)
+    subgroups = supergroups.map(&:subgroups_as_hash).flatten
+    facet = find_facet(content_item, "content_purpose_subgroup")
```
emmabeynon: Should we guard against this not finding the facet?

features/advanced_search.feature
```diff
@@ -0,0 +1,21 @@
...
+    Given a collection of tagged documents in supergroup 'news_and_communications'
+    When I search by taxon and by supergroup
+    Then I only see documents tagged to the taxon within the supergroup
+
+  Scenario: Filters documents by taxon, supergroup and subgroups
+    Given a collection of tagged documents in supergroup 'news_and_communications' and subgroups 'news,updates_and_alerts'
+    When I search by taxon, supergroup and subgroups
+    Then I only see documents tagged to the taxon within the supergroup and subgroups
+
+  #Scenario: Additionally filtering documents by subgroup
```
vanitabarrett: Should this either be uncommented or removed completely?

Gemfile.lock
```diff
@@ -389,4 +391,4 @@ RUBY VERSION
    ruby 2.4.2p198
 
 BUNDLED WITH
-   1.15.4
```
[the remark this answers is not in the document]
steventux: Yes - thanks for spotting this.

app/presenters/advanced_search_finder_presenter.rb
```diff
@@ -23,6 +23,12 @@ def taxon_link
     link_to taxon['title'], taxon['base_path'], class: 'taxon-link'
   end
 
+  def breadcrumbs
+    @data ||= GovukNavigationHelpers::TaxonBreadcrumbs.new(content_item).breadcrumbs
+    filtered = @data[:breadcrumbs].reject { |bc| bc[:is_page_parent] || bc[:is_current_page] }
```
emmabeynon: govuk_navigation helpers 9.2.1 ensures that current page breadcrumbs are no longer returned (see https://github.com/alphagov/govuk_navigation_helpers/pull/125/files) so once the version in this app is bumped we should no longer need to check for `:is_current_page` 🎉

app/lib/advanced_search_finder_api.rb
steventux: It's unlikely to be missing as it's configured in the content item, but now I've said that it's bound to go missing so I've short-cut this method if not present (next line)

PR #425, opened 2018-03-06
Allow Govspeak in email signup description

rubenarakelyan:
This commit allows Govspeak to be embedded in the description field for email signup pages. This is so we can link to other government organisations or privacy policies where the emails sent as part of the subscription don’t come from GDS (such as drug safety updates).

Trello: https://trello.com/c/QeBmdZCN/670-update-drug-safety-update-email-signup-page

### DIFF
--- a/app/views/email_alert_subscriptions/new.html.erb
+++ b/app/views/email_alert_subscriptions/new.html.erb
@@ -35,7 +35,9 @@
         <% end %>
       <% end %>
 
-      <p><%= @signup.body %></p>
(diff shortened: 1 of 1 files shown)


rubenarakelyan:
These are all published by specialist-publisher as content items when a rake task is run (i.e. they are hardcoded into the repo)

thomasleese:
That sounds fine to me, is it worth thinking if we could render it as if it were Markdown/GovSpeak?

rubenarakelyan:
@thomasleese This is why we have you here! I've changed it to render Govspeak instead

PR #446, opened 2018-03-22
Ensure that the Home breadcrumb is always displayed

steventux:
https://trello.com/c/hfKXtUcQ/11-list-page-breadcrumbs-inconsistent

#### Missing root breadcrumb:

screenshot from 2018-03-22 10-37-44

This breadcrumb can be `is_parent_page: true` (eg. for "Money" taxon) - this should not be excluded
when filtering breadcrumbs.

#### Proposed fix:
https://finder-frontend-pr-446.herokuapp.com/search/advanced?topic=/money&group=news_and_communications

### DIFF
--- a/app/presenters/advanced_search_finder_presenter.rb
+++ b/app/presenters/advanced_search_finder_presenter.rb
@@ -25,11 +25,16 @@ class AdvancedSearchFinderPresenter < FinderPresenter
 
   def breadcrumbs
     @data ||= GovukNavigationHelpers::TaxonBreadcrumbs.new(content_item).breadcrumbs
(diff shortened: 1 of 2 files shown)


fofr:
Should the crown and the text line up?

screen shot 2018-03-22 at 11 33 54

vanitabarrett:
@fofr @steventux Looks like everything (breadcrumbs, title, search boxes etc) has a padding left of 15px which is pushing it out of line with the crown.

steventux:
@vanitabarrett I'd prefer to address the header spacing issues in a separate PR rather than block this one, it's intended to fix a missing breadcrumb.

steventux:
@fofr @vanitabarrett  the header offset problem is noted here https://trello.com/c/6kMaqrEb/14-list-pages-header-misaligned

PR #451, opened 2018-03-26
Use contextual navigation breadcrumbs

tijmenb:
Using the component means that finders can be part of the taxonomy, step by step and normal breadcrumbs.

https://trello.com/c/xQMa0luE


app/controllers/search_controller.rb
```diff
@@ -12,7 +12,6 @@ def index
     search_params = SearchParameters.new(params)
 
     @content_item = content_store.content_item("/search").to_hash
-    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(@content_item)
```
steventux: Do we also need to remove/rework https://github.com/alphagov/finder-frontend/blob/master/app/presenters/advanced_search_finder_presenter.rb#L27 ? Seems there might be a gap in test coverage if this wasn't caught.

PR #452, opened 2018-03-26
Present relevant supergroup metadata

steventux:
https://trello.com/c/5dqDUMJR/3-make-metadata-rules-for-supergroups-consistent-with-topic-pages

screenshot from 2018-03-26 17-07-15

Example: https://finder-frontend-pr-452.herokuapp.com/search/advanced?group=guidance_and_regulation&page=3&topic=%2Feducation%2Fprimary-curriculum-key-stage-1-tests-and-assessments

DNM: Depends on https://github.com/alphagov/rummager/pull/1209

### DIFF
--- a/app/lib/advanced_search_finder_api.rb
+++ b/app/lib/advanced_search_finder_api.rb
@@ -60,5 +60,9 @@ private
     raise TaxonNotFound.new(msg)
   end
 
+  def query_builder_class
+    AdvancedSearchQueryBuilder
+  end
+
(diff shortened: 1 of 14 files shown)


vanitabarrett:
Hardcoding component styling isn't ideal as it means we have styling getting out of sync with the actual component, but I'm not sure of an alternative when working with mustache templates. It may be worth making a note of this problem somewhere so it can be investigated in the future.

PR #462, opened 2018-04-05
Use search component from gem, not static

andysellick:
Search component has moved into the components gem.

Related to this PR to [add the search component to the gem](https://github.com/alphagov/govuk_publishing_components/pull/267).

### DIFF
--- a/app/views/search/_search_field.html.erb
+++ b/app/views/search/_search_field.html.erb
@@ -1,29 +1,24 @@
-<div class="searchfield">
-  <fieldset class="search-input">
-    <label for="search-main">
-      <%- if @search_term.present? -%>
-        <h1>Search results for <span class="visuallyhidden">&ldquo;<%= @search_term %>&rdquo;</span ></h1>
-      <%- else -%>
-        <h1>Search GOV.UK</h1>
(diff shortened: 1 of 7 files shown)


app/views/search/_search_field.html.erb
```diff
@@ -1,3 +1,29 @@
...
+<%= hidden_field_tag("show_organisations_filter", params[:show_organisations_filter]) if params[:show_organisations_filter] %>
+<%= hidden_field_tag("filter_manual[]", params[:filter_manual]) if params[:filter_manual] %>
+<%= hidden_field_tag(:debug_score, params[:debug_score]) if params[:debug_score] %>
+<%= hidden_field_tag(:debug, params[:debug]) if params[:debug] %>
+
+<% if @spelling_suggestion %>
+  <p>Did you mean <%= link_to "#{@spelling_suggestion.query}", @spelling_suggestion.link %></p>
+<% end %>
+
+<% if false %>
```
tijmenb: This doesn't seem right?

andysellick: Eeep, left in some debug. Sorry, will correct, thanks.

PR #467, opened 2018-04-10
[WIP]

vanitabarrett:


Gemfile.lock
```diff
@@ -388,7 +388,7 @@ DEPENDENCIES
   webmock (~> 3.3.0)
 
 RUBY VERSION
-   ruby 2.4.2p198
+   ruby 2.4.4p296
```
cbaines: Would be good to remove this change, or put it in a separate commit if you want to keep it.

leenagupte: It should have been included as part of #466

vanitabarrett: Updated - pulled out into separate commit

PR #508, opened 2018-06-08
Use title component from the gem

tijmenb:
No test changes seem necessary.

https://trello.com/c/EruCrkqO

### DIFF
--- a/app/views/advanced_search_finder/show.html.erb
+++ b/app/views/advanced_search_finder/show.html.erb
@@ -26,7 +26,7 @@
 
     <div class="title-and-metadata">
       <%= finder.taxon_link %>
-      <%= render partial: 'govuk_component/title', locals: { title: finder.title } %>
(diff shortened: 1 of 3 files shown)


app/views/advanced_search_finder/show.html.erb
```diff
@@ -26,7 +26,7 @@
 
     <div class="title-and-metadata">
       <%= finder.taxon_link %>
-      <%= render partial: 'govuk_component/title', locals: { title: finder.title } %>
+      <%= render partial: 'govuk_publishing_components/components/title', locals: { title: finder.title } %>
```
vanitabarrett: I don't think we need the `locals:` bit.

tijmenb: I think `locals:` is needed when you use the `partial:` syntax. Perhaps in a separate PR we can clean all of those up?

PR #619, opened 2018-09-28
[Redux] Translate content purpose fields

andrewgarner:
This brings back #606 with a fix for a bug identified from manual testing.

When we translate the content purpose fields the `subgroup` fields need to take precedence over the `supergroup`. For example, if we have a URL such as `/search/advanced?subgroup[]=policy&topic=/transport&group=policy_and_engagement` we now need to only lookup the document types for the subgroup "policy". The regression meant we were also lookup for the document types for the group "policy_and_engagement" meaning our query to the Search API was too broad.

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -33,6 +33,8 @@ private
       params: filter_params,
     ).call
 
+    query = TranslateContentPurposeFields.new(query).call
+
     Services.rummager.search(query).to_hash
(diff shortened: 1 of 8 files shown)


sihugh:
@andrewgarner can you resolve the merge conflict?  I'll approve it after that and get it deployed

andrewgarner:
@sihugh good to go now

PR #629, opened 2018-10-10
Revert "Bump govuk_publishing_components from 11.2.0 to 12.0.0"

cbaines:
This seems to be causing the title and breadcrumbs to be black, rather
than white.

This reverts commit 1543d0fe5769b68d6354b9e39c77c6a43ae32f62.


cbaines:
## Bad
bad

## Good
good

PR #631, opened 2018-10-15
Add Google Analytics tracking for finder filters

vanitabarrett:
Trello: https://trello.com/c/TNs02i4D/10-adding-event-tracking-to-the-finder-frontend-finders

Adds Google Analytics tracking to finder filters, i.e: search box, dropdowns, and date filter.

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -17,14 +17,22 @@
 
     if(GOVUK.support.history()){
       this.saveState();
-      this.$form.on('change', 'input[type=checkbox], input[type=text], input[type=radio]', this.formChange.bind(this));
 
-      this.$form.find('input[type=text]').keypress(
-        function(e){
-          if(e.keyCode == 13) {
-            // 13 is the return key
+      this.$form.find('input[type=checkbox], input[type=text], input[type=radio]').on('change',
+        function(e) {
+          if (e.target.type == "text") {
             LiveSearch.prototype.fireTextAnalyticsEvent(e);
-            this.formChange();
+          }
+          this.formChange(e)
+        }.bind(this)
+      );
+
+      this.$form.find('input[type=text]').on('keypress',
+        function(e){
+          var ENTER_KEY = 13
+
+          if(e.keyCode == ENTER_KEY) {
+            this.formChange(e);
             e.preventDefault();
           }
         }.bind(this)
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -19,10 +19,11 @@
       this.saveState();
       this.$form.on('change', 'input[type=checkbox], input[type=text], input[type=radio]', this.formChange.bind(this));
 
-      this.$form.find('input[type=text]').keypress(
+      this.$form.find('input[type=text]').on('keypress blur',
         function(e){
-          if(e.keyCode == 13) {
+          if(e.keyCode == 13 || e.type == 'blur') {
```
bilbof: I find it easier to read things like keycodes if they've been put into a variable (e.g. `const ENTER_KEY = 13`). Not a required change though as it's reasonably clear already :)

app/assets/javascripts/live_search.js
```diff
@@ -65,10 +66,26 @@
...
     }
   };
 
+  LiveSearch.prototype.fireTextAnalyticsEvent = function(event) {
+    if (GOVUK.analytics && GOVUK.analytics.trackPageview) {
+      var options = {transport: 'beacon'};
+      var category = "filterClicked";
+      var action = $('label[for="' + event.target.id + '"]')[0].innerText;
+
+      options.label = $(event.target)[0].value;
```
bilbof: You could do

```
var options = {
    transport: 'beacon',
    label: $(event.target)[0].value
};
```

app/assets/javascripts/live_search.js
```diff
@@ -65,10 +66,26 @@
             GOVUK.analytics.trackPageview(newPath);
           }
         }.bind(this)
-      );
+      )
     }
   };
 
+  LiveSearch.prototype.fireTextAnalyticsEvent = function(event) {
+    if (GOVUK.analytics && GOVUK.analytics.trackPageview) {
```
bilbof: This check (`GOVUK.analytics && GOVUK.analytics.trackPageview`) is done elsewhere in the file. Do you think it would be worth adding a new function for `LiveSearch.canTrackPageview`?

app/models/select_facet.rb
```diff
@@ -12,6 +12,11 @@ def options
         value: allowed_value['value'],
         label: allowed_value['label'],
         id: allowed_value['value'],
+        data_attributes: {
+          track_category: "filterClicked",
```
bilbof: This string is defined in a few different places. Would it be possible to define it as a constant in one place and import it?

vanitabarrett:
@bilbof Thanks for your comments - I think I've addressed most of them! I've also tweaked the event tracking a bit (see the latest commit). Please could you re-review?

bilbof:
🚀

PR #644, opened 2018-11-02
Serve a 422 (unprocessable) when rummager returns 422

sihugh:
At present when requests fail Rummager's validation ([recent example][1]), Rummager will return a 422 (unprocessable) response.  Finder-frontend receives this as a GdsApi::HTTPUnprocessableEntity exception, which results in a 500 error.

If we handle the 422 and forward it on then it makes it clear to the user (usually a script exploring
the site for vulnerabilities) that we're handling the request appropriately and that they need to do something else.

This also reduces the 500 count for alerting purposes, which at times is very noisy.  It's hard to tell the difference between legitimate server issues and vexatious use.

[1]: https://github.com/alphagov/rummager/pull/1307

### DIFF
--- a/features/support/document_helper.rb
+++ b/features/support/document_helper.rb
@@ -43,6 +43,10 @@ module DocumentHelper
     )
   end
 
+  def stub_rummager_api_request_with_422_response(page_number)
+    stub_request(:get, rummager_policy_other_page_search_url(page_number)).to_return(status: 422)
+  end
+
   def stub_rummager_api_request_with_policies_finder_results
     stub_request(:get, rummager_policies_finder_search_url).to_return(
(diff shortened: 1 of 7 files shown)


features/support/document_helper.rb
```diff
@@ -157,6 +161,16 @@ def rummager_policy_page_2_search_url
     )
   end
 
+  def rummager_policy_other_page_search_url(page_number)
+    rummager_url(
+      policy_search_params.merge(
+        "order" => "-public_timestamp",
+        "count" => 10,
+        "start" => ((page_number -1) * 10)
```
bilbof: Is the multiplication by ten increasing the page number for good measure?

PR #645, opened 2018-11-05
Add ability to sort finders

andrewgarner:
Adds an option to the news and communications finder to provide sorting options for the results shown.

screen shot 2018-11-23 at 09 11 26

-------

https://trello.com/c/CF6DmPXk

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -18,10 +18,14 @@
 
     this.emailSignupHref = this.$emailLink.attr('href');
 
+    this.$orderSelect = this.$form.find('.js-order-results');
+    this.$relevanceOrderOption = this.$orderSelect.find('option[value=' + this.$orderSelect.data('relevance-sort-option') + ']');
+    this.$relevanceOrderOptionIndex = this.$relevanceOrderOption.index();
+
     if(GOVUK.support.history()){
       this.saveState();
 
-      this.$form.find('input[type=checkbox], input[type=text], input[type=radio]').on('change',
+      this.$form.find('input[type=checkbox], input[type=text], input[type=radio], select').on('change',
         function(e) {
           if (e.target.type == "text") {
             LiveSearch.prototype.fireTextAnalyticsEvent(e);
@@ -40,6 +44,8 @@
           }
         }.bind(this)
       );
+
+      this.updateOrder();
 
       $(window).on('popstate', this.popState.bind(this));
     } else {
@@ -62,6 +68,7 @@
   LiveSearch.prototype.popState = function popState(event){
     if(event.originalEvent.state){
       this.saveState(event.originalEvent.state);
+      this.updateOrder();
       this.updateResults();
       this.restoreBooleans();
       this.restoreTextInputs();
@@ -72,6 +79,7 @@
     var pageUpdated;
     if(this.isNewState()){
       this.saveState();
+      this.updateOrder();
       pageUpdated = this.updateResults();
       pageUpdated.done(
         function(){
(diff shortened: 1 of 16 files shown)


app/views/finders/_facet_collection.html.erb
```diff
@@ -1,12 +1,10 @@
...
-    <form method="get" action="<%=finder.slug%>" class="js-live-search-form">
-      <div class="filter text-filter">
-        <label class="legend" for="finder-keyword-search">Search</label>
-        <input value="<%= @results.user_supplied_keywords %>" type="text" name="keywords" id="finder-keyword-search" aria-controls="js-search-results-info" class='text'/>
-      </div>
-      <%= render facet_collection.filters %>
-      <input type="submit" value="Filter results" class="button js-live-search-fallback"/>
-    </form>
+    <div class="filter text-filter">
+      <label class="legend" for="finder-keyword-search">Search</label>
```
andysellick: I had a go at using components for some of these labels and inputs so the styles match with the new select but some work needs to be done on the components first so they support all we need. Have made a note on the card.

app/lib/search_query_builder.rb
```diff
@@ -65,7 +65,23 @@ def metadata_fields
...
   def order_query
-    keywords ? order_by_relevance_query : default_order_query
+    sort_options = finder_content_item.dig('details', 'sort')
+
+    if sort_options.present?
+      sort_option = if params['order']
+                      sort_options.detect { |option| option['name'].parameterize == params['order'] }
+                    else
+                      sort_options.detect { |option| option['default'] } || { 'key' => default_order }
+                    end
```
sihugh: We should probably also catch when someone's used an invalid option for `order` so that it reverts to the default rather than erroring: https://finder-frontend-pr-645.herokuapp.com/news-and-communications?order=crumpets

andrewgarner: Fixed with 12d03ac584eee7dc47c7c289c0acd1de12a373ba

app/assets/javascripts/live_search.js
```diff
@@ -172,17 +208,17 @@
 
   LiveSearch.prototype.restoreTextInputs = function restoreTextInputs(){
     var that = this;
-    this.$form.find('input[type=text]').each(function(i, el){
+    this.$form.find('input[type=text], select').each(function(i, el){
```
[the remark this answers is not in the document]
andysellick: I know what it is :) my point was that it isn't used in the each, so there's no need to declare it.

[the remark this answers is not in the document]
bilbof: From the looks of it, you could use `this` to refer to the element, and then you don't need to declare `i` or `el`. Seems fine either way, and I appreciate you haven't added it here! :-)

app/views/finders/_facet_collection.html.erb
andrewgarner: See https://trello.com/c/fhhTwfLo

features/fixtures/news_and_communications.json
```diff
@@ -31,6 +31,25 @@
...
+        "key": "-relevance"
+      },
+      {
+        "name": "Updated (newest)",
+        "key": "-public_timestamp",
+        "default": true
+      },
+      {
+        "name": "Updated (oldest)",
+        "key": "public_timestamp"
```
bilbof: It might be nice to have a list of sort options one can default to. If the above are going to be quite common it would save us republishing finder content items.

app/assets/javascripts/live_search.js
```diff
@@ -118,6 +126,34 @@
...
+          this.state = this.$form.serializeArray();
+        }
+      } else {
+        relevanceOrderOption.removeAttr('disabled');
+        relevanceOrderOption.remove();
+
+        if (previousKeywords !== "") {
+          orderSelect.val(this.$orderSelect.data('default-sort-option'));
+          this.state = this.$form.serializeArray();
+        }
```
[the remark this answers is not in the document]
bilbof: You could also turn the big if statement into a return clause (` if (!orderSelect.length) { return }`)

app/views/finders/show.html.erb
```diff
@@ -59,19 +59,35 @@
...
+
+    <% if finder.sort.present? %>
+      <div class="filter-form">
+        <div class="govuk-form-group">
+          <%= label_tag 'order', 'Sort by', class: 'govuk-label' %>
+          <%= select_tag 'order', finder.sort_options,
+                         class: 'govuk-select js-order-results',
+                         'aria-controls': 'js-search-results-info',
+                         data: { 'default-sort-option' => finder.default_sort_option_value,
+                                 'relevance-sort-option' => finder.relevance_sort_option_value } %>
```
bilbof: Perhaps you could provide a `sort_options` method containing this hash, then you could expose a simpler sorting api in finder presenter, and perhaps introduce a separate class that handles sorting logic? This would simplify the finder presenter and views and would make it easier to skim read the sorting logic.

app/assets/javascripts/live_search.js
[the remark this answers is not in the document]
andrewgarner: Great feedback. I've added ab0c15e to address.

app/assets/javascripts/live_search.js
```diff
@@ -118,6 +126,34 @@
...
+    var orderSelect = this.$orderSelect;
+
+    if (orderSelect.length) {
+      var keywords = this.getTextInputValue('keywords', this.state);
+      var previousKeywords = this.getTextInputValue('keywords', this.previousState);
+      var relevanceOrderOption = this.$relevanceOrderOption;
+      var relevanceOrderOptionIndex = this.$relevanceOrderOptionIndex;
+
+      if (keywords !== "") {
+        orderSelect.children("option").eq(relevanceOrderOptionIndex).before(relevanceOrderOption);
```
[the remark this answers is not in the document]
andrewgarner: I've extracted this into it's own method in ab0c15e to hopefully make this easier to understand.

app/lib/search_query_builder.rb
```diff
@@ -65,7 +65,23 @@ def metadata_fields
...
+      sort_option ||= sort_options.detect { |option| option['default'] } || { 'key' => default_order }
+
+      if %w(relevance -relevance).include?(sort_option['key'])
+        order_by_relevance_query
+      else
+        { 'order' => sort_option['key'] }
+      end
+    else
+      keywords ? order_by_relevance_query : default_order_query
+    end
```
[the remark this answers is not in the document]
andrewgarner: Fixed with 5f35b94

app/presenters/result_set_presenter.rb
```diff
@@ -98,6 +99,18 @@ def user_supplied_keywords
     @filter_params.fetch('keywords', '')
   end
 
+  def sort_options
+    return unless finder.respond_to?(:sort) && finder.sort.present?
```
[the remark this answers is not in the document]
andrewgarner: Fixed with 7904dd6a3fc9376376f01c828cf294d8e14963cd

app/lib/search_query_builder.rb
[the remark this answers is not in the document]
andrewgarner: This is confusing because we're supporting ordering by either the `sort` or `default_order` (deprecated) attributes of the finder.

If the sort option is present we want to sort by it, unless the key is `relevance` or `-relevance`.

app/assets/javascripts/live_search.js
```diff
@@ -118,6 +126,65 @@
...
+
+  LiveSearch.prototype.insertRelevanceOption = function insertRelevanceOption() {
+    var adjacentOption = this.$orderSelect.children("option").eq(this.$relevanceOrderOptionIndex);
+
+    adjacentOption.before(this.$relevanceOrderOption);
+  };
+
+  LiveSearch.prototype.removeRelevanceOption = function removeRelevanceOption() {
+    this.$relevanceOrderOption.removeAttr('disabled');
+    this.$relevanceOrderOption.remove();
```
bilbof: Is it necessary to remove an attribute from the element before removing it?

andrewgarner: The server side render will disable the relevance option so we either need to remove it here, or when we add it back, or maybe when we initialise the live search.

bilbof: Ah I see, thanks

PR #653, opened 2018-11-12
Add world locations facet to news and communications finder

bilbof:
This is a schema change to the news and communications finder, enabling end users (in development mode) to filter from a selection of world locations.

**NOTE:** Should be merged _after_ https://github.com/alphagov/rummager/pull/1318

screen shot 2018-11-12 at 14 29 32

### DIFF
--- a/features/fixtures/news_and_communications.json
+++ b/features/fixtures/news_and_communications.json
@@ -65,6 +65,14 @@
         "filterable": true
       },
       {
+        "key": "world_locations",
+        "name": "World location",
+        "preposition": "in",
+        "type": "text",
+        "display_as_result_metadata": true,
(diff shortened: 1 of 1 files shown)


features/fixtures/news_and_communications.json
```diff
@@ -64,6 +64,14 @@
         "display_as_result_metadata": true,
         "filterable": true
       },
+      {
+        "key": "world_locations",
+        "name": "World locations",
```
andrewgarner: In https://trello.com/c/JfXHOotH it is "World location" (singular).

bilbof: Thanks - changed that!

PR #659, opened 2018-11-16
Enable faceting by World Location

bilbof:
With this change, World Locations become available as a facet in Finders. For the moment this will only be visible on the news and communications finder (currently not available outside of development mode).

This change is necessary because rummager does not provide the required data to provide world location options in a facet. We need to get additional world location data from Whitehall, until world locations are migrated.  This data is cached in a Registry.

A longer term solution would be to bring world locations back into rummager (they were removed earlier this year). This is being considered but it may be some time before this is worked on.

I added a World Locations Registry, similar to the registries we have in Rummager. Initially I had planned to add this world location registry to rummager, but having the registry downstream of rummager (it would only be used by finder-frontend) ensures there's no dependency on Whitehall in rummager.

This unmerged PR demonstrates how this could have been implemented in Rummager:
https://github.com/alphagov/rummager/pull/1318/commits/d54d0a6e3f22148ea645cf8c34333c1174fd9fc2.

By default, the world locations finder will not be populated in development mode. It relies on Whitehall being up, so if whitehall-admin happens to be up, or if you start finder frontend with the --live flag, the world locations registry will be populated with data.

### Screenshots 

screen shot 2018-11-16 at 17 50 34
screen shot 2018-11-16 at 17 50 26

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -70,21 +70,33 @@ private
   def augment_facets_with_dynamic_values(content_item, search_response)
     search_response.fetch("facets", {}).each do |facet_key, facet_details|
       facet = content_item['details']['facets'].find { |f| f['key'] == facet_key }
-      facet['allowed_values'] = allowed_values_for_facet_details(facet_details) if facet
+      facet['allowed_values'] = allowed_values_for_facet_details(facet_key, facet_details) if facet
     end
   end
 
-  def allowed_values_for_facet_details(facet_details)
-    values = facet_details.fetch("options", {})
-      .reject { |f| f.dig("value", "title").nil? }
+  def allowed_values_for_facet_details(facet_key, facet_details)
+    facet_details.fetch("options", {})
       .map { |f| f.fetch("value", {}) }
+      .map { |value| present_facet_option(value, facet_key) }
+      .reject { |f| f["label"].blank? || f["value"].blank? }
+  end
 
-    values.map { |value|
-      {
-        'label' => value.fetch("title", ""),
-        'value' => value.fetch("slug", ""),
-      }
+  def present_facet_option(value, facet_key)
+    slug = value.fetch("slug", "")
+    label = value.fetch("title", find_facet_title_by_slug(slug, facet_key))
+
+    {
+      "label" => label,
+      "value" => slug,
     }
+  end
+
+  def find_facet_title_by_slug(slug, facet_key)
+    registry = registries.all[facet_key]
+    return "" if registry.nil?
+
+    item = registry[slug] || {}
+    item.fetch("title", "")
   end
 
   def build_pagination(documents_per_page, start_offset, total_results)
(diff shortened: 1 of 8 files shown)


app/lib/registries/world_locations_registry.rb
```diff
@@ -0,0 +1,37 @@
...
+          'title' => result.dig('title'),
+          'slug' => result.dig('details', 'slug')
+        }
+      }
+    end
+
+    def fetch_locations_from_worldwide_api
+      Services.worldwide_api.world_locations.with_subsequent_pages.to_a
+    rescue GdsApi::HTTPServerError, GdsApi::HTTPBadGateway
+      puts "ERROR: Could not get WorldLocation from Whitehall"
```
sihugh: Might be worth tweaking this to be "the API" or something rather than Whitehall?

I'm wondering what we want the behaviour to be here.  If it errors then we cache an empty array for an hour which is probably not quite what we want.  Might it be worth moving the `rescue` to outside the cache?

sihugh: Adding a test for this unhappy path would be amazing 👍

bilbof: Both very good points, thank you!

app/lib/finder_api.rb
```diff
@@ -70,23 +70,35 @@ def augment_content_item_details_with_results(content_item, search_response)
...
   end
 
-  def allowed_values_for_facet_details(facet_details)
-    values = facet_details.fetch("options", {})
-      .reject { |f| f.dig("value", "title").nil? }
+  def allowed_values_for_facet_details(facet_key, facet_details)
+    facet_details.fetch("options", {})
       .map { |f| f.fetch("value", {}) }
+      .map { |value| present_facet_option(value, facet_key) }
+      .reject { |f| !f.dig("label").present? || !f.dig("value").present? }
```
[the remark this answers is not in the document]
bilbof: Nope, changed this - thank you 😄

app/lib/registries/world_locations_registry.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+  private
+
+    def cached_locations
+      Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
+        fetch_locations
+      end
+    rescue GdsApi::HTTPServerError, GdsApi::HTTPBadGateway
+      GovukStatsd.increment "ERROR: Could not get world locations from the API!"
```
sihugh: I think the text should be like a key and you can namespace it if you want.  Something like `registries.world_location_api_errors` would work? Here's an example of an existing one: https://github.com/alphagov/signon/blob/d8b0202e1c601045560845c42d86c296c4adb5bd/app/controllers/passwords_controller.rb#L41 (sorry 🙈 ).

You can navigate the ones we have at the moment under `stats.govuk.app....` on https://graphite.publishing.service.gov.uk/dashboard/

bilbof: Aha, thank you very much 😃

sihugh:
🗺 🇬🇧 👍 🚀

PR #664, opened 2018-11-19
Add multiple facets to finder email signup

AlanGabbianelli:
This allows users to sign up to email alerts from finders and filtering from multiple facets rather than currently where they can only choose one.

### DIFF
--- a/lib/email_alert_title_builder.rb
+++ b/lib/email_alert_title_builder.rb
@@ -1,0 +1,61 @@
+class EmailAlertTitleBuilder
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def initialize(filter:, subscription_list_title_prefix:, facets:)
+    @filter = filter
+    @subscription_list_title_prefix = subscription_list_title_prefix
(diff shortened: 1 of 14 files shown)


lib/email_alert_title_builder.rb
```diff
@@ -53,16 +42,20 @@ def plural_or_single
...
   def prefix
-    if facets.empty?
-      subscription_list_title_prefix.to_s
-    elsif facets.size == 1
+    if facets.size == 1
       subscription_list_title_prefix[plural_or_single].to_s
     elsif selected_facets.empty?
       subscription_list_title_prefix.to_s
     elsif subscription_list_title_prefix
       "#{subscription_list_title_prefix}with "
```
bevanloon: is there a missing space before `with`?
`"#{subscription_list_title_prefix}with "` vs `"#{subscription_list_title_prefix} with "`

thomasleese:
@bevanloon Thanks, I've done that and pushed!

PR #667, opened 2018-11-20
Remember choices for email subscriptions

AlanGabbianelli:
Persist choices selected in the finder at the moment of signing up for email subscriptions, and allow choices to be automatically checked based on the query string.

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -61,6 +61,8 @@
     }
   };
 
+  var currentEmailSignupPath = $("p.email-link a").attr('href')
+
   LiveSearch.prototype.formChange = function formChange(e){
     var pageUpdated;
     if(this.isNewState()){
@@ -69,6 +71,8 @@
       pageUpdated.done(
         function(){
           var newPath = window.location.pathname + "?" + $.param(this.state);
+          var newEmailSignupPath = currentEmailSignupPath + "?" + $.param(this.state);
+          $("p.email-link a").attr('href', newEmailSignupPath)
           history.pushState(this.state, '', newPath);
           if (this.canTrackPageview()) {
             GOVUK.analytics.trackPageview(newPath);
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -23,7 +23,7 @@ private
   end
 
   def signup_presenter
-    @signup_presenter ||= SignupPresenter.new(content)
+    @signup_presenter ||= SignupPresenter.new(content, params)
   end
 
   def valid_choices?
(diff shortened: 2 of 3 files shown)


lib/email_alert_title_builder.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def prefix
+    if facets.empty?
+      subscription_list_title_prefix.to_s
+    elsif facets.size == 1
+      subscription_list_title_prefix[plural_or_single].to_s
+    elsif selected_facets.empty?
+      subscription_list_title_prefix.to_s
+    elsif subscription_list_title_prefix
+      "#{subscription_list_title_prefix}with "
```
bevanloon: I think there's a missing space here 
/s/`"#{subscription_list_title_prefix}with "`/`"#{subscription_list_title_prefix} with "`

lib/email_alert_title_builder.rb
```diff
@@ -0,0 +1,68 @@
...
+    @facets = facets
+  end
+
+  def call
+    if facets.empty?
+      prefix
+    elsif facets.size == 1
+      (prefix + topic_names_sentence(facets.first)).upcase_first
+    elsif selected_facets.empty?
+      prefix
```
bevanloon: feels like we are calling `prefix` a lot. I'm not sure it's a problem, but I wonder if the code might read a little more elegantly if we pulled out the conditional logic into a method.

Every case here uses `prefix` so something like:

```
def call
  prefix + suffix
end

def suffix
  if facets.size == 1
      topic_names_sentence(facets.first)).upcase_first
  elsif facets.size > 1
    facet_sentences = selected_facets.map do |facet|
      facet["facet_name"] + " of " + topic_names_sentence(facet)
    end
    facet_sentences.to_sentence).upcase_first
  end
end
```

I'm not sure if I've captured the logic there though? i.e. is that catch-all `elsif` always captured by the case of `facets.size > 1`

lib/email_alert_title_builder.rb
bevanloon: In a sort of direct contrast to my comment above about creating a prefix and suffix method - there's lots of duplicate logic between the `call` and `prefix` - we are running mostly the same conditions so is there any value in splitting these methods up?

Also it looks like `facets.empty?` and `selected_facets.empty?` conditions return the same title regardless, so perhaps they could be merged into an `or`?

AlanGabbianelli: It doesn't look ideal, but `subscription_list_title_prefix` includes a trailing space.

thomasleese: Yeah, the logic used to be all in one method but we found it quite hard to read... I'm not sure if this is any better but we wanted to have a go at refactoring.

We could use an order for `facets.empty?` and `selected_facets.empty?` but we would have to include `facets.size < 1` for the `selected_facets` side because annoyingly the logic isn't quite as simple as an or. We can give it a go and see how it looks.

PR #675, opened 2018-11-22
Add Q&A feature

rubenarakelyan:
This PR adds a new Q&A feature for finders. This feature walks users through a finder's facets step by step and then sends them to the finder with their selections intact.

Paired with @edwardkerry


lib/prepare_business_uk_leaving_eu.yaml
```diff
@@ -0,0 +1,129 @@
+---
```
barrucadu: Should this file be in this PR?  If we're not going live yet, it seems we want this feature but without this finder.

rubenarakelyan: Yup, the Q&A won't be live yet though since that will involve publishing the route from rummager.

app/assets/stylesheets/views/_overrides.scss
```diff
@@ -0,0 +1,8 @@
+/* Overrides styling of legends in finder-frontend.scss */
+legend.gem-c-title--margin-bottom-5 {
+  margin-bottom: $gutter-half !important;
+
+  @include media(tablet) {
+    margin-bottom: $gutter * 1.5 !important;
```
DilwoarH: Should `!important` be used? Was there no other solution?

app/controllers/qa_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+  def page
+    params.permit(:page)
+    params[:page] = 1 if params[:page].nil?
+    params[:page].to_i.clamp(1, facets.length)
+  end
+
+  def fetch_breadcrumbs
+    parent_content_item = Services.content_store.content_item(qa_config["finder_base_path"])
+    @breadcrumbs = FinderBreadcrumbsPresenter.new(parent_content_item, @content_item)
+  end
```
AlanGabbianelli: Why not
```
def fetch_breadcrumbs
  @breadcrumbs = FinderBreadcrumbsPresenter.new(raw_finder, @content_item)
end
```
?

AlanGabbianelli: Also, isn't `@content_item` the same as `raw_finder`? Aren't you passing the same argument twice?

AlanGabbianelli: If two arguments are expected, the method could be
```
def fetch_breadcrumbs
  @breadcrumbs = FinderBreadcrumbsPresenter.new(raw_finder, raw_finder)
end
```
This way, `@content_item = raw_finder` could also be removed from the `show` action (assuming it was only needed here).

app/controllers/qa_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def multiple_question?
+    question_type == "multiple"
+  end
+  helper_method :multiple_question?
+
+  def facet_grouped_allowed_values
+    @facet_grouped_allowed_values ||= current_facet["facet"]["allowed_values"].group_by { |filter| filter["value"] }
+  end
+  helper_method :facet_grouped_allowed_values
```
AlanGabbianelli: Why is `facet_grouped_allowed_values ` a `helper_method ` if it's not used in the views?

app/assets/stylesheets/views/_overrides.scss
andysellick: We shouldn't be using !important... 👎

app/controllers/qa_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+        value: filter
+      }
+    end
+  end
+  helper_method :nested_options
+
+  def first_page?
+    page == 1
+  end
+  helper_method :first_page?
```
AlanGabbianelli: Is `first_page?` not used anywhere, or am I missing something?

app/assets/stylesheets/views/_overrides.scss
rubenarakelyan: One for @tombye

app/controllers/qa_controller.rb
rubenarakelyan: Fixed

app/controllers/qa_controller.rb
rubenarakelyan: Removed

app/controllers/qa_controller.rb
rubenarakelyan: Removed

PR #678, opened 2018-11-22
Track QA choices as events

steventux:
https://trello.com/c/b4DOb2HJ/83-add-analytics-to-qa-pages

Possible solution to tracking QA pages with events:

On form submission for each step of the QA journey, the checked items generate one GA event with the question key contained in the label and the checked value as the event value. The event category is `QA option chosen`.

See screenshot for example of generated events...

screenshot from 2018-11-23 11-30-02

### DIFF
--- a/app/assets/javascripts/modules/track-qa-choices.js
+++ b/app/assets/javascripts/modules/track-qa-choices.js
@@ -1,0 +1,30 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
+
+  GOVUK.Modules.TrackQaChoices = function () {
+    this.start = function (element) {
+      track(element)
+    }
+
+    function track (element) {
+      element.on('submit', function (event) {
+        var $checkedOption, eventLabel, options
+        var $submittedForm = $(event.target)
+        var $checkedOptions = $submittedForm.find('input:checked')
+
+        $checkedOptions.each(function(index) {
+          $checkedOption = $(this)
+          eventLabel = $checkedOption.attr('name').replace('[]', '')
+          options = { transport: 'beacon', label: eventLabel }
+
+          GOVUK.analytics.trackEvent('QA option chosen', $checkedOption.val(), options)
+        })
+      })
+    }
+  }
+})(window, window.GOVUK);
--- a/spec/javascripts/modules/track-qa-choices.spec.js
+++ b/spec/javascripts/modules/track-qa-choices.spec.js
@@ -1,0 +1,54 @@
+/* global describe beforeEach it spyOn expect */
+
+var $ = window.jQuery
+
+describe('QA choices tracker', function () {
+
+  var GOVUK = window.GOVUK || {}
+  var tracker
+  var $element
+
+  GOVUK.analytics = GOVUK.analytics || {}
+
+  beforeEach(function () {
+    GOVUK.analytics.trackEvent = function () {}
+    spyOn(GOVUK.analytics, 'trackEvent')
+
+    $element = $(
+      '<div>' +
+        '<form onsubmit="event.preventDefault()">' +
+          '<input name="sector_business_area[]" type="checkbox" value="construction">' +
+          '<input name="sector_business_area[]" type="checkbox" value="accommodation">' +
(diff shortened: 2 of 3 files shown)


startup.sh
```diff
@@ -3,10 +3,10 @@
 bundle install
 
 if [[ $1 == "--live" ]] ; then
-  GOVUK_APP_DOMAIN=www.gov.uk \
-  GOVUK_WEBSITE_ROOT=https://www.gov.uk \
-  PLEK_SERVICE_SEARCH_URI=${PLEK_SERVICE_SEARCH_URI-https://www.gov.uk/api} \
-  PLEK_SERVICE_CONTENT_STORE_URI=${PLEK_SERVICE_CONTENT_STORE_URI-https://www.gov.uk/api} \
+  GOVUK_APP_DOMAIN=www-origin.integration.publishing.service.gov.uk \
```
thomasleese: Did you mean to commit this?

app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,30 @@
+window.GOVUK = window.GOVUK || {}
```
barrucadu: Does this file need including or referencing in the template somewhere, or does that happen magically?

app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,30 @@
...
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
+
+  GOVUK.Modules.TrackQaChoices = function () {
+    this.start = function (element) {
```
tombye: Can `element` be `$element` here to show it's jQuery-wrapped?

startup.sh
steventux: :face_with_head_bandage: Nope, rebased over this.

app/assets/javascripts/modules/track-qa-choices.js
steventux: I believe https://github.com/alphagov/finder-frontend/blob/master/app/assets/javascripts/application.js#L19 includes it in the asset pipeline so it's compressed into `application.js`

spec/javascripts/modules/track-qa-choices.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    )
+    expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+      'QA option chosen', 'construction', { transport: 'beacon', label: 'sector_business_area' }
+    )
+  })
+
+  it('does not track events when no choice is made', function () {
+    element.find('form').trigger('submit')
+
+    expect(GOVUK.analytics.trackEvent).not.toHaveBeenCalledWith(jasmine.anything())
```
tombye: Seems like this should be checking `GOVUK.analytics.trackEvent` wasn't called at all instead of that it wasn't called with anything. That's what I get from reading the test name.

app/assets/javascripts/modules/track-qa-choices.js
steventux: :+1: I prefer that convention

spec/javascripts/modules/track-qa-choices.spec.js
steventux: :+1: That seems cleaner, willdo

app/assets/javascripts/modules/track-qa-choices.js
steventux: https://github.com/alphagov/finder-frontend/pull/678/commits/bed2ad90c7a98d84aa7f8fe010043910b57120e8#diff-e1ff46625134b9f4a5a8424b263484a0R9

spec/javascripts/modules/track-qa-choices.spec.js
steventux: https://github.com/alphagov/finder-frontend/pull/678/files#diff-e1ff46625134b9f4a5a8424b263484a0R52

tombye: 👍

app/assets/javascripts/modules/track-qa-choices.js
tombye: 👍

PR #682, opened 2018-11-23
Add generic description option for finders

rubenarakelyan:
This commit adds an option for finders to display a generic filter description of “X publications matched your criteria” rather than constructing a sentence based on the selected filters.

### DIFF
--- a/app/views/finders/_result_count_generic.mustache
+++ b/app/views/finders/_result_count_generic.mustache
@@ -1,0 +1,8 @@
+{{#any_filters_applied}}
+<p class="result-info">
+  <span class="result-count">
+    {{total}}
+  </span>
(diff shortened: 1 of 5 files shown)


app/views/finders/_result_count_generic.mustache
```diff
@@ -0,0 +1,8 @@
+{{#any_filters_applied}}
+<p class="result-info">
+  <span class="result-count">
+    {{total}}
+  </span>
+  {{{generic_description}}}
```
thomasleese: Is this meant to have three `{{{`?

rubenarakelyan: I guess technically not since it's just a string but in case we add HTML this will allow it to appear correctly.

PR #687, opened 2018-11-27
Topic faceting

bilbof:
This makes it possible to filter by topic and sub topic (top and second level taxons) on the news and communications finder. Once this is merged it will be possible to add this facet to all other finders by updating their content items.

https://trello.com/c/IVd56TDH/132-to-add-a-taxonomy-facet-to-the-news-and-communications-finder

It will probably be easiest to review commit by commit. See the commit messages for more details of the changes.

### Screenshots:

screen shot 2018-11-27 at 13 30 31

screen shot 2018-11-27 at 13 30 15

screen shot 2018-11-27 at 13 30 25


app/lib/filters/checkbox_filter.rb
```diff
@@ -0,0 +1,12 @@
+module Filters
+  class CheckboxFilter < Filter
+    def value
+      # 'params' for a checkbox should be true or false. the value
+      # we send to rummager can be params or a static value provided in
+      # the finder content item
+      if params
+        facet['filter_value'] || params
+      end
+    end
```
bilbof: The filters have all been moved (without changes) into individual class files for readability.

spec/lib/registries/topic_taxonomy_registry_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  let(:content_id_one) { "top-level-taxon-one" }
+  let(:content_id_two) { "top-level-taxon-two" }
+  let(:top_level_taxon_one) { top_level_taxon(content_id_one) }
+  let(:top_level_taxon_two) { top_level_taxon(content_id_two) }
+
+  describe "when topic taxonomy API is unavailable" do
+    it "will return an (uncached) empty hash" do
+      clear_cache
+      topic_taxonomy_api_is_unavailable
```
koetsier: Could use `content_store_isnt_available` from the gds-api-adapters content store helper: https://github.com/alphagov/gds-api-adapters/blob/1dadd526f523a744413e166f42c4cd3f01e3084b/lib/gds_api/test_helpers/content_store.rb#L81

spec/lib/registries/topic_taxonomy_registry_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      expect(fetched_document['title']).to eq(top_level_taxon_two['title'])
+    end
+  end
+
+  def topic_taxonomy_api_is_unavailable
+    stub_request(:get, topic_taxonomy_endpoint).to_return(status: 500)
+  end
+
+  def topic_taxonomy_has_taxons
+    stub_request(:get, topic_taxonomy_endpoint).
```
koetsier: Could use content_store_has_item from the gds-api-adapers test helpers:  https://github.com/alphagov/gds-api-adapters/blob/1dadd526f523a744413e166f42c4cd3f01e3084b/lib/gds_api/test_helpers/content_store.rb#L22

app/views/finders/_taxon_facet.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <select class="govuk-select" id="second_level_taxon" name="second_level_taxon">
+      <% taxon_facet.sub_topics.each do |topic| %>
+        <option
+          data-topic-parent="<%= topic[:parent] %>"
+          value="<%= topic[:value] %>"
+          <% if topic[:selected] %>selected<% end %>>
+          <%= topic[:text] %>
+        </option>
+      <% end %>
+    </select>
```
bilbof: @andysellick FYI I've not used the component here as I need this attribute on the options: `data-topic-parent="<%= topic[:parent] %>"`. Would be interested in your thoughts!

app/lib/registries/base_registries.rb
```diff
@@ -1,5 +1,5 @@
 module Registries
-  NAMESPACE = 'registries'
+  NAMESPACE = Rails.env.test? ? 'test/registries' : 'registries'
```
koetsier: Maybe instead of changing namespaces, you could use a different cache_store?

so in development/test.rb you set config.cache_store = :mem_cache_store
and production.rb config.cache_store = :file_store (the default)

See: https://guides.rubyonrails.org/caching_with_rails.html section 2.4?

app/lib/filters/checkbox_filter.rb
koetsier: 👍

app/views/finders/_taxon_facet.html.erb
andysellick: @bilbof totally reasonable that you would do this, but we can add that option to the component for you. Let's talk offline.

app/lib/filters/taxon_filter.rb
```diff
@@ -0,0 +1,11 @@
+module Filters
+  class TaxonFilter < Filter
+    def value
+      topic = params['top_level_taxon']
```
andrewgarner: Could we make this more generic so that it supports any taxon tree?

Knowledge of how the taxon filtering works in the UI, in the form of a primary and secondary dropdown is leaking into the implementation here.

bilbof:
As part of this deploy I need to remove the cached data for world locations. The registry will check for invalid data and delete the cache as appropriate.

Gemfile
```diff
@@ -42,4 +42,5 @@ group :test do
   gem 'rails-controller-testing'
   gem 'simplecov', '~> 0.16.1'
   gem 'webmock', '~> 3.4.2'
+  gem 'dalli', '~> 2.7.9'
```
bilbof: This is under group test as it is only used for testing locally. Dalli is used for cacheing during tests

app/lib/filters/taxon_filter.rb
bilbof: We can do though for now I think this suits our purposes, as I don't think we plan to make this facet support any other levels of the taxonomy tree.

app/views/finders/_taxon_facet.html.erb
andysellick: @bilbof this option has now been added (as you know). Here's [the documentation](https://govuk-publishing-components.herokuapp.com/component-guide/select/with_data_attributes) for doing it.

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -125,6 +125,11 @@
     .js-enabled & .button {
       display:none;
     }
+
+    .js-required {
```
andrewgarner: The convention I've seen across GOV.UK is to use `js-hidden`.

app/lib/search_query_builder.rb
```diff
@@ -61,7 +61,18 @@ def base_return_fields
...
-    finder_content_item['details']['facets'].map { |f| (f['filter_key'] || f['key']) }
+    finder_content_item['details']['facets'].map { |f|
+      unfilterise(f['filter_key'] || f['key'])
+    }
+  end
+
+  def unfilterise(field)
+    # Removes filter-y prefixes from facet keys.
+    # For example, filter_x or filter_all_x will become x.
+    match = /^(?'full_name'(?'operation'filter|reject|any|all)_(?:(?'multivalue_query'any|all)_)?(?'name'.*))$/.match(field)
```
andrewgarner: I think you could use `gsub` here:

```ruby
field.gsub(/^(?'full_name'(?'operation'filter|reject|any|all)_(?:(?'multivalue_query'any|all)_)?(?'name'.*))$/, '\k<name>')
```

app/models/checkbox_facet.rb
```diff
@@ -1,44 +1,45 @@
...
   end
 
   def value
     facet['value']
   end
 
-  def checked?(checkbox)
-    @value && @value.include?(checkbox.value)
+  def is_checked?
+    !@value.nil? # nil if unchecked
```
andrewgarner: How about `@value.present?` to avoid the double negative?

app/models/dropdown_select_facet.rb
```diff
@@ -0,0 +1,48 @@
...
+      {
+        value: allowed_value['value'],
+        text: allowed_value['text'],
+        selected: selected_value == allowed_value,
+      }
+    end
+  end
+
+  def name
+    "Filter by #{facet['name']}"
```
andrewgarner: This is a view concern. Can we extract to a presenter? How might we handle localisation?

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,94 @@
+class TaxonFacet < DropdownSelectFacet
+  TOP_LEVEL_TAXON_KEY = 'top_level_taxon'.freeze
```
andrewgarner: In the domain of content schemas and links we call this a `level_one_taxon` – how about we adopt that language here too?

spec/helpers/taxonomy_spec_helper.rb
```diff
@@ -0,0 +1,61 @@
+require "gds_api/test_helpers/content_store"
+
+module TaxonomySpecHelper
+  include ::GdsApi::TestHelpers::ContentStore
+
+  CONTENT_ID_1 = "top-level-taxon-one".freeze
```
andrewgarner: "top-level-taxon-one" isn't representative of a typical UUID content id.

sistephens:
Did the labels `'Filter by topic'` and `'Filter by sub-topic'` come from somewhere? And if not should they be `'Topic'` and `'Sub-topic'`?

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
+class TaxonFacet < DropdownSelectFacet
+  TOP_LEVEL_TAXON_KEY = 'level_one_taxon'.freeze
```
andrewgarner:
```suggestion
  LEVEL_ONE_TAXON = 'level_one_taxon'.freeze
```

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
...
+  SECOND_LEVEL_TAXON_KEY = 'level_two_taxon'.freeze
```
andrewgarner:
```suggestion
  LEVEL_TWO_TAXON_KEY = 'level_two_taxon'.freeze
```

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
...
+  def value_fragment(value, key)
+    return nil if value.nil?
+
+    {
+      'label' => value[:text],
+      'parameter_key' => key
+    }
+  end
+
+  def top_level_taxons
```
andrewgarner:
```suggestion
  def level_one_taxons
```

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
...
+      {
+        value: v['content_id'],
+        text: v['title'],
+        sub_topics: v['children'],
+        selected: v['content_id'] == @value[TOP_LEVEL_TAXON_KEY]
+      }
+    }
+  end
+
+  def second_level_taxons
```
andrewgarner:
```suggestion
  def level_two_taxons
```

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
...
+      }
+  end
+
+  def selected_sub_value
+    @selected_sub_value ||= second_level_taxons.find { |v|
+      v[:value] == @value[SECOND_LEVEL_TAXON_KEY]
+    }
+  end
+
+  def selected_value
```
andrewgarner:
```suggestion
  def selected_level_one_value
```

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
...
+        {
+          text: v['title'],
+          value: v['content_id'],
+          parent: v['parent'],
+          selected: v['content_id'] == @value[SECOND_LEVEL_TAXON_KEY]
+        }
+      }
+  end
+
+  def selected_sub_value
```
andrewgarner:
```suggestion
  def selected_level_two_value
```

app/models/taxon_facet.rb
```diff
@@ -0,0 +1,92 @@
+class TaxonFacet < DropdownSelectFacet
+  TOP_LEVEL_TAXON_KEY = 'level_one_taxon'.freeze
+  SECOND_LEVEL_TAXON_KEY = 'level_two_taxon'.freeze
+
+  def topics
```
andrewgarner: This is a `TaxonFacet` but it returns `topics`. I wonder whether we're leaking in too much context about how that GOV.UK taxonomy is structured. You need to know that the taxons are topics.

Maybe we could change the API to something like `level_one_values` and `level_two_values`?

app/views/finders/_taxon_facet.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="app-taxonomy-select">
+  <%= render "govuk_publishing_components/components/select", {
+    id: 'level_one_taxon',
+    label: "Filter by topic",
+    options: taxon_facet.topics
+  }
+  %>
+  <div class="js-required govuk-form-group gem-c-select">
```
andrewgarner: Does this need a `js-hidden` class now too?

app/models/taxon_facet.rb
bilbof: 👍

app/models/taxon_facet.rb
bilbof: 👍

app/views/finders/_taxon_facet.html.erb
bilbof: Looks like there is some different behaviour clashing with js-hidden, so I'm not going to use that class.

PR #703, opened 2018-12-03
Adds UK nations in the EU triage question / answer to content page

DilwoarH:
https://trello.com/c/qPFClWF6/16-build-qa-screen-for-country-selection

## Screenshot
screenshot-finder-frontend dev gov uk-2018 12 04-12-25-38

### DIFF
--- a/app/controllers/qa_to_content_controller.rb
+++ b/app/controllers/qa_to_content_controller.rb
@@ -1,0 +1,70 @@
+require 'gds_api/helpers'
+
+# This class currently supports only one question
+# If there is a need for multiple questions, this class needs to be modified
+
+class QaToContentController < ApplicationController
+  layout "finder_layout"
+  include GdsApi::Helpers
+
+  def show
+    return error_not_found unless ENV["FINDER_FRONTEND_ENABLE_QA_TO_CONTENT"]
+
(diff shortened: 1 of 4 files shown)


app/controllers/qa_to_content_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      redirect_to_guidance params[first_question["id"]]
+    else
+      render 'qa_to_content/show'
+    end
+  end
+
+private
+
+  def qa_config
+    @qa_config ||= YAML.load_file("lib/#{request.path.tr('-', '_')}.yaml")
```
thomasleese: I wonder if we need some protection here in case someone puts `../../` in the URL and manages to read a YAML file from somewhere else. I imagine there isn't much of a risk here given it has to be a valid YAML file, but I believe YAML files can contain executable code blocks.

DilwoarH: I think it will be fine because the routes are defined and if the user adds any routes which don't match it will through a 404.

Unless we start allowing dynamic routes - in which case, what you are saying might be possible.

thomasleese: Good point!

PR #704, opened 2018-12-03
Redirect final Q&A page submission to the finder page

edwardkerry:
If the final question in a Q&A is a `yes-no` radio button option with nested checkboxes, currently choosing some `yes` checkboxes and then choosing the `no` radio button will submit all choices directly to the Finder.

This change adds the ability for the Q&A view to submit the last Q&A response with an incremented page param. This allows the Q&A controller to recognise the last facet has been answered, to filter the submitted parameters to remove any `yes-no` discrepancies, and redirect to the finder page.

### DIFF
--- a/app/controllers/qa_controller.rb
+++ b/app/controllers/qa_controller.rb
@@ -7,6 +7,7 @@ class QaController < ApplicationController
   def show
     return error_not_found unless ENV["FINDER_FRONTEND_ENABLE_QA"]
 
+    redirect_to_finder if finder_page?
     raw_finder
   end
 
@@ -28,8 +29,7 @@ private
 
   def page
     params.permit(:page)
-    params[:page] = 1 if params[:page].nil?
-    params[:page].to_i.clamp(1, facets.length)
+    params[:page].to_i.clamp(1, facets.length + 1)
   end
 
   def question_type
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -10,3 +10,6 @@ Metrics/BlockLength:
   Exclude:
     - 'features/**/*.rb'
     - 'spec/**/*'
+
+Style/AndOr:
+  EnforcedStyle: conditionals
```
AlanGabbianelli: This is the expected behaviour. See our styleguide: https://github.com/alphagov/govuk-lint/blob/master/configs/rubocop/gds-ruby-styleguide.yml#L127-L130

app/controllers/qa_controller.rb
```diff
@@ -29,7 +29,7 @@ def facets
   def page
     params.permit(:page)
     params[:page] = 1 if params[:page].nil?
-    params[:page].to_i.clamp(1, facets.length)
+    params[:page].to_i.clamp(1, facets.length + 1)
```
AlanGabbianelli: Is `params[:page] = 1 if params[:page].nil?` needed? Even if `params[:page]` is `nil` the next line will return `1`:
```
[1] pry(main)> nil.to_i.clamp(1,5)
=> 1
```

app/controllers/qa_controller.rb
```diff
@@ -96,25 +97,26 @@ def nested_options(filter_group)
...
+    page == facets.length + 1
+  end
 
+  def next_page_url
     qa_config["base_path"]
   end
   helper_method :next_page_url
 
+  def redirect_to_finder
+    redirect_to qa_config["finder_base_path"] + '?' + filtered_params.to_query and return
```
barrucadu: Is the `and return` needed?

edwardkerry: I thought that this would ensure that we stopped controller execution, but as there is no other option it doesn't seem to make any difference in this case, and also resolves the Rubocop problem @AlanGabbianelli wrote about below.

I'll remove it 👍

app/controllers/qa_controller.rb
edwardkerry: nice one, will remove

.rubocop.yml
edwardkerry: There's some discussion about why this rule is not quite right for `and return` situations here:
https://github.com/rubocop-hq/rubocop/issues/5550
but, @barrucadu has suggested another way round in this case anyway 😄

PR #725, opened 2018-12-07
Support 'or' combine_mode facets

thomasleese:
This changes finder-frontend to use the Rummager batch search endpoint to perform the searches. 

- When the query should be an AND across the facets (which is the default behaviour) it just performs a single search as before but using the new endpoint.
- When the query should be an OR across the facets, it performs a query for each facet and then merges and de-duplicates the results.

Recommended to look at the diff without whitespace changes.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/848.

### DIFF
--- a/app/lib/search_query_builder.rb
+++ b/app/lib/search_query_builder.rb
@@ -7,17 +7,19 @@ class SearchQueryBuilder
   end
 
   def call
-    [
-      [
-        pagination_query,
-        return_fields_query,
-        keyword_query,
-        filter_query,
-        reject_query,
-        order_query,
-        facet_query,
-      ].reduce(&:merge)
-    ]
+    base_query = [
+      pagination_query,
+      return_fields_query,
+      keyword_query,
+      filter_query,
+      reject_query,
+      order_query,
+      facet_query,
+    ].reduce(&:merge)
+
+    [base_query] + extra_filter_queries.map do |query|
+      base_query.clone.merge(query)
+    end
   end
 
 private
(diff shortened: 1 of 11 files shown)


app/lib/search_query_builder.rb
```diff
@@ -116,12 +122,28 @@ def default_order
...
-  def filter_query
-    filter_params
-      .merge(base_filter)
-      .reduce({}) { |query, (k, v)|
-        query.merge("filter_#{k}" => v)
-      }
+  def base_filter_query
+    @base_filter_query ||= base_filter.each_with_object({}) do |(k, v), query|
+      query["filter_#{k}"] = v
+    end
```
barrucadu: Should this include the `and_filter_query`?

app/lib/search_query_builder.rb
```diff
@@ -116,12 +122,28 @@ def default_order
...
+
+  def or_filter_queries
+    @or_filter_queries ||= filter_params(combine_mode: 'or')
+      .map do |k, v|
+        { "filter_#{k}" => v }
+      end
+  end
+
+  def filter_queries
+    (and_filter_query.empty? ? [] : [and_filter_query]) + or_filter_queries
```
barrucadu: If the `and_filter_query` is put into the `base_filter_query`, I think `filter_queries` here can be dropped in favour of just using `or_filter_queries` directly.

app/lib/search_query_builder.rb
thomasleese: It shouldn't. We've changed the way the finder works so that the `or` facets are now completely independent of the `and` facets - after much discussion we've concluded this logic makes the most sense. Therefore, this `base_filter_query` (which is the base for both the `and` and the `or` queries) needs to only contain the finder base filter and nothing else.

thomasleese: We also realised that if the `and` query was to be included as the base query for each of the `or` queries, ticking any `or` facets wouldn't surface any more results as they would have already been included in the base `and` query which would've had a wider criteria than the `and` + `or` query.

barrucadu: Ok, so the desired logic now is: all AND facets hold, or at least one OR facet holds?

thomasleese: Yeah, that's correct. It's basically: `(and_facet_1 AND and_facet_2) OR or_facet_1 OR or_facet_2 OR ...`

barrucadu: Maybe instead of the "X results for businesses which Yes" we should instead say "X results for (facet_1 && ...) || facet_3 || ..."!

PR #726, opened 2018-12-07
Use grid in finders

andysellick:
Use standard grid for finder layouts, rather than custom classes. This should result in no changes to the appearance of finders but will enable us to remove some unnecessary CSS and also iterate the layout more easily in future.


vanitabarrett:
I'm just restarting the review app on heroku, as it didn't seem to create one automatically ^

PR #729, opened 2018-12-10
Omit search box when there are no results

sihugh:
We want the search box to be visible if there:
 - are results
 - are keywords specified
 - are other facets present that may have filtered the results to zero

We're trying to reduce the number of times the search box is present when it is useless.  A drawback is that we don't want it to keep appearing and disappearing without an obvious reason for it.  For this reason, we're being a little cautious about the scenarios in which this occurs.

I suspect further user testing might allow us to expand on this.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -149,6 +149,10 @@ class FinderPresenter
     filters.map(&:sentence_fragment).compact
   end
 
+  def show_keyword_search?
+    keywords.present? || facets.any? || results.total.positive?
+  end
+
   def show_summaries?
     content_item['details']['show_summaries']
   end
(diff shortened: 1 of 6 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -149,6 +149,10 @@ def filter_sentence_fragments
     filters.map(&:sentence_fragment).compact
   end
 
+  def show_keyword_search?
+    keywords.present? || facets.any? || results.total.positive?
+  end
```
bilbof: Yes it looks like this will cover all the scenarios where the results could be filtered by the user.

I suppose this could also be used for showing a different response message, as it might be confusing to have no search input without an explanation as to why.

sihugh: Yeah, we'll have a think about that 🤔

PR #730, opened 2018-12-10
increase the unicorn worker count to 4

JonathanHallam:
This commit increases the number of unicorn workers from 2 to 6 per instance. This allows more requests to be handled simultaneously. This affects proper finders and advanced search and should make them able to handle larger numbers of unique requests at the same time 

### Testing
This was tested using gatling on staging. In general it has helped to lower the upper limits of response time. The following screenshots were from data following tests involving ~5k requests using about 500 unique URL's

Before
screen shot 2018-12-10 at 13 57 46

After
screen shot 2018-12-10 at 13 57 57

### DIFF
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -1,2 +1,4 @@
 require "govuk_app_config/govuk_unicorn"
 GovukUnicorn.configure(self)
+
+worker_processes 4


thomasleese:
The results looks good!

Ideally, this would be done in Puppet to enabling configuring the number of unicorn workers per environment. For example, the way we configure it for `government-frontend`:

https://github.com/alphagov/govuk-puppet/blob/f182779c220fa0855a883164056843dab69cd3d4/hieradata_aws/common.yaml#L490

thomasleese:
@JonathanHallam Yeah, I think it's fine to assume that. Doing `govuk::apps::finder_frontend::unicorn_worker_processes: '4'` is equivalent to `worker_processes 4`.

JonathanHallam:
@thomasleese Awesome, thanks, will close this off. https://github.com/alphagov/govuk-puppet/pull/8431

PR #734, opened 2018-12-11
[DO NOT MERGE] update uk nationals to uk citizens

DilwoarH:

PR #735, opened 2018-12-11
Remove "in the" and "in" from nationals in the eu routes

bevanloon:
Related - https://github.com/alphagov/finder-frontend/pull/734 - but I think this is the actual change we want as at 9:40am 11 December.
Still needs a corresponding PR in rummager.

### DIFF
--- a/lib/uk_nationals_living_in_the_eu.yaml
+++ b/lib/uk_nationals_living_in_the_eu.yaml
@@ -1,5 +1,5 @@
 ---
-base_path: "/uk-nationals-in-eu"
+base_path: "/uk-nationals-eu"
 title: UK nationals in the EU
 questions:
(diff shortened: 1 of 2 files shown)


nicholsj:
Corresponding rummager PR https://github.com/alphagov/rummager/pull/1369/files

lib/uk_nationals_living_in_the_eu.yaml
```diff
@@ -1,5 +1,5 @@
 ---
```
edwardkerry: I think this file may need re-naming to match the path.
 The controller loads it as:
``` 
def qa_config
    @qa_config ||= YAML.load_file("lib/#{request.path.tr('-', '_')}.yaml")
  end
```

PR #739, opened 2018-12-12
Add facet tags to finders

vanitabarrett:
Authored by @oscarwyatt @karlbaker02 and @vanitabarrett

Trello: https://trello.com/c/JAYPFEEn/140-update-styling-to-the-facet-tags

## Desktop
screen shot 2018-12-19 at 15 48 43

## Mobile
screen shot 2018-12-19 at 15 48 54

https://finder-frontend-pr-739.herokuapp.com/news-and-communications


app/presenters/result_set_presenter.rb
```diff
@@ -43,50 +43,25 @@ def generic_description
...
-    }
-
-    if fragment['type'] == "text"
-      values.to_sentence(two_words_connector: ' or ', last_word_connector: ' or ')
-    elsif fragment['type'] == "date" || fragment['type'] == "checkbox"
-      values.to_sentence(two_words_connector: ' and ')
+    all_filter_params = {}
+    selected_filters.each do |filter|
+      all_filter_params[filter.key] = filter.value
     end
```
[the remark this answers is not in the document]
bilbof: Alternatively, [each_with_object](https://apidock.com/rails/Enumerable/each_with_object) is useful.

`all_filter_params = selected_filters.each_with_object({}) { |filter, result| result[filter.key] = filter.value }`

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -0,0 +1,37 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
+
+  GOVUK.Modules.RemoveFilter = function () {
```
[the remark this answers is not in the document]
vanitabarrett: It's how we do our JS modules: it's called in one of the views by adding `data-module="remove-filter"` as an attribute to a HTML element.

Will add tests 👍

app/models/select_facet.rb
```diff
@@ -30,12 +30,26 @@ def sentence_fragment
...
       'values' => value_fragments,
+      'word_connectors' => or_word_connectors
     }
   end
 
+  def has_filters?
+    selected_values.any?
+  end
+
+  def description
```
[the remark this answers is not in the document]
vanitabarrett: Doesn't look like it is being used... Will delete for now.

app/presenters/advanced_search_result_set_presenter.rb
```diff
@@ -41,9 +41,9 @@ def applied_filters_or_all_subgroups
       if describe_filters_in_sentence.blank?
         subgroups_as_sentence
       elsif subgroup_facet.value.blank?
-        [subgroups_as_sentence, describe_filters_in_sentence].to_sentence
+        [subgroups_as_sentence, filters_to_sentence(describe_filters_in_sentence)].to_sentence
       else
-        describe_filters_in_sentence
+        [filters_to_sentence(describe_filters_in_sentence)].to_sentence
```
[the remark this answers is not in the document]
vanitabarrett: `describe_filters_in_sentence` (or now `selected_filter_descriptions`) is also used in the `result_set_presenter` in the `to_hash` method

app/presenters/facet_filter_presenter.rb
```diff
@@ -0,0 +1,49 @@
+class FacetFilterPresenter
```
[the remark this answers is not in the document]
vanitabarrett: Good catch. I've renamed it to `FacetTagPresenter` for now, but welcome to other suggestions 🙂

app/models/facet.rb
```diff
@@ -29,5 +29,17 @@ def metadata?
 
 private
 
+  def and_word_connectors
+    { two_words_connector: ' and ' }
+  end
```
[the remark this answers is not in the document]
vanitabarrett: I've had a look at the other finders and they seem to look ok. We can always adjust wording later if we find it performs badly in user research, for example.

app/models/select_facet.rb
```diff
@@ -46,11 +60,16 @@ def unselected?
 
 private
 
+  def description_sentence
+    value_fragment_labels.to_sentence(two_words_connector: ' or ', last_word_connector: ' or ')
```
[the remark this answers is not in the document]
vanitabarrett: Outdated - this method has been removed

app/assets/javascripts/live_search.js
```diff
@@ -30,6 +30,7 @@
       this.$form.find('input[type=checkbox], input[type=text], input[type=radio], select').on('change',
         function(e) {
           if (e.target.type == "text") {
+            $(e.target).attr("value", e.target.value);
```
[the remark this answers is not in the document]
vanitabarrett: @karlbaker02 I've had a look but I can't see the purpose of this line - are you able to shed some light?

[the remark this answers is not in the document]
bilbof: I don't think it's necessary. The value being assigned to the target is sourced from the target:

```
var html = '<input type="text" value="hello">'

$(html).attr('value', html.value)
```

app/assets/javascripts/modules/remove-filter.js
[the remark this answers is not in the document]
bilbof: Thanks for explaining!

app/assets/javascripts/live_search.js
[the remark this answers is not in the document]
vanitabarrett: Agreed, it doesn't look like setting the value is needed, I'll remove it for now.

@karlbaker02 Please correct if we're wrong!

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -0,0 +1,56 @@
+window.GOVUK = window.GOVUK || {}
```
andysellick: Nitpick: there's a lot of lines in this file without a semi colon at the end.

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -233,16 +233,91 @@
...
         margin-right: $gutter-one-third / 2;
       }
     }
 
+    .filters {
+      margin-top: $gutter-one-third;
+      margin-bottom: $gutter;
+    }
+
+    .facet-tag-group {
```
andysellick: Did you consider BEM for this naming convention? I think it would work better, and would reduce the need to nest some of this CSS.

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -233,16 +233,91 @@
...
+        @include media(tablet) {
+          display: table-cell;
+        }
+
+        .facet-tag-text {
+          display: inline-block;
+          margin-left: 5px;
+        }
+
+        .remove-filter {
```
andysellick: The remove button makes for quite a small click target (on desktop and mobile). Did you consider making the entire facet tag clickable to remove?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -233,16 +233,91 @@
...
+        font-weight: bold;
+
+        &::first-letter {
+          text-transform: uppercase;
+        }
+      }
+
+      .facet-tag-preposition {
+        padding: 0 5px 0 2px;
+        vertical-align: middle;
```
andysellick: Spacing between this and the facet tag on mobile is quite close, might be good to space it out a little more.

screen shot 2018-12-18 at 16 27 34

app/views/finders/_applied_filter.mustache
```diff
@@ -0,0 +1,18 @@
+<div class="facet-tag-group">
+{{#.}}
+	<div class="facet-tag-wrapper">
```
andysellick: Nitpick: indentation on this file seems too big?

app/assets/javascripts/modules/remove-filter.js
vanitabarrett: I had a feeling we didn't use semi-colons in our JS? But I agree we should probably pick one or the other in this file.

app/assets/stylesheets/finder_frontend.scss
vanitabarrett: I had already added some extra padding around the button to increase the clickable area from the default size. I think making the whole tag clickable to remove is probably a separate conversation and shouldn't be blocking this PR.

app/assets/stylesheets/finder_frontend.scss
vanitabarrett: Good spot 👍

vanitabarrett: Done

app/views/finders/_applied_filter.mustache
vanitabarrett: Yeah, this is strange... it looks ok in my editor (which uses the correct spacing for every other file). I've tried re-indenting but it doesn't show up as a change, so my editor at least thinks it's correct.

app/assets/stylesheets/finder_frontend.scss
andysellick: Let's make a note for when we look into turning this into a component.

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -308,4 +306,82 @@
...
+
+    @include media(tablet) {
+      display: inline-block;
+    }
+  }
+
+  .facet-tags__wrapper:first-of-type .facet-tags__preposition:first-of-type {
+    font-weight: bold;
+
+    &::first-letter {
```
andysellick: Is this rule necessary? Looking at the source of the example heroku page it seems to be coming through with an uppercase first letter already.

I have a similar question about line 343 `text-transform: lowercase;` but I guess this is related?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -308,4 +306,82 @@
...
+  .facet-tags__wrapper {
+    vertical-align: middle;
+    margin: 5px 0;
+
+    @include media(tablet) {
+      display: inline-block;
+    }
+  }
+
+  .facet-tags__wrapper:first-of-type .facet-tags__preposition:first-of-type {
```
andysellick: Does this rule need to apply to `.facet-tags__wrapper:first-of-type`? It seems like the bold font is only needed on `facet-tags__preposition`?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -308,4 +306,82 @@
...
+    }
+  }
+
+  .facet-tags__preposition {
+    padding: 0 5px 5px 2px;
+    vertical-align: middle;
+    text-transform: lowercase;
+
+    @include media(tablet) {
+      display: table-cell;
```
andysellick: I think it would be better to consistently use display inline-block rather than table-cell, here and on line 364. Mainly because inline-block is standalone but table-cell is part of display table, so it could be unpredictable on some browsers? I have no evidence for that though.

app/views/finders/_applied_filter.mustache
vanitabarrett: Creating a new mustache file and deleting the old one has fixed it.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -0,0 +1,56 @@
...
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
```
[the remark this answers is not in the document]
vanitabarrett: No, have removed 👍

app/assets/stylesheets/finder_frontend.scss
vanitabarrett: Yes, the news and communications finder only has facet labels that are a single word, so it isn't needed. But other finders, like [aaib-reports](https://finder-frontend-pr-739.herokuapp.com/aaib-reports?keywords=&date_of_occurrence%5Bfrom%5D=2018&date_of_occurrence%5Bto%5D=), have the label for date as "Occurred After" (for example).

app/assets/stylesheets/finder_frontend.scss
vanitabarrett: This was to help stop awkward text wrapping, like this:

screen shot 2018-12-20 at 11 34 50

I initially tried to do this using `white-space` and `word-wrap` but it got very complicated and was a bit buggy, so this seemed like the simplest solution. Happy to try another way though, if you have something in mind 🙂

app/assets/stylesheets/finder_frontend.scss
vanitabarrett: Doing just `facet-tags__preposition:first-of-type` applies the styling to every `facet-tags__preposition` that is the first child of it's parent element, whereas we only want to add the bold styling to the first `facet-tags__preposition` within the first `facet-tags__wrapper`.

Otherwise you end up with this:
screen shot 2018-12-20 at 11 48 22

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -0,0 +1,51 @@
...
+    this.start = function (element) {
+      $(element).on('click', '[data-module="remove-filter-link"]', toggleFilter);
+    };
+
+    function toggleFilter(e) {
+      e.preventDefault();
+
+      var removeFilterName = $(this).data("name");
+      var removeFilterValue = $(this).data("value");
+      var removeFilterFacet = $(this).data("facet");
```
andysellick: Just noticed here we're using `"` but in other lines we're using `'` - sorry, small thing, shouldn't block the PR.

karlbaker02:
This actually seems somewhat off the original design - have I missed anything? Currently the _Brexit_ filter is not alongside other topics, people and organisation filters are displayed separately and it seems we're missing a bunch of prepositions?

**Original Design:**
Original design

**Actual:**
screen shot 2018-12-21 at 10 46 23

vanitabarrett:
This is ready to merge, but I'm going to hold off for now until after Xmas / more devs will be in.

sihugh:
I'm going to merge this, as it's quite big already.  We'll do a design review of it and write up a new card

PR #740, opened 2018-12-12
Adds accessible autocomplete facet to finders

oscarwyatt:
Handles cases where the input loses focus after the value has been removed (i.e. the user no longer
wants to filter by that value) and must remove the filter value
Also fixes some linting issues

Trello: https://trello.com/c/fkbJr5kX/173-add-autocomplete-to-finders


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -157,6 +157,10 @@
     width:14px;
   }
 
+  .select-filter {
+    margin-bottom: $gutter-one-third;
```
andysellick: I'm going to remove this because the autocomplete component should come with it's own default bottom margin and it doesn't, which I'm correcting in https://github.com/alphagov/govuk_publishing_components/pull/680

spec/javascripts/live_search_spec.js
```diff
@@ -297,4 +297,18 @@ describe("liveSearch", function(){
...
+    });
+
+    it('clears an accessibleAutocomplete if blur on an empty input', function() {
```
andysellick: This test is currently failing, however I'm not sure it's necessary. Components should be isolated, which means they should have their own tests. If this test is needed and isn't part of the component, we should look at adding it to the component and not here.

app/assets/javascripts/templates.js
```diff
@@ -0,0 +1,10 @@
+window.templates = {};
```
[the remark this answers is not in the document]
andysellick: `templates.js` shouldn't have been added to this PR - it's an auto generated file that gets updated automatically under certain conditions (not sure which, but it changes itself all the time on my local). Unfortunately I think it's necessary so presumably we can't add it to the gitignore. We'll have to remove it from this PR @oscarwyatt

PR #741, opened 2018-12-14
Topic faceting improvements

bilbof:
This fixes an issue in Safari where unavailable sub-topic options aren't removed from the sub-topic select.

Safari [does not support](https://stackoverflow.com/a/15025961) `$(option).toggle()`. So rather than trying to hide unavailable sub-topics, we need to remove them from the DOM.

### DIFF
--- a/app/assets/javascripts/taxonomy-select.js
+++ b/app/assets/javascripts/taxonomy-select.js
@@ -9,12 +9,13 @@
 
   function TaxonomySelect(options) {
     this.$el = options.$el;
+    this.options = this.instantiateOptions();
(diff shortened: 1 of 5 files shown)


sihugh:
Where's the review app? I need to find the handle

bilbof:
@sihugh usually it'd be here: https://finder-frontend-pr-741.herokuapp.com/news-and-communications 😕

andrewgarner:
screen shot 2018-12-17 at 08 39 07

bilbof:
Closing for Heroku.

PR #742, opened 2018-12-17
Style checkboxes to be smaller in size

alex-ju:
I'm aware there is a new version of [small version for checkboxes](https://github.com/alphagov/govuk-design-system-backlog/issues/162) coming from the Design System, but until that is ready I'd add these temporary overwrites. When the new ones will be ready and implemented in the [shared library of components](https://github.com/alphagov/govuk_publishing_components) this code will be removed and we will rely on the [Design System](https://github.com/alphagov/govuk-frontend) code.

@miaallers @markhurrell I'd like your input for this quick fix when you have time and I'm happy to discuss how can we further improve this component (i.e. replacing the top-right chevron – currently a stretched raster image).

### Before
Default
screen shot 2018-12-17 at 09 38 23

Selected
screen shot 2018-12-17 at 09 38 54

### After
Default
screen shot 2018-12-17 at 09 38 38

Selected
screen shot 2018-12-17 at 09 39 01

### DIFF
--- a/app/assets/stylesheets/components/_option-select.scss
+++ b/app/assets/stylesheets/components/_option-select.scss
@@ -103,4 +103,44 @@
       }
     }
   }
+
+  // This is a temporary overwrite for checkboxes
+  // These styles should be removed in favour of small checkboxes from the Design System
+  .govuk-checkboxes__item {
+    min-height: 26px;
+    margin-bottom: 10px;
+    padding: 0 0 0 30px;
+  }
+
+  .govuk-checkboxes__item:last-child,
+  .govuk-checkboxes__item:last-of-type {
+    margin-bottom: 0;
+  }
+
+  .govuk-checkboxes__input {
+    top: 5px;
+    left: 0;
+    width: 26px;
+    height: 26px;
+  }
+
+  .govuk-checkboxes__label {
+    padding: 8px 5px 2px;
+    font-size: 16px;
+  }
+
+  .govuk-checkboxes__input + .govuk-checkboxes__label::before {
+    top: 5px;
+    width: 26px;
+    height: 26px;
+  }
+
+  .govuk-checkboxes__input + .govuk-checkboxes__label::after {
+    top: 12px;
+    left: 5.5px;
+    width: 12px;
+    height: 5px;
+    border-width: 0 0 3px 3px;
+  }
+
 }


andysellick:
I think the plan is to replace this component entirely with a multi select accessible autocomplete at some point.

miaallers:
I think the updates look good. The only tweak I would make would be to lower the chevron to sit on the same baseline as the box title.

As @andysellick said, the plan is to replace the component with an autocomplete, but in the meantime this is a nice change. Thanks.

app/assets/stylesheets/components/_option-select.scss
```diff
@@ -103,4 +103,44 @@
       }
     }
   }
+
+  // This is a temporary overwrite for checkboxes
+  // These styles should be removed in favour of small checkboxes from the Design System
+  .govuk-checkboxes__item {
+    min-height: 26px;
```
andysellick: This `min-height` seems to be overriding the larger min-height already defined, but does it need to be specified as 26px? Seems to work okay if it's reset, i.e. set to 0, but there may be a browser/usage case I'm unaware of. Just curious.

app/assets/stylesheets/components/_option-select.scss
```diff
@@ -103,4 +103,44 @@
...
+    margin-bottom: 10px;
+    padding: 0 0 0 30px;
+  }
+
+  .govuk-checkboxes__item:last-child,
+  .govuk-checkboxes__item:last-of-type {
+    margin-bottom: 0;
```
andysellick: Is this necessary? Looking at it on https://finder-frontend-pr-742.herokuapp.com/component-guide/option-select there's already a rule in place that does this, not sure where it's coming from though.

app/assets/stylesheets/components/_option-select.scss
```diff
@@ -103,4 +103,44 @@
...
+  .govuk-checkboxes__input {
+    top: 5px;
+    left: 0;
+    width: 26px;
+    height: 26px;
+  }
+
+  .govuk-checkboxes__label {
+    padding: 8px 5px 2px;
+    font-size: 16px;
```
andysellick: Should we be using a mixin for font sizes or is that not appropriate here?

alex-ju: Yep, all values, including other spacing values would ideally use variables and mixins, but I didn't bother since there's no precedent in this repo and this is a 'temporary' code.

app/assets/stylesheets/components/_option-select.scss
alex-ju: It's coming from the overwritten version. Doesn't need to be here too, I can remove this.

app/assets/stylesheets/components/_option-select.scss
alex-ju: Good question. Maybe not, but I needed to overwrite the existing one anyway.

PR #743, opened 2018-12-17
Topic faceting improvements

bilbof:
This fixes an issue in Safari where unavailable sub-topic options aren't removed from the sub-topic select.

Safari [does not support](https://stackoverflow.com/a/15025961) `$(option).toggle()`. So rather than trying to hide unavailable sub-topics, we need to remove them from the DOM.

Review app: https://finder-frontend-pr-743.herokuapp.com

### DIFF
--- a/app/assets/javascripts/taxonomy-select.js
+++ b/app/assets/javascripts/taxonomy-select.js
@@ -9,12 +9,13 @@
 
   function TaxonomySelect(options) {
     this.$el = options.$el;
+    this.options = this.instantiateOptions();
   }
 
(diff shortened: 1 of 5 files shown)


andrewgarner:
Should we apply an order the select options? Maybe A-Z?

screen shot 2018-12-17 at 10 32 34

andrewgarner:
Should the topic be included in the summary if you have chosen a sub-topic?

I would expect it to say "1,000+ documents about **Rail** sorted by Most viewed" not "1,000+ documents about **Transport** and **Rail** sorted by Most viewed".

PR #745, opened 2018-12-17
Use updated checkboxes component in option select

andysellick:
Checkbox component has been retired and merged into checkboxes component. This PR updates finder-frontend to use the checkboxes component in the option select component.

This PR will not pass tests until a version of the publishing components gem has been released that contains [this PR](https://github.com/alphagov/govuk_publishing_components/pull/659) and should not be merged until the publishing components gem version has been bumped to include this change and added here. **<- this has now been done, we're all good** 👍 

Examples of checkbox usage:

- https://finder-frontend-pr-745.herokuapp.com/news-and-communications
- https://finder-frontend-pr-745.herokuapp.com/prepare-business-uk-leaving-eu
- https://finder-frontend-pr-745.herokuapp.com/cma-cases
- https://finder-frontend-pr-745.herokuapp.com/administrative-appeals-tribunal-decisions


app/assets/stylesheets/components/_option-select.scss
```diff
@@ -3,6 +3,15 @@
   padding: 5px;
   margin-bottom: $gutter;
 
+  .govuk-hint {
+    position: absolute;
+    left: -9999em;
+    top: auto;
+    width: 1px;
+    height: 1px;
+    overflow: hidden;
```
bilbof: I'm not able to see the hint in the demo app (with and without JS). Do you know where I can find it? I can see it in the DOM, though! :)

```
<span id="checkboxes-4259dd0a-hint" class="govuk-hint">Select all that apply.</span>
```

app/views/finders/_select_facet.html.erb
```diff
@@ -4,7 +4,7 @@
   :title => select_facet.name,
   :aria_controls_id => "js-search-results-info",
   :options_container_id => select_facet.key,
-  :options => select_facet.options,
+  :options => select_facet.options("js-search-results-info", select_facet.key),
```
bilbof: It looks like we could avoid options requiring params. Is this the only place `select_facet.options` is used?

app/assets/stylesheets/components/_option-select.scss
vanitabarrett: Yes - is this intentionally being hidden?

vanitabarrett: Is there a reason why you chose this method and not applying the `visuallyhidden` class?

andysellick: Yeah, I can't put that class on that element because it's part of the component. I wanted to use a mixin but the govuk one didn't work. I thought it was easiest to put it in longhand and hope that someone would point out a solution :)

vanitabarrett: Good point, didn't spot that. The only solutions I can think of are 1) making it an option in the component, but I'm not sure how often this use case will come up; or 2) putting a separate hint outside of the component, perhaps where we already indicate '[x] selected' when people select checkboxes? But not sure if moving it makes it less useful.

andysellick: I thought about making it an option in the component but if removed it causes an axe accessibility error. But the design for the finders doesn't have it in, so I decided the best thing to do was this i.e. hide it except from screen readers.

These things are going to be (mostly) replaced by accessible autocompletes shortly anyway.

app/views/finders/_select_facet.html.erb
andysellick: @bilbof not quite sure what you're suggesting but agree that this probably isn't the cleanest way of doing it. Let's catch up and discuss it in person.

app/assets/stylesheets/components/_option-select.scss
andysellick: This PR https://github.com/alphagov/govuk_publishing_components/pull/684 updates the checkboxes component so that we can hide the hint text using options, so once that's merged I'll modify this PR to use it accordingly, should hopefully deal with this issue.

app/views/finders/_select_facet.html.erb
andysellick: @sihugh Bill's on leave now and I'm not sure I understand this comment. If you have time could you look at this PR and see if it's okay?

andysellick:
@vanitabarrett the components gem has now been upgraded in this PR to include the options for visually hiding the checkboxes legend, which I think addresses your concern. Do you want to take a look?

app/assets/stylesheets/components/_option-select.scss
```diff
@@ -3,6 +3,15 @@
   padding: 5px;
   margin-bottom: $gutter;
 
+  .govuk-hint {
```
vanitabarrett: This can be removed now ?

andysellick:
@vanitabarrett are you happy to approve this PR? The comment from Bill is still outstanding, trying to get some advice on whether it needs changing.

andysellick:
@vanitabarrett thanks, think I'll wait til after Christmas before merging, have made a note on the card.

PR #751, opened 2018-12-19
Add tracking to taxonomy facet

bilbof:
https://trello.com/c/5iPpo7IR/193-add-tracking-to-topic-subtopic-facet

This adds tracking attributes to select options on level one and two taxons.

End users won't see a difference as a result of this change.


andysellick:
@vanitabarrett is correct, checkboxes don't automatically fire tracking events in the same way that links do, so some additional JS will be needed to make this work. This wasn't added to the component but maybe it should be? Happy to discuss.

PR #756, opened 2018-12-20
Fix bug with displayed sort option on Sort By select box

karlbaker02:
This PR fixes the Sort By select box to display the correct sort option when the page is refreshed.

Trello card: https://trello.com/c/zdcfNCoZ

Solo: @karlbaker02

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -136,7 +136,7 @@ class FinderPresenter
 
     disabled_option = keywords.blank? ? relevance_sort_option_value : ''
 
-    selected_option = values['order'] unless values['order'].nil? && sort.detect { |option| option['name'].parameterize == values['order'] }.nil?
+    selected_option = values['order'] unless values['order'].nil? || sort.detect { |option| option['name'].parameterize == values['order'] }.nil?
     selected_option ||= default_sort_option_value
 
     options_for_select(options, selected: selected_option, disabled: disabled_option)
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -136,8 +136,11 @@ class FinderPresenter
 
     disabled_option = keywords.blank? ? relevance_sort_option_value : ''
 
-    selected_option = values['order'] unless values['order'].nil? || sort.detect { |option| option['name'].parameterize == values['order'] }.nil?
-    selected_option ||= default_sort_option_value
+    selected_option = if values['order'].present? && sort.any? { |option| option['name'].parameterize == values['order'] }
+                        values['order']
+                      else
+                        default_sort_option_value
+                      end
 
     options_for_select(options, selected: selected_option, disabled: disabled_option)
   end
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -134,15 +134,12 @@ class FinderPresenter
 
     options = Hash[sort.collect { |option| [option['name'], option['name'].parameterize] }]
 
-    disabled_option = keywords.blank? ? relevance_sort_option : ''
+    disabled_option = keywords.blank? ? relevance_sort_option_value : ''
 
-    selected_option = if values['order']
-                        sort.detect { |option| option['name'].parameterize == values['order'] }
-                      end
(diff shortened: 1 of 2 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -136,10 +136,7 @@ def sort_options
 
     disabled_option = keywords.blank? ? relevance_sort_option : ''
 
-    selected_option = if values['order']
-                        sort.detect { |option| option['name'].parameterize == values['order'] }
-                      end
-
+    selected_option = values['order'] unless values['order'].nil? && sort.detect { |option| option['name'].parameterize == values['order'] }.nil?
```
andrewgarner: We could structure the code like so to reduce the line length and improve readability?

```ruby
def sort_values
  sort
    .map { |option| option['name'].parameterize }
    .compact
end

selected_option = values['order'] if sort_values.include?(values['order'])
selected_option ||= default_sort_option_value
```

koetsier: can we add a test for this?

karlbaker02: @andrewgarner Your suggestion definitely improves on line length and readability, at the expense of performing the computation of `sort_values` regardless of whether of `values['order']` is present. We can definitely improve on the readability though.

@koetsier I'm sure we could

karlbaker02: Tests added @koetsier

spec/presenters/finder_presenter_spec.rb
```diff
@@ -68,4 +80,77 @@
...
+    end
+
+    it "returns sort options when sort is present" do
+      expected_options = "<option value=\"most-viewed\">Most viewed</option>\n<option value=\"updated-newest\">Updated (newest)</option>"
+
+      expect(presenter_with_sort.sort_options).to eql(expected_options)
+    end
+
+    it "returns sort options without relevance when keywords is not present" do
+      expected_options = "<option value=\"most-viewed\">Most viewed</option>\n<option value=\"updated-newest\">Updated (newest)</option>"
```
koetsier: this test is identical to the one starting line 89

app/presenters/finder_presenter.rb
koetsier: I'm not sure this is right:

```selected_option = values['order'] unless values['order'].nil? && sort.detect { |option| option['name'].parameterize == values['order'] }.nil?```

is the same as

```selected_option = values['order'] if !values['order'].nil? || !sort.detect { |option| option['name'].parameterize == values['order'] }.nil?```

so this implies that in the second part ```values['order']``` can be set to nil and so the second part could be simplified to ```sort.any?{| |option| option['name'].nil?}

File: spec/presenters/finder_presenter_spec.rb (hunk 2 above)
Comment (karlbaker02) [reply spec/presenters/finder_presenter_spec.rb]: It's close, but not the same. The test on line 85 uses `presenter`, which has `nil` sort options. This test on line 95 uses `presenter_with_sort`, which has sort options (including relevance).

File: app/presenters/finder_presenter.rb (hunk 1 above)
Comment (karlbaker02) [reply app/presenters/finder_presenter.rb]: You're right, not the same.

This should actually be `unless ... || ...`. Will update.

File: app/presenters/finder_presenter.rb [hunk 3]
```diff
@@ -134,15 +134,12 @@ def sort_options
...
 
-    disabled_option = keywords.blank? ? relevance_sort_option : ''
-
-    selected_option = if values['order']
-                        sort.detect { |option| option['name'].parameterize == values['order'] }
-                      end
+    disabled_option = keywords.blank? ? relevance_sort_option_value : ''
 
+    selected_option = values['order'] unless values['order'].nil? || sort.detect { |option| option['name'].parameterize == values['order'] }.nil?
     selected_option ||= default_sort_option_value
```
koetsier: what about:

```
    selected_option = if values['order'].present? && sort.any? { |option| option['name'].parameterize == values['order'] }
                        values['order']
                      else
                        default_sort_option_value
                      end
```

karlbaker02: Updated @koetsier

PR #780, opened 2019-01-15
Revert "Adds accessible autocomplete facet to finders"

bilbof:
Reverts alphagov/finder-frontend#740

This should be reverted as removing filters does not work with autocomplete items.

Steps to reproduce:

1. Go to this link: https://www-origin.integration.publishing.service.gov.uk/aaib-reports?keywords=&aircraft_category=commercial-rotorcraft&report_type=&date_of_occurrence%5Bfrom%5D=&date_of_occurrence%5Bto%5D=
2. Click the 'X' next to the filter 'commercial - aircraft'

### DIFF
--- a/app/assets/stylesheets/components/_date-filter.scss
+++ b/app/assets/stylesheets/components/_date-filter.scss
@@ -5,3 +5,8 @@
     margin-bottom: $gutter;
   }
 }
+
+.app-c-date-filter__help-text {
+  @include core-16;
+  display: block;
+}
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -234,7 +234,7 @@
       margin: 0 0 $gutter-half 0;
 
       .result-count {
-        @include bold-27;
+        @include bold-48;
         margin-right: $gutter-one-third / 2;
       }
     }
--- a/app/models/select_facet.rb
+++ b/app/models/select_facet.rb
@@ -3,15 +3,24 @@ class SelectFacet < FilterableFacet
     facet['allowed_values']
   end
 
-  def options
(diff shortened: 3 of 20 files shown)


leenagupte:
@bilbof The remove is still doing odd things. If you select two things, but only remove one, both facets are removed. The results displayed are still filtered by the existing filter. 

Go here and try removing one of the filters: https://finder-frontend-pr-780.herokuapp.com/cma-cases?keywords=&case_type%5B%5D=review-of-orders-and-undertakings&market_sector%5B%5D=aerospace&closed_date%5Bfrom%5D=&closed_date%5Bto%5D=

bilbof:
@leenagupte I figured out why we're seeing that other issue.

It looks like there are two issues:

1. We can't remove filters after the autocomplete PR is merged
2. We can only remove all filters when removing a checkbox type filter

The first issue can be resolved by merging this PR, which reverts the autocomplete PR. I am not 100% sure of the cause, but I think it is related to the way autocomplete works (has multiple inputs/selects for a single input field, but only one of them has a name etc). I need to look into that more.

The second issue is caused by this global JS: 

https://github.com/alphagov/govuk-frontend/blob/e41a9b39ccc8fff13fe64a09f868ce64a7e4d502/src/components/checkboxes/checkboxes.js#L44

It's certainly unintended behaviour.

This is the way that code works:

1. User adds filters via a checkbox
2. We display a tag element for that checkbox
3. User clicks on the tag element, to 'unclick' the checkbox.
4. This code listens for any click on the form, and is informed there has been a click. It then adds `govuk-checkboxes__conditional--hidden` to the results block, hiding it.

I have fixed this issue here: https://github.com/alphagov/finder-frontend/pull/782

So we need to merge this and fix the issue you found with [this other pr](https://github.com/alphagov/finder-frontend/pull/782)

PR #787, opened 2019-01-17
Finder for Research and Statistics supergroup

bilbof:
This adds a research and statistics finder representing the research_and_statistics supergroup [1].

This JSON has not been published as a content item yet, so are only available for development for now and will be published later (and removed from here) like the news and communications JSON

Trello link [3]

[1] https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml#L568
[2] https://www.gov.uk/government/statistics
[3] https://trello.com/c/h026plmk/201-create-finder-for-research-and-statistics-supergroup

### DIFF
--- a/app/lib/filters/dropdown_select_filter.rb
+++ b/app/lib/filters/dropdown_select_filter.rb
@@ -1,0 +1,17 @@
+module Filters
+  class DropdownSelectFilter < Filter
+    def value
+      Array(parsed_value)
+    end
+
+  private
+
+    def parsed_value
+      return unless params.present?
+
+      JSON.parse params
+    rescue JSON::ParserError
+      params
+    end
+  end
+end
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -27,35 +27,6 @@ private
     end
   end
 
-  def development_env_finder_json
-    return all_content_json if is_all_content?
-    return news_and_communications_json if is_news_and_communications?
-    return services_json if is_services?
-    return transparency_and_freedom_of_information_releases_json if is_transparency_and_freedom_of_information_releases?
-
-    ENV["DEVELOPMENT_FINDER_JSON"]
-  end
-
-  def all_content_json
-    # Hard coding this in during development
-    "features/fixtures/all_content.json"
-  end
-
-  def news_and_communications_json
-    # Hard coding this in during development
-    "features/fixtures/news_and_communications.json"
-  end
-
-  def services_json
-    # Hard coding this in during development
-    "features/fixtures/services.json"
-  end
-
-  def transparency_and_freedom_of_information_releases_json
-    # Hard coding this in during development
-    "features/fixtures/transparency_and_freedom_of_information_releases.json"
-  end
-
   def merge_and_deduplicate(search_response)
     results = search_response.fetch("results")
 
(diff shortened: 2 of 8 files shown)


sihugh:
@bilbof there are some conflicts at the mo 😢

bilbof:
Thanks @sihugh, they should be fixed now. It's this stuff in https://github.com/alphagov/finder-frontend/pull/787/files#diff-942dd67f8c579db2a0fc47e90ad8fcecR134 that is supposed to make in-development finders easier to add.

andrewgarner:
Can we remove the following too?

https://github.com/alphagov/finder-frontend/blob/1a05456bc4e3d5843c3422e29887e68ac1c65d7f/app.json#L32-L43

features/fixtures/statistics.json
```diff
@@ -0,0 +1,116 @@
...
+  "publishing_request_id":"",
+  "links":{},
+  "description":"Find research and statistics from government",
+  "details":{
+    "document_noun":"statistic",
+    "filter":{
+    },
+    "format_name":"statistic",
+    "show_summaries":true,
+    "sort": [
```
karlbaker02: I don't think this is correct. From the card:

```
updated, popularity, relevance
(AND release date ordering if user selects 'Upcoming' in Statistics facet)
```

bilbof: Sorry the spec changed and I didn't update the PR. This iteration has been approved by Si+Mia, and we'll be making some further changes later on.

bilbof:
@andrewgarner I removed those env vars, please could you take another look at this? :)

app/lib/filters/dropdown_select_filter.rb
```diff
@@ -0,0 +1,17 @@
+module Filters
+  class DropdownSelectFilter < Filter
```
andrewgarner: A very minor point but perhaps we should mirror the HTML naming convention and call this `SelectFilter`. `Dropdown` alludes to specific behaviour if the element has a size of 1.

app/lib/finder_api.rb
```diff
@@ -156,20 +127,27 @@ def query_builder_class
     SearchQueryBuilder
   end
 
-  def is_all_content?
-    base_path == "/all-content"
-  end
+  FINDERS_IN_DEVELOPMENT = {
+    "/news-and-communications" => "news_and_communications",
+    "/research-and-statistics" => "statistics",
```
andrewgarner: Was is a conscious decision to drop `research` from the content item file name? By doing so I wonder if we're risk loosing important context?

andrewgarner: So I've looked at this some more and can see that in the content item we're not including the document types for the research supertype.

https://github.com/alphagov/govuk_document_types/blob/2645f543bea870b9469a70d6c93fde4db6e493dd/data/supertypes.yml#L468-L472

With that being the case we can't really include `research` in the URL. Also how does this effect what displays on the organisation pages?

bilbof: I updated this and the content item so that this is explicitly a stats finder. There's another card to iterate on the behaviour (https://trello.com/c/40IA4pPe/241-design-research-statistics-finder) before we ship this.

bilbof:
@andrewgarner I have pushed a change removing mentions of 'research' in the finder.

PR #793, opened 2019-01-17
Make result count the specific count, rather than 1,000

bilbof:
https://trello.com/c/ES5v3WIb/234-make-result-count-specific


bilbof:
Waiting for design approval before merging.

bilbof:
This is coming in a big PR that Andy is working on.

PR #796, opened 2019-01-18
[DO NOT MERGE] Fix facet tags for accessible-autocomplete

karlbaker02:
This PR fixes the facet tags for the accessible-autocomplete component so that clicking on the remove button has the desired behaviour of removing this filter from the search.

In order to achieve this, it is necessary to determine if the input we're dealing with has created by the accessible-autocomplete component. The `accessible-autocomplete` does not set the `name` attribute of the generated `<input>` element, so this is not helpful to us.

Instead, an `accessible-autocomplete` will enhance a `select` element and rename that element to be `<ID>__listbox`, where `<ID>` is the original `id` attribute on the `select` element. Therefore we can look for the presence of a `select` with an id == `<FACET_VALUE>__listbox` to determine if we're dealing with an autocomplete. If this is the case, we can get the input without needing to `find` anything further inside the element, giving us the appropriate element type and input type, before entirely replacing the current value of the autocomplete with a blank string, removing the filter from the search.

PR: https://finder-frontend-pr-796.herokuapp.com/news-and-communications

Solo: @karlbaker02

Trello card: https://trello.com/c/Gk9FE0Sg/238-fix-finders-bugs


karlbaker02:
As discussed with @andysellick, this may not be the best place for this commit. Given that the accessible-autocomplete in this repo was previously merged and then reverted, we could either revert the revert and merge this commit in separately, or open a new branch / use this branch with the autocomplete changes and deploy the fix in the same release.

andysellick:
Superseded by https://github.com/alphagov/finder-frontend/pull/819

PR #799, opened 2019-01-18
Frontend changes to finders

andysellick:
Makes a few frontend adjustments to the finders, specifically:

- reduces size of results number (was too big, matched H1)
- make results number show actual number of results, not "1,000+"
- reorder the facets to improve ease of use
- hide the 'sorted by' facet display from sighted users but not screen readers (sighted users can see what they've selected, screen reader users have the visuallyhidden text announced to them when they change the sort option)
- update the date facets to use components, improving the position of the hint text and the spacing in general

Trello card: https://trello.com/c/oKo6gvZw/190-make-frontend-changes-to-news-and-comms-finder-following-review

Example urls:

- https://finder-frontend-pr-799.herokuapp.com/administrative-appeals-tribunal-decisions
- https://finder-frontend-pr-799.herokuapp.com/news-and-communications

Some of these changes apply to the news and communications finder, which isn't live yet, so the screenshots below aren't brilliant examples, sorry.

Before:

screen shot 2019-01-18 at 14 01 45

After:

screen shot 2019-01-18 at 14 02 04

### DIFF
--- a/app/presenters/result_set_presenter.rb
+++ b/app/presenters/result_set_presenter.rb
@@ -80,7 +80,7 @@ class ResultSetPresenter
 
     sort_option ||= finder.default_sort_option
 
-    "sorted by <strong>" + sort_option['name'] + "</strong>" if sort_option.present?
+    "<span class='visually-hidden'>sorted by <strong>" + sort_option['name'] + "</strong></span>" if sort_option.present?
   end
 
 private
--- a/app/presenters/result_set_presenter.rb
+++ b/app/presenters/result_set_presenter.rb
@@ -1,5 +1,6 @@
 class ResultSetPresenter
   include ERB::Util
+  include ActionView::Helpers::NumberHelper
 
   attr_reader :finder, :document_noun, :results, :total
 
@@ -19,7 +20,7 @@ class ResultSetPresenter
 
   def to_hash
     {
-      total: total > 1000 ? "1,000+" : total,
+      total: number_with_delimiter(total),
       generic_description: generic_description,
       pluralised_document_noun: document_noun.pluralize(total),
       applied_filters: selected_filter_descriptions,
--- a/app/assets/stylesheets/components/_date-filter.scss
+++ b/app/assets/stylesheets/components/_date-filter.scss
@@ -5,8 +5,3 @@
     margin-bottom: $gutter;
   }
 }
-
-.app-c-date-filter__help-text {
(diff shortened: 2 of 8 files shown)


app/presenters/result_set_presenter.rb
```diff
@@ -19,7 +20,7 @@ def initialize(finder, filter_params, view_context)
 
   def to_hash
     {
-      total: total > 1000 ? "1,000+" : total,
+      total: number_with_delimiter(total, delimiter: ','),
```
vanitabarrett: The default is `,` so not sure we need to explicitly pass in delimiter here. https://apidock.com/rails/ActionView/Helpers/NumberHelper/number_with_delimiter

app/presenters/result_set_presenter.rb
```diff
@@ -80,7 +80,7 @@ def sort_options
 
     sort_option ||= finder.default_sort_option
 
-    "sorted by <strong>" + sort_option['name'] + "</strong>" if sort_option.present?
+    "<span class='visually-hidden'>, sorted by <strong>" + sort_option['name'] + "</strong></span>" if sort_option.present?
```
vanitabarrett: Is there a reason why this starts with a comma?

app/presenters/result_set_presenter.rb
andysellick: If I change it to just `total: total` it doesn't have a comma.

app/presenters/result_set_presenter.rb
andysellick: Probably only because it had one before. I'll check it with a screenreader to see if it provides a useful pause or not, but otherwise yeah, could be removed, thanks.

app/presenters/result_set_presenter.rb
vanitabarrett: @andysellick Sorry, that wasn't clear. I meant the default for `number_with_delimiter` is a comma, so not sure `delimiter` needs to be explicitly defined.

` total: number_with_delimiter(total)`

andysellick: Ah yes, right you are. Have changed, thanks.

app/presenters/result_set_presenter.rb
andysellick: I can't decide so I've taken it out. Better not to second guess these things.

PR #801, opened 2019-01-21
[wip] Use subscription component for email link

sihugh:
This also adds tracking for the link, and some tests for the email link behaviour.

I've removed the email link from the advanced search as this isn't used.

https://finder-frontend-pr-801.herokuapp.com/cma-cases (for example)

https://trello.com/c/l6kGfBhh/230-add-missing-facet-tag-and-email-alert-tracking

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -14,8 +14,8 @@
     this.$countBlock = options.$results.find('#js-search-results-info');
     this.action = this.$form.attr('action') + '.json';
     this.$atomAutodiscoveryLink = options.$atomAutodiscoveryLink;
-    this.$emailLink = $("p.email-link a");
 
+    this.$emailLink = $(".gem-c-subscription-links a");
     this.emailSignupHref = this.$emailLink.attr('href');
 
     this.$orderSelect = this.$form.find('.js-order-results');
@@ -78,6 +78,9 @@
 
     this.$emailLink.attr(
       'href', this.emailSignupHref + "?" + $.param(this.state)
+    );
+    this.$emailLink.attr(
+      'data-track-action', this.emailSignupHref + "?" + $.param(this.state)
     );
   };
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -79,6 +79,9 @@
     this.$emailLink.attr(
       'href', this.emailSignupHref + "?" + $.param(this.state)
     );
+    this.$emailLink.attr(
+      'data-track-action', this.emailSignupHref + "?" + $.param(this.state)
```
andrewgarner: I assume we don't want this to diverge from `this.$emailLink.attr('href')`? If so maybe we we could extract a variable or reference the `$emailLink` href attribute?

features/finders.feature
```diff
@@ -7,6 +7,10 @@ Feature: Filtering documents
     Given a collection of documents exist
     Then I can get a list of all documents with matching metadata
 
+  Scenario: Email alert signups
+    Given a finder exists with an email signup link
+    And I can see an email alert signup
```
andrewgarner:
```suggestion
    Then I can see an email alert signup
```

features/step_definitions/filtering_steps.rb
```diff
@@ -331,3 +336,9 @@
 
   expect(page).to have_content('sorted by A-Z')
 end
+
+And(/^I can see an email alert signup$/) do
+  visit finder_path('mosw-reports')
+  expect(page).to have_link('Subscribe to email alerts', href: 'https://www.gov.uk/mosw-reports/email-signup')
+
```
andrewgarner:
```suggestion
```

app/assets/javascripts/live_search.js
sihugh: I think this URL construction is fundamentally broken as it omits the case where there is an existing query string.  I think I'm going to have to do a bit more on it.  Thanks for the sanity check

features/finders.feature
sihugh: Thank you!  I am becoming robot

app/assets/javascripts/live_search.js
bilbof: FYI, this is fixed in https://github.com/alphagov/finder-frontend/pull/832.

PR #807, opened 2019-01-22
Policy papers and consultations finder

bilbof:
This finder will enable users to find things within the policy_and_engagement supergroup, using relevant facets. This won't be visible to end users.

A follow-up PR will prevent default dropdown select options from being displayed by the facet tag presenter.

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -133,6 +133,7 @@ private
     "/services" => "services",
     "/transparency-and-freedom-of-information-releases" => "transparency_and_freedom_of_information_releases",
     "/all-content" => "all_content",
+    "/policy-papers-and-consultations" => 'policy_and_engagement',
   }.freeze
 
   def development_env_finder_json
--- a/features/fixtures/policy_and_engagement.json
+++ b/features/fixtures/policy_and_engagement.json
@@ -1,0 +1,114 @@
+{
+  "analytics_identifier":null,
+  "base_path":"/policy-papers-and-consultations",
+  "content_id":"622e9691-4b4f-4e9c-bce1-098b0c4f5ee2",
+  "content_purpose_document_supertype":"navigation",
+  "document_type":"finder",
+  "email_document_supertype":"other",
+  "first_published_at":"2018-11-07T14:47:48.000+00:00",
+  "government_document_supertype":"other",
+  "locale":"en",
+  "navigation_document_supertype":"other",
+  "phase":"alpha",
+  "public_updated_at":"2018-11-07T14:47:48.000+00:00",
+  "publishing_app":"finder-frontend",
+  "rendering_app":"finder-frontend",
+  "schema_name":"finder",
+  "search_user_need_document_supertype":"government",
+  "title":"Policy papers and consultations",
+  "updated_at":"2018-11-07T14:47:48.000+00:00",
+  "user_journey_document_supertype":"finding",
+  "withdrawn_notice":{
+
+  },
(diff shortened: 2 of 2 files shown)


bilbof:
@andrewgarner good spot, I've corrected this so policy papers should now be included.

features/fixtures/policy_and_engagement.json
```diff
@@ -0,0 +1,114 @@
...
+        "key": "content_store_document_type",
+        "name": "Consultation status",
+        "preposition": "with consultation status",
+        "type":"dropdown_select",
+        "display_as_result_metadata": true,
+        "filterable": true,
+        "allowed_values": [
+          {
+            "text": "All consultations",
+            "value": "[\"open_consultation\",\"closed_consultation\",\"consultation_outcome\",\"impact_assessment\",\"case_study\",\"policy_paper\"]",
```
andrewgarner: Our rules now differ from `govuk_document_types`. We're labelling an impact assessment, policy paper and case study as a consultation. I'm not sure that is correct?

See https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml#L458-L467

```yaml
    - id: policy
      document_types:
        - impact_assessment
        - case_study
        - policy_paper
    - id: consultations
      document_types:
        - open_consultation
        - closed_consultation
        - consultation_outcome
```

bilbof: I followed the supergroup for `policy_and_engagement` which includes these document types:

```
    - id: policy_and_engagement
      document_types:
        - impact_assessment
        - case_study
        - policy_paper
        - open_consultation
        - closed_consultation
        - consultation_outcome
```

https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml#L559-L566

The default label for the consultations facet is 'all consultations', which I intended to imply 'everything tagged to the policy and engagement supergroup'. Would it be worth us asking a content specialist about this?

bilbof: I need a follow up PR to hide the facet tag when the default option is selected for this facet, currently this only happens if the value is `""`.

bilbof: Here's that Pr which kind of solves the problem: https://github.com/alphagov/finder-frontend/pull/809.

So we can remove "All consultations" as being a group of all document_types included in the supergroup and have something similar to the research and stats finder.

app/lib/finder_api.rb
```diff
@@ -133,6 +133,7 @@ def query_builder_class
     "/services" => "services",
     "/transparency-and-freedom-of-information-releases" => "transparency_and_freedom_of_information_releases",
     "/all-content" => "all_content",
+    "/policy-and-engagement" => 'policy_and_engagement',
```
andrewgarner: We should adopt the latest user language, see https://www.gov.uk/search/advanced?group=policy_and_engagement&topic=%2Ftransport

```suggestion
    "/policy-papers-and-consultations" => 'policy-papers-and-consultations',
```

features/fixtures/policy_and_engagement.json
```diff
@@ -0,0 +1,114 @@
+{
+  "analytics_identifier":null,
+  "base_path":"/policy-and-engagement",
```
andrewgarner:
```suggestion
  "base_path":"/policy-papers-and-consultations",
```

sihugh:
Merging for now.  We have a follow up card to improve this: https://trello.com/c/40IA4pPe/241-design-research-statistics-finder

PR #809, opened 2019-01-22
Allow filtering on document_type and supergroup

bilbof:
This will allow users to filter by document type and supergroup.

Supergroup filters are to be deprecated. Internally, finder-frontend
has a class TranslateContentPurposeFields that transforms supergroup
filters (content_purpose_supergroup) to document type filters
(content_store_document_type).

The problem is that, if you have a document_type facet/filter, then
this can only expand the search. For example, if you are viewing the
news and communications finder, this will be filtered by
the news_and_communications supergroup (this gets turned into a
document_type filter of document_type[]=news_story,article,... etc).
If you wanted to just view documents of type press release on that
finder, this wouldnt be possible.

This code enables one to filter by supergroup OR by a provided set
of document types. The supergroup will be the default unless the
document types are provided. The expected use case is filtering
down supergroups by subgroupings of document types.

An example of this is included in this commit, where I make
the research_and_statistics finder filterable by document type
sets of Published/Upcoming/All statuses of stats.

### DIFF
--- a/app/lib/translate_content_purpose_fields.rb
+++ b/app/lib/translate_content_purpose_fields.rb
@@ -15,7 +15,7 @@ private
 
   def translate_fields_with_prefix(prefix)
     document_types = Array(content_store_document_types(prefix))
-    document_types += Array(subgroup_document_types(prefix) || supergroup_document_types(prefix))
+    document_types += Array(subgroup_document_types(prefix) || supergroup_document_types(prefix)) unless document_types.any?
 
     @query.delete("#{prefix}_content_purpose_subgroup")
     @query.delete("#{prefix}_content_purpose_supergroup")
--- a/features/fixtures/statistics.json
+++ b/features/fixtures/statistics.json
@@ -25,6 +25,7 @@
   "details":{
     "document_noun":"statistic",
     "filter":{
+      "content_purpose_supergroup": "research_and_statistics"
     },
     "format_name":"statistic",
     "show_summaries":true,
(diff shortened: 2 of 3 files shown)


features/fixtures/statistics.json
```diff
@@ -25,6 +25,7 @@
   "details":{
     "document_noun":"statistic",
     "filter":{
+      "content_purpose_supergroup": "research_and_statistics"
```
bilbof: Go to https://finder-frontend-pr-809.herokuapp.com/statistics to view this finder following the changes.

app/lib/translate_content_purpose_fields.rb
```diff
@@ -15,7 +15,7 @@ def call
 
   def translate_fields_with_prefix(prefix)
     document_types = Array(content_store_document_types(prefix))
-    document_types += Array(subgroup_document_types(prefix) || supergroup_document_types(prefix))
+    document_types += Array(subgroup_document_types(prefix) || supergroup_document_types(prefix)) unless document_types.any?
```
sihugh: This might break some existing things.  I've raised this to fix one that I know about: https://github.com/alphagov/rummager/pull/1408

sihugh: The prepare-eu-exit ones definitely do.  The advanced-search ones (AKA the taxonomy topic finders) also do.  Apart from those, I'm pretty sure it's just our new ones.

bilbof: It looks like the advanced search finder doesn't use document type filtering (no document type filter in the content item) so I think we're in the clear there.

https://www.gov.uk/api/content/search/advanced

Did https://github.com/alphagov/rummager/pull/1408 solve this for the prepare-eu-exit ones? I saw you deployed this to prod earlier. 😄

PR #813, opened 2019-01-23
Add guidance and regulation supergroup finder

koetsier:
trello: https://trello.com/c/J5XuATwO/198-create-finder-for-guidance-supergroup

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -129,6 +129,7 @@ private
 
   FINDERS_IN_DEVELOPMENT = {
     "/news-and-communications" => "news_and_communications",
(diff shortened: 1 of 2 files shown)


app/lib/finder_api.rb
```diff
@@ -129,6 +129,7 @@ def query_builder_class
 
   FINDERS_IN_DEVELOPMENT = {
     "/news-and-communications" => "news_and_communications",
+    "/guidance_and_regulation" => "guidance_and_regulation",
```
bilbof: The other supergroup finders don't have underscores in the URL.

koetsier: Done

PR #821, opened 2019-01-24
Add tracking to the business finder sort dropdown.

peteglondon:
We want to track the selections the user makes.

Now we get a GA event when user selects an option from
the `Sort by` dropdown.

Trello: https://trello.com/c/JgohmaDW/3-add-sort-order-to-the-business-finder-results-page

### DIFF
--- a/app/views/finders/show.html.erb
+++ b/app/views/finders/show.html.erb
@@ -75,7 +75,8 @@
                          class: 'govuk-select js-order-results',
                          'aria-controls': 'js-search-results-info',
                          data: { 'default-sort-option' => finder.default_sort_option_value,
-                                 'relevance-sort-option' => finder.relevance_sort_option_value } %>
+                                 'relevance-sort-option' => finder.relevance_sort_option_value,
+                                 'module' => 'track-select-change'} %>
         </div>
       <% end %>
 
--- a/spec/presenters/finder_presenter_spec.rb
+++ b/spec/presenters/finder_presenter_spec.rb
@@ -82,18 +82,31 @@ RSpec.describe FinderPresenter do
   end
 
   describe "#sort_options" do
+    def sort_option(label, value, disabled: false, selected: false)
+      disabled_attr = disabled ? 'disabled="disabled" ' : ''
+      selected_attr = selected ? 'selected="selected" ' : ''
+      "<option data-track-category=\"dropDownClicked\" data-track-action=\"clicked\" data-track-label=\"#{label}\" #{disabled_attr}#{selected_attr}value=\"#{value}\">#{label}</option>"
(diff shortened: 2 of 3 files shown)


peteglondon:
[Add sort order to the business finder results page](https://trello.com/c/JgohmaDW/3-add-sort-order-to-the-business-finder-results-page)

vanitabarrett:
It would be a good idea to move the relevant JavaScript we're making use of here (which currently lives in [select.js](https://github.com/alphagov/govuk_publishing_components/tree/master/app/assets/javascripts/govuk_publishing_components/components), implying it's only ever used by the select component) into [lib/](https://github.com/alphagov/govuk_publishing_components/tree/master/app/assets/javascripts/govuk_publishing_components/lib) so it's clear it's being used in other places.

vanitabarrett:
cc @sihugh just a heads up that this will add tracking to the sort button on other finders too.

sihugh:
@vanitabarrett I was keeping an eye out for this :-) . Haur will be stoked 👍

spec/presenters/finder_presenter_spec.rb
```diff
@@ -82,50 +82,85 @@
   end
 
   describe "#sort_options" do
+    def sort_option(label, value, disabled: false, selected: false)
+      disabled_attr = if disabled
```
DilwoarH: This may be cleaner:
```
disabled_attr = disabled ? 'disabled="disabled" ' : ''
```

spec/presenters/finder_presenter_spec.rb
```diff
@@ -82,50 +82,85 @@
...
+                        'disabled="disabled" '
+                      else
+                        ''
+                      end
+      selected_attr = if selected
```
DilwoarH:
```
selected_attr = selected ? 'selected ="selected" ' : ''
```

spec/presenters/finder_presenter_spec.rb
peteglondon: I think the linter will not like that, but maybe it's configured differently here. I'll give it a go.

app/views/finders/show.html.erb
```diff
@@ -75,7 +75,8 @@
                          class: 'govuk-select js-order-results',
                          'aria-controls': 'js-search-results-info',
                          data: { 'default-sort-option' => finder.default_sort_option_value,
-                                 'relevance-sort-option' => finder.relevance_sort_option_value } %>
+                                 'relevance-sort-option' => finder.relevance_sort_option_value ,
```
DilwoarH: extra space after variable but other than that everything look good

👍 💯

PR #823, opened 2019-01-24
[WIP][DO NOT MERGE] WIP Feedback Form

vanitabarrett:
Trello: https://trello.com/c/a9WXTd6f/64-build-a-feedback-component-that-sits-at-the-bottom-of-finder-results

This starts to build the feedback component that will appear under results on the business finder. The frontend (visual) part of the work is done. Still left to do:
- Decide where we submit the "Yes" votes to
- Implement the above
- Add JS to submit the form to Feedex OR somewhere else in the short term. If submitting to feedex, we need to make sure we can differentiate between data from here and data from the global feedback form across GOV.UK.

We are not considering the following as part of this work. There are other stories for this:

- JS disabled
- Tracking
- Making it a prod-ready component

### DIFF
--- a/app/assets/javascripts/modules/feedback.js
+++ b/app/assets/javascripts/modules/feedback.js
@@ -1,0 +1,69 @@
+window.GOVUK = window.GOVUK || {};
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (GOVUK) {
+  "use strict";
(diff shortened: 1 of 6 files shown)


vanitabarrett:
Superseded by https://github.com/alphagov/finder-frontend/pull/836

PR #834, opened 2019-01-29
Tracking not working when updated results in form change

MahmudH:
- When a change is made in the form, js-live-search-results-block, tracking stops working.
- When a form change is triggered, the JavaScript modules need reinitializing to be able to
add tracking again.

Trello: https://trello.com/c/q1W29H42

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -102,11 +102,16 @@
         function(){
           var newPath = window.location.pathname + "?" + $.param(this.state);
           history.pushState(this.state, '', newPath);
+          this.trackingInit();
         }.bind(this)
       )
     }
     this.trackPageView();
   };
+
+  LiveSearch.prototype.trackingInit = function() {
+    GOVUK.modules.start($('.js-live-search-results-block'));
+  }
 
   LiveSearch.prototype.trackPageView = function trackPageView() {
     if (this.canTrackPageview()) {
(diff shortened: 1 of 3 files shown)


features/step_definitions/filtering_steps.rb
```diff
@@ -331,6 +332,7 @@
     expect(page).to have_content('Register a magical spell')
   end
 
+  expect(page).to have_css("a[data-track-category='navFinderLinkClicked']")
```
DilwoarH: does the test check to see if the analytics work after live search?

app/assets/javascripts/live_search.js
```diff
@@ -101,13 +101,18 @@
       pageUpdated.done(
         function(){
           var newPath = window.location.pathname + "?" + $.param(this.state);
+          this.trackingInit();
```
DilwoarH: it makes more sense to have it in this order

```
var newPath = window.location.pathname + "?" + $.param(this.state);
history.pushState(this.state, '', newPath);
this.trackingInit();
```

features/step_definitions/filtering_steps.rb
MahmudH: Any changes to the form, removed the tracking, so the test checks if the tracking is still available on form change.

app/assets/javascripts/live_search.js
MahmudH: 👍

features/step_definitions/filtering_steps.rb
MahmudH: I have added a Jasmin test for this

PR #836, opened 2019-01-29
[DO NOT MERGE] Add new feedback form to business finder

vanitabarrett:
Trello: https://trello.com/c/a9WXTd6f/64-build-a-feedback-component-that-sits-at-the-bottom-of-finder-results

Left to do before this is production-ready:
- Confirm tracking events are correct/what we want
- Consider users with JS disabled
- Write tests
- Think about making it an actual component in finder-frontend


vanitabarrett:
Closing this for now as it didn't test well in user research. Leaving the branch in case we revisit this in the future.

PR #837, opened 2019-01-30
Adjust checkbox facet properties

sihugh:
The checkbox facet adds a few new schema properties which we can avoid if we reuse some of the existing ones a bit more judiciously.

I've removed `label`, `checkbox_label` and `value` from the required properties.  

I've defaulted `value` to true (and we can add that property if we really need it).  
`checkbox_name` seems to be used how other facets use `name`.
`short_name` and `label` seem synonymous.

Check it at: https://finder-frontend-pr-837.herokuapp.com/news-and-communications

https://trello.com/c/oE1sdBdt/278-add-ability-to-publish-news-and-communications-finder-from-rummager


sihugh:
@bilbof hope this is OK with the email work

PR #843, opened 2019-01-31
Allow Atom feeds on more things

sihugh:
When the default order is set to `-public_timestamp`, this matches the behaviour of an atom feed so we should allow this.

I've also added some tests around this behaviour so that we can be more confident about changing it in future.  I'm not sure it's quite right to expect the atom feed to behave exactly like search results in a finder.


bilbof:
@sihugh sadly govuk-lint-ruby has found an unused variable :(

```
[build]         presenter = described_class.new(content_item(sort_options_with_default), values)
[build]         ^^^^^^^^^
```

PR #846, opened 2019-02-01
Looping and reloading question pages

MahmudH:
- Timeouts are taking the users to the first page because of caching.
- This was resolved by adding a caching limit and this limit is not long enough, therefore this issue
still persists, therefore the caching limit has been increased to 15.
- Migrating the application to AWS will solve this issue without the need
for caching.

Trello: https://trello.com/c/cg8UDuJU

### DIFF
--- a/app/controllers/qa_controller.rb
+++ b/app/controllers/qa_controller.rb
@@ -23,7 +23,7 @@ private
   def raw_finder
     # FIXME: stop caching this once the app has migrated to AWS
     @raw_finder ||= begin
-      item_hash = Rails.cache.fetch("QaController/#{slug}", expires_in: 5.minutes) do
+      item_hash = Rails.cache.fetch("QaController/#{slug}", expires_in: 15.minutes) do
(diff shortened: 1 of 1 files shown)


MuriloDalRi:
Looks like this has changed on master

steventux:
@MahmudH Is there a reason why 15 minutes was chosen? Perhaps we should err on the side of doubling the cache time to 10 mins and see what the effect is.
It looks like the controller has some changes so this needs PR a rebase.

sihugh:
We now cache all content store requests from finder frontend for 5 minutes.  There's only been one 5xx error today on the qa controller (and four yesterday).  I'm not sure you'll get much better than that with a 15 minute cache.  Could it be worth trying to redirect people directly to the finder if the page number in the URL doesn't match the rendered page (as that would indicate you're on the mirror)?

vanitabarrett:
Closing this as we don't think increasing from 5 - 15 minutes will improve things

PR #848, opened 2019-02-01
Memoize cache response for registries

sihugh:
We're currently hitting the cache for each request here.  It's better in memory once we've got it.

snippet of cache reads for the world locations registry

### DIFF
--- a/app/lib/registries/topic_taxonomy_registry.rb
+++ b/app/lib/registries/topic_taxonomy_registry.rb
@@ -7,7 +7,7 @@ module Registries
     end
 
     def taxonomy_tree
-      Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
+      @taxonomy_tree ||= Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
         taxonomy_tree_as_hash
       end
     rescue GdsApi::HTTPServerError
--- a/app/lib/registries/world_locations_registry.rb
+++ b/app/lib/registries/world_locations_registry.rb
@@ -14,7 +14,7 @@ module Registries
   private
 
     def cached_locations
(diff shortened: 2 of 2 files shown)


app/lib/registries/topic_taxonomy_registry.rb
```diff
@@ -7,7 +7,7 @@ def [](content_id)
     end
 
     def taxonomy_tree
-      Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
+      @taxonomy_tree = Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
         taxonomy_tree_as_hash
       end
     rescue GdsApi::HTTPServerError
```
bilbof: Will the memoization variable be set as the default `{}` if there's an error fetching the registry data? If so, do we want to retry fetching the data if that happens?

app/lib/registries/topic_taxonomy_registry.rb
```diff
@@ -7,7 +7,7 @@ def [](content_id)
     end
 
     def taxonomy_tree
-      Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
+      @taxonomy_tree = Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
```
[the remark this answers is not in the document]
sihugh: That was my total intention 🤦‍♂️

app/lib/registries/topic_taxonomy_registry.rb
sihugh: The default `{}` is outside the memoization block so it should be fine and will be retried

bilbof: Awesome :)

PR #849, opened 2019-02-04
Adds different category of GA event for deselect of checkbox facet

oscarwyatt:
The tracking event was firing when checkboxes were being deselected
The first commit made it not fire the GA event at all on deselect
The second commit was made after a discussion and decision to make it fire a _different_ event on deselect. It will now fire an event with category 'filterClicked' on select and 'filterRemoved' on deselect. All other values for the event will stay the same

Trello: https://trello.com/c/l6kGfBhh/230-add-missing-facet-tag-email-alert-tracking-and-checkbox-issue


karlbaker02:
Looks 👍 to me. Will let you get FE review

oscarwyatt:
Closing as we've decided on a different technical approach which is to make this an option in govuk_publishing_components

PR #850, opened 2019-02-04
Power owning organisation breadcrumbs from rummager

sihugh:
At present we're fetching a content item to check whether the supplied parent path is a valid organisation.  This was OK when we implemented this initially, but it can be quite slow now because of the latency to content store, and may even cause the page to time out altogether which isn't great.

Because of that, I'm adding another registry to store all the slugs and titles of the organisations that we support.  We can look the `parent` slug up against that and only have to refresh it every so often.  The query for all orgs takes about 0.09 seconds for Rummager to process so shouldn't adversely impact the request that has to populate the cache.

I changed the name of the param to `parent` from `parent_path` as it's no longer a path.

Example URLs:
- https://finder-frontend-pr-850.herokuapp.com/news-and-communications (just a home breadcrumb)
- https://finder-frontend-pr-850.herokuapp.com/news-and-communications?parent=ministry-of-silly-walks (just a home breadcrumb)
- https://finder-frontend-pr-850.herokuapp.com/news-and-communications?parent=ministry-of-defence (org breadcrumb)

https://trello.com/c/3OdXjIoE/321-make-org-breadcrumbs-independent-of-content-store

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -64,16 +64,9 @@ private
   end
 
   def fetch_breadcrumbs
-    parent_content_item = {}
-    unless params.dig("parent_path").to_s.empty?
-      begin
-        parent_content_item = Services.content_store.content_item(params["parent_path"])
-      rescue GdsApi::HTTPNotFound
-        # parent_path is user input so we don't mind if it's bad
-        GovukStatsd.increment("breadcrumb.parent_path_not_found")
-      end
-    end
-    FinderBreadcrumbsPresenter.new(parent_content_item, @content_item)
+    parent_slug = params["parent"]
+    org_info = org_registry[parent_slug] if parent_slug.present?
+    FinderBreadcrumbsPresenter.new(org_info, @content_item)
   end
 
   def finder_presenter_class
@@ -86,5 +79,9 @@ private
 
   def result_set_presenter_class
     ResultSetPresenter
+  end
+
+  def org_registry
+    @org_registry ||= Registries::OrganisationsRegistry.new
   end
 end
--- a/app/lib/registries/organisations_registry.rb
+++ b/app/lib/registries/organisations_registry.rb
@@ -1,0 +1,39 @@
+module Registries
+  class OrganisationsRegistry
+    CACHE_KEY = "registries/organisations".freeze
+
+    def [](slug)
+      organisations[slug]
+    end
+
+    def organisations
+      @organisations ||= Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
+        organisations_as_hash
+      end
+    rescue GdsApi::HTTPServerError
+      GovukStatsd.increment("registries.organisations_api_errors")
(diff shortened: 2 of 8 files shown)


app/controllers/finders_controller.rb
```diff
@@ -64,16 +64,9 @@ def raw_finder
...
-        parent_content_item = Services.content_store.content_item(params["parent_path"])
-      rescue GdsApi::HTTPNotFound
-        #parent_path is user input so we don't mind if it's bad
-        GovukStatsd.increment("breadcrumb.parent_path_not_found")
-      end
-    end
-    FinderBreadcrumbsPresenter.new(parent_content_item, @content_item)
+    parent_slug = params["parent"]
+    org_info = org_registry[parent_slug] if parent_slug.present?
+    FinderBreadcrumbsPresenter.new(org_info, @content_item)
```
bilbof: What will `FinderBreadcrumbsPresenter` do when `org_info` is nil?

app/lib/registries/organisations_registry.rb
```diff
@@ -0,0 +1,39 @@
+module Registries
+  class OrganisationsRegistry
+    CACHE_KEY = "registries/organisations".freeze
```
bilbof: Could do `CACHE_KEY = "#{NAMESPACE}/organisations".freeze` for consistency.

app/lib/registries/organisations_registry.rb
```diff
@@ -0,0 +1,39 @@
...
+        }
+    end
+
+    def fetch_organisations_from_rummager
+      params = {
+        filter_format: 'organisation',
+        fields: %w(title slug),
+        count: 1500
+      }
+      response = Services.rummager.search(params)
```
bilbof: We could use `.with_subsequent_pages` as with `WorldLocationsRegistry`, just in case we create over 1500 organisations in the future.

app/lib/registries/organisations_registry.rb
sihugh: I started with this, but for some reason it couldn't resolve it when running the tests 🤔

app/lib/registries/organisations_registry.rb
sihugh: The 1500 is used in a few other places where orgs are concerned.  I might write a card up for the backlog to fix them all

bilbof:
Also, it might be worth adding a spec for this registry similar to `spec/lib/registries/world_locations_registry_spec.rb`

app/controllers/finders_controller.rb
sihugh: It turns out I'd deleted that bit when I rebased.  Re-added, thanks

sihugh:
@bilbof I've added the spec - I wasn't sure whether it was worth it, given the feature tests, but the error case is good to include so 👍 .  It's worth looking at expanding this registry to help with the organisations facet, but I wanted to keep this change quite small at present.

PR #854, opened 2019-02-05
Permit non whitespaced email title prefixes

bilbof:
Email sign up pages have prefixes like `"News and communications: "`

But the prefixes often have whitespaces at the end. This looks a bit odd, so this computes whether the whitespace is needed at the end.

See `subscription_list_title_prefix` on https://www.gov.uk/api/content/cma-cases/email-signup.
See also https://github.com/alphagov/rummager/pull/1419/files#diff-f74339b4005b4bfdbf8d693f5da78c6bR16

### DIFF
--- a/lib/email_alert_title_builder.rb
+++ b/lib/email_alert_title_builder.rb
@@ -10,7 +10,7 @@ class EmailAlertTitleBuilder
   end
 
   def call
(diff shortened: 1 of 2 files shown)


bilbof:
I'll put this on integration to double check.

PR #856, opened 2019-02-05
Display grouped results with pinned section results

steventux:
https://trello.com/c/Fv5j6ITe/89-production-make-finder-sort-by-topic-prototype-code-production-ready

PR to refactor finder 'group by facets' prototype into a production ready feature.

Depends on https://github.com/alphagov/govuk-app-deployment-secrets/pull/61 for finder sort options and grouping metadata.

screenshot from 2019-02-14 19-10-10


vanitabarrett:
Is the "Loading" text style something which came from a design decision?

## Before
screen shot 2019-02-11 at 15 25 38

## After
screen shot 2019-02-11 at 15 23 57

app/presenters/finder_presenter.rb
```diff
@@ -140,6 +145,9 @@ def sort_options
     return [] if sort.blank?
 
     options = sort.collect do |option|
+      # FIXME: prototype only
+      option['name'] = 'Topic' if option['name'] == 'Most relevant'
```
vanitabarrett: Does this still require refactoring or can the FIXME be deleted?

app/presenters/result_set_presenter.rb
```diff
@@ -49,8 +64,91 @@ def selected_filter_descriptions
...
+    end
+
+    [sector_facets.sort.to_h, other_facets.sort.to_h, all_businesses].inject(&:merge).values
+  end
+
+  def metadata_for_filters(metadata, filters)
+    metadata.select { |m| filters.key?(m[:id]) }
+  end
+
+  def empty_facet_group(key, name)
```
vanitabarrett: I think this can be made private?

app/presenters/result_set_presenter.rb
```diff
@@ -49,8 +64,91 @@ def selected_filter_descriptions
...
+        end
+      end
+
+      all_businesses = {} if all_businesses[:all_businesses][:documents].empty?
+    end
+
+    [sector_facets.sort.to_h, other_facets.sort.to_h, all_businesses].inject(&:merge).values
+  end
+
+  def metadata_for_filters(metadata, filters)
```
vanitabarrett: This method isn't being used at the moment - can it be removed?

app/presenters/result_set_presenter.rb
```diff
@@ -49,8 +64,91 @@ def selected_filter_descriptions
...
+    metadata.select { |m| filters.key?(m[:id]) }
+  end
+
+  def empty_facet_group(key, name)
+    { facet_key: key, facet_name: name, documents: [] }
+  end
+
+  def facet_label_for(key)
```
vanitabarrett: I think this can be made private?

app/presenters/result_set_presenter.rb
```diff
@@ -49,8 +64,91 @@ def selected_filter_descriptions
...
+    allowed_values = finder.facets.map(&:allowed_values).flatten
+    facet = allowed_values.find { |v| v["value"] == key }
+    facet["label"] if facet
+  end
+
+  def sort_by_promoted(results)
```
vanitabarrett: I think this can be made private?

app/presenters/result_set_presenter.rb
```diff
@@ -49,8 +64,91 @@ def selected_filter_descriptions
...
+
+    primary_facet = :sector_business_area
+    primary_facet_group = %W(sector_business_area business_activity)
+
+    facet_filters = @filter_params.without("order", "keywords")
+
+    documents.select! { |d| d[:document][:metadata].present? }
+    sorted_documents = documents.sort { |x, y| x[:document][:title] <=> y[:document][:title] }
+    sorted_documents = sorted_documents.sort do |x, y|
+      return -1 if x[:document][:promoted]
```
vanitabarrett: It would be good if we could somehow have a common method for this and `sort_by_promoted`, so if we ever changed the promoted sorted logic we wouldn't have to update it in 2 places.

app/presenters/result_set_presenter.rb
```diff
@@ -49,8 +64,91 @@ def selected_filter_descriptions
...
+
+      y[:document][:promoted] ? 1 : 0
+    end
+
+    # If no filters are selected then put in all business
```
vanitabarrett: Typo: "all business" => "All Businesses"

app/views/finders/_grouped_heading.mustache
```diff
@@ -0,0 +1 @@
+<h1 id="grouped-heading" class="govuk-heading-l search-results-title">Results</h1>
```
vanitabarrett: Does this need to be in its own mustache template?

app/views/finders/show.html.erb
```diff
@@ -73,20 +73,23 @@
           <div aria-live='assertive' id='js-search-results-info' data-module="remove-filter">
             <% if finder.show_generic_description? %>
```
vanitabarrett: Not part of this change, but we don't use `show_generic_description?` anymore [since it was removed from the business finder](
https://github.com/alphagov/govuk-app-deployment-secrets/pull/45/files), so it may be worth removing this now

app/views/finders/_grouped_heading.mustache
steventux: As far as I can tell it gets used by `live_search.js` as the results count is rewritten. It's completely static though which kinda defeats the point, the other option is to put more exceptions in the live search javascript which seemed to be the worst of two bad options.
My 2p is we should choose 1 way of presenting the result count and do that for all finders. Perhaps some design review of this would be helpful.

app/presenters/result_set_presenter.rb
steventux: Agreed, I think this is possible.

app/presenters/finder_presenter.rb
steventux: This should come from the finder content item sort options.

app/presenters/result_set_presenter.rb
steventux: https://github.com/alphagov/finder-frontend/pull/856/commits/125b3dde5fd533d3dc1287e01923fe91b734a20b tidies this up a bit

app/presenters/finder_presenter.rb
steventux: Removed in https://github.com/alphagov/finder-frontend/pull/856/commits/97fb0e9cfc1090633780c923c9186c6ab6758640 and present in the finder now see: https://github.com/alphagov/govuk-app-deployment-secrets/pull/61/commits/9e1cbc970cb67f6e2405d47d1c2074e32438a6d2

app/views/finders/show.html.erb
steventux: Removed here https://github.com/alphagov/finder-frontend/pull/856/commits/3a90d6f83ebf938a512521f9ff1cf8ebfdcac64c

app/views/finders/_results.mustache
```diff
@@ -6,37 +6,19 @@
...
-        {{/document.metadata}}
-      </dl>
-      {{#document.is_historic}}
-        <p class="historic">First published during the {{document.government_name}}</p>
-      {{/document.is_historic}}
-    </li>
-  {{/documents}}
+  {{#display_grouped_results }}
+    {{#documents_by_facets}}
+      {{> finders/_grouped_results }}
```
vanitabarrett: `grouped_results` contains a `div` and `heading` which means that when this is rendered this actually produces invalid HTML as you're not meant to have a `div` inside a `ul`. I'd suggest something like this:

```
<ul id="this is the overall parent list which wraps all results">
    <li id="this is the title of a sector/facet group">
        Aerospace
        <ul id="this is a subgroup results list">
            <li>Document 1</li>
            <li>Document 2</li>
        </ul>
    </li>
</ul>
```

instead of what we have now, which looks like this:

```
<ul id="this is the overall parent list which wraps all results">
    <div id="this is the title of a sector/facet group">
        <h2>Aerospace</h2>
        <li>Document 1</li>
        <li>Document 2</li>
    </div>
</ul>
```

steventux: Think I've addressed this here: https://github.com/alphagov/finder-frontend/pull/856/commits/d4efab6114b8ed4376147de1304e0640a005a8bb

PR #857, opened 2019-02-05
Suppress filterClicked analytics event when removing facet tag

karlbaker02:
This commit suppresses the `filterClicked` event from firing when a facet tag is removed which corresponds to a text input type (i.e. keywords, published before / after).

Solo: @karlbaker02

https://finder-frontend-pr-857.herokuapp.com/news-and-communications

Trello: https://trello.com/c/FA9IOPMG

### DIFF
--- a/spec/javascripts/live_search_spec.js
+++ b/spec/javascripts/live_search_spec.js
@@ -34,7 +34,12 @@ describe("liveSearch", function(){
   };
 
   beforeEach(function () {
-    $form = $('<form action="/somewhere" class="js-live-search-form"><input type="checkbox" name="field" value="sheep" checked><input type="submit" value="Filter results" class="button js-live-search-fallback"/></form>');
+    $form = $('<form action="/somewhere" class="js-live-search-form">' +
+                '<input type="checkbox" name="field" value="sheep" checked>' +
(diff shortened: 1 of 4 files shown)


spec/javascripts/live_search_spec.js
```diff
@@ -161,21 +166,23 @@ describe("liveSearch", function(){
...
       spyOn(liveSearch, 'updateResults').and.returnValue(promise);
       //spyOn(liveSearch, 'pageTrack').and.returnValue(promise);
-      $form.find('input').prop('checked', false);
+      $form.find('input[name="field"]').prop('checked', false);
 
       liveSearch.formChange();
-      expect(liveSearch.state).toEqual([]);
+      expect(liveSearch.state).toEqual([{name: 'published_at', value: '2004'}]);
       expect(liveSearch.updateResults).toHaveBeenCalled();
       promise.done.calls.mostRecent().args[0]();
```
oscarwyatt: I've checked and this does nothing (the function just returns the first argument for the call, which is undefined), it was added in 2014 and untouched since. Would you mind removing it and the stuff thats commented out

karlbaker02:
Ready for your re-review @oscarwyatt 👍

PR #859, opened 2019-02-06
Added uncheck track category

oscarwyatt:
Adds custom uncheck event category for unchecking of checkbox and option facets. Relies on govuk publishing components >=15.0.0, from my PR https://github.com/alphagov/govuk_publishing_components/pull/729

Trello: https://trello.com/c/l6kGfBhh/230-add-missing-facet-tag-email-alert-tracking-and-checkbox-issue


app/views/components/_option-select.html.erb
```diff
@@ -1,6 +1,5 @@
 <% title_id = "option-select-title-#{title.parameterize}" %>
 <div class="app-c-option-select"
-  <% if options.any? { |option| option[:data_attributes] } %>data-module="track-click"<% end %>
```
bilbof: Why is it necessary to remove this attribute from options in the select input? Does that mean that `uncheck_track_category: "filterRemoved",` won't be applied on this element?

oscarwyatt: This line added `data-module='track-click'` if there is a `data_attributes` key in the hash (which is used for Google analytics tracking). There is a listener that looks for events in the DOM with this attribute and will send the necessary GA tracking event on that change. The checkbox now fires its own GA events (so if this line were left in we'd get two GA events) so it's necessary to remove this data attribute

bilbof: Ah makes sense, thanks for explaining.

PR #861, opened 2019-02-06
Fix tracking bug where showing incorrect result count

peteglondon:
when the pageview event is fired after changing filter/sort.
dimension 27 should contain the number of results.

Currently it shows the number of results for the previous page view.

This is because the event pageview event is fired before
the results are received (the promise for updateResults has not resolved).

Moving triggering the pageview event into the `done` handler for updateResults.

Trello: https://trello.com/c/lEgaXNX7/98-bug-with-finder-event-tracking


peteglondon:
[Bug with finder event tracking](https://trello.com/c/lEgaXNX7/98-bug-with-finder-event-tracking)

PR #863, opened 2019-02-06
Remove Development code for new supergroup finders

bilbof:
This code is no longer required as the finders will be published.


bilbof:
@sihugh good idea. I've left the code, but removed the development finders we were working on.

sihugh:
Nice! 🚀

PR #864, opened 2019-02-07
Fix facet tag space key bug

andysellick:
screen shot 2019-02-07 at 09 54 13

When using the space key on a facet tag the required action was occurring but also causing the link to be activated, reloading the page with an incorrect URL (separate issue, the link wasn't being generated quite right).

This was because the link had `role="button"` on it (correct) but this was causing both the JS to fire and the default action (weirdly difficult to prevent using standard methods e.g. `event.preventDefault`). See the first 'note' section here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

Solution is to turn this into an actual button, works fine now. Also restyled the button slightly to cover the full area of the facet tag, to make clicking on it easier.

Example URL: https://finder-frontend-pr-864.herokuapp.com/news-and-communications
Trello card: https://trello.com/c/yzUrDbq5/263-fix-facet-tag-bug-with-keyboard-interaction

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -9,6 +9,7 @@
 //
 // Read Sprockets README (https://github.com/sstephenson/sprockets#sprockets-directives) for details
 // about supported directives.
+//= require govuk_publishing_components/dependencies
 //
 //= require support
 //
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -23,7 +23,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var inputType = $input.prop('type');
 
       setInputState(elementType, inputType, $input, removeFilterValue, removeFilterFacet);
-      fireRemoveTagTrackingEvent(removeFilterValue, removeFilterFacet)
+      fireRemoveTagTrackingEvent(removeFilterValue, removeFilterFacet);
     }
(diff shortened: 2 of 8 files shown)


vanitabarrett:
The size of the cross is now much smaller and looks like it's nearer the top of the facet tag than the bottom - is that intentional?


## Before
screen shot 2019-02-07 at 13 26 01

## After
screen shot 2019-02-07 at 13 26 04

vanitabarrett:
The focus style looks a bit strange as it's not the full size of the facet tag:

screen shot 2019-02-07 at 13 29 06

I think this and the cross not being in the middle would both be solved by adding `height: 100%` to the remove button link.

andysellick:
@vanitabarrett thanks for the review, good spot on the visual change, thanks. Have fixed, should be better now (although still can't get these tests to pass for some reason, continuing to look into it).

andysellick:
Currently unable to get the tests passing - have added a WIP commit to show some of my working. Current problem:

- JS appears to be executing, output from test shows body has `js-enabled` class
- changes made to the DOM in `remove-filter.js` don't work inside the `start` function, but do further up
- calling `GOVUK.Modules.start()` manually inside the test causes the `remove-filter.js` module to activate
- button click still doesn't cause the checkbox to clear

andysellick:
@karlbaker02 excellent work thankyou!

PR #879, opened 2019-02-13
Persist parent param on finders

bilbof:
Previously the breadcrumbs would be removed on page reload if you changed any search facets. This change will ensure that the breadcrumbs are now persisted.

Authors: Fli.* & Bill

https://trello.com/c/2eWij236/354-bug-disappearing-breadcrumb-from-org-pages

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -13,6 +13,7 @@ class FindersController < ApplicationController
         @results = results
         @content_item = raw_finder
         @breadcrumbs = fetch_breadcrumbs
+        @parent = parent
       end
       format.json do
         if %w[finder search].include? finder_api.content_item['document_type']
@@ -83,5 +84,9 @@ private
 
   def org_registry
     @org_registry ||= Registries::OrganisationsRegistry.new
+  end
+
+  def parent
+    params.fetch(:parent, '')
   end
 end
--- a/app/views/finders/show.html.erb
+++ b/app/views/finders/show.html.erb
@@ -63,6 +63,7 @@
 
 <div class="grid-row">
   <form method="get" action="<%= finder.slug %>" class="js-live-search-form">
+    <input type="hidden" name="parent" value="<%= @parent %>">
     <div class="column-third">
       <%= render finder.facets %>
     </div>
--- a/features/finders.feature
+++ b/features/finders.feature
@@ -72,6 +72,9 @@ Feature: Filtering documents
     And I can see a breadcrumb for all organisations
     And I can see a breadcrumb for the organisation
     And I can see a breadcrumb that not a link for the finder
+    And I sort by most viewed
+    And I filter the results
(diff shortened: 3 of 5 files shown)


app/views/finders/show.html.erb
```diff
@@ -62,6 +62,7 @@
 </header>
 
 <form method="get" action="<%= finder.slug %>" class="js-live-search-form">
+  <div class="hidden"><input type="text" name="parent" value="<%= @parent %>"></div>
```
andysellick: Instead of wrapping this in a hidden div, could this just be `<input type="hidden" ... >` ?

bilbof: Good tip, thank you @andysellick!

app/controllers/finders_controller.rb
```diff
@@ -84,4 +85,8 @@ def result_set_presenter_class
   def org_registry
     @org_registry ||= Registries::OrganisationsRegistry.new
   end
+
+  def parent
+    params.fetch(:parent, '')
```
aliuk2012: I'm not 100% clear on how and where the params are sanitised but if they aren't then should this not clean the values before being used? I know the field is not visible to the user but that wouldn't stop a hacker from attempting to pass in any value they like.

bilbof: @aliuk2012 you're right that a user can provide any value they like for the parent param. Could you provide an example of how this param might be abused?

The parent param is already used for [looking up an organisation in a Hash](https://github.com/alphagov/finder-frontend/pull/879/files/ac7f87e44d48b98b765d173dbe787c454eada966#diff-eb2d1de38a399388bf54245e59b8dfc7R68) and with this change we also render the value as the value of a hidden field.

We don't use this param for any queries or in any other way so we don't need to call `ParamsCleaner` on it. In this case the biggest threat is probably that the hacker manipulates the DOM so the user gets a different experience than intended.

If we try some XSS by setting the parent param as ``, this is rendered in an ERB template, and eventually ends up as this in this DOM:

```
<input type="hidden" name="parent" value="<img src=x\\x11onerror=\&quot;javascript:alert(1)\&quot;>">
```

Because ERB escapes strings when we do things like `value="<%= @parent %>`, this handles the majority of issues like XSS.

https://finder-frontend-pr-879.herokuapp.com/news-and-communications?parent=%3Cimg%20src=x\\x11onerror=\%22javascript:alert(1)\%22%3E.

sihugh:
@bilbof we've got some conflicts on this now

app/controllers/finders_controller.rb
aliuk2012: I don't have an example of what might be passed through but if its not used in any query or in other way then it shouldn't be a problem. Also it's also good to know that ERB provides some security by escaping strings

bilbof:
Thanks, @sihugh. I rebased.

bilbof:
@sihugh sure thing, I added to the feature test for this situation.

PR #881, opened 2019-02-13
Uses component for email and rss links

oscarwyatt:
As the links are now part of the page that gets refreshed when a facet is changed
we no longer need to set the email link parameters with javascript so that code
can be removed

https://trello.com/c/l40K2yBl/353-use-subscription-component-for-email-alerts-and-rss-feed-signup


andysellick:
Is there enough space between the links and the search results?

screen shot 2019-02-14 at 08 48 48

I can add a margin option to the component if that would be useful?

oscarwyatt:
If you think would look better, go wild!

andysellick:
@oscarwyatt I've added a margin option to the subscription links component, once [this PR](https://github.com/alphagov/govuk_publishing_components/pull/748) is merged and put into a new version of the gem, we can add that into this PR.

andysellick:
@oscarwyatt I've added the extra spacing as mentioned, but in the process I've broken your tests. Sorry! Just looking now, but might not finish it before tomorrow.

vanitabarrett:
Have you considered putting the subscription links to the right of the sort button? It would mean there isn't another line separating facet tags and results.

andysellick:
@vanitabarrett rough mockup:

screen shot 2019-02-15 at 08 39 07

andysellick:
@vanitabarrett not sure. Now you've pointed it out it definitely looks cluttered. Didn't it used to be at the top, immediately beneath the page header? Why is it here now? @oscarwyatt

vanitabarrett:
I thought of this because, in one of our user research prototypes, we moved the subscription links to that position:

screen shot 2019-02-15 at 09 26 04

andysellick:
@vanitabarrett @oscarwyatt how about if the subscription links were above the 'sort by' control, beneath the chosen facets? This distances them from the search results (making it look less busy) and associates them more strongly with the idea that the user is signing up to receive updates on the things they have specifically chosen.

screen shot 2019-02-15 at 11 13 07

screen shot 2019-02-15 at 11 14 38

miakepia:
I agree with @andysellick — they make the most sense above the sort control.

bilbof:
I rebased. It shouldn't have conflicts but we still need to fix the filtering issue.

andysellick:
Thanks @bilbof for the rebase. Will look into the updating issue - I probably didn't check the JS when I moved the links, it's might be location specific.

features/step_definitions/filtering_steps.rb
```diff
@@ -51,6 +51,12 @@
   expect(page).to have_css('.filtered-results .document', count: 1)
 end
 
+
+# And("I see email and feed sign up links") do
```
Seraphiana: This test is commented out because I couldn't get it to work.

Since the new code works in a way that (I think at least) is very fragile, by looking for a[href*=".atom" or "email-signup"] I kind of wanted some sort of functional test to confirm that its working as intended. 
I'm not really sure how to achieve this within cucumber though, both to confirm that its not finding the wrong thing, and that it is finding the right thing, though the chance of having more than one anchor with a href containing .atom or email-signup is probably pretty low.

app/presenters/result_set_presenter.rb
```diff
@@ -114,4 +121,63 @@ def sort_option
 
     sort_option
   end
+
+  def tagged_to_all?(facet_key, metadata)
```
bilbof: I couldn't find where `tagged_to_all` and some of the other new methods on this class are used. Should these be removed?

app/presenters/result_set_presenter.rb
```diff
@@ -114,4 +121,63 @@ def sort_option
...
+  end
+
+  def email_signup_link
+    return '' unless finder.respond_to?(:email_alert_signup_url)
+
+    finder.email_alert_signup_url
+  end
+
+  def feed_link
+    return '' unless finder.respond_to?(:atom_url)
```
bilbof: Are these checks due to `finder` potentially being another presenter rather than `FinderPresenter`?

features/step_definitions/filtering_steps.rb
bilbof: It sounds worthwhile so it would be cool to get this working.

I think this may do the trick:

```
# in features/finders.feature:
  Scenario: Email links
    When I view the news and communications finder
    Then I see email and feed sign up links
    And I select a Person
    And I filter the results
    Then I see email and feed sign up links with filters applied


# in features/step_definitions/filtering_steps.rb:
And("I see email and feed sign up links") do
  expect(page).to have_link('Get email alerts', href: "/news-and-communications/email-signup?")
  expect(page).to have_link('Subscribe to feed', href: "/news-and-communications.atom")
end

And("I see email and feed sign up links with filters applied") do
  expect(page).to have_link('Get email alerts', href: "/news-and-communications/email-signup?people%5B%5D=rufus-scrimgeour")
  expect(page).to have_link('Subscribe to feed', href: "/news-and-communications.atom?order=updated-newest&amp;people%5B%5D=rufus-scrimgeour&amp;public_timestamp%5Bfrom%5D=&amp;public_timestamp%5Bto%5D=")
end
```

spec/javascripts/live_search_spec.js
```diff
@@ -190,6 +192,15 @@ describe("liveSearch", function(){
       expect(liveSearch.updateResults).toHaveBeenCalled();
     });
 
+    it("should update email and atom subscription links when checkbox is changed", function(){
+      spyOn(liveSearch, 'updateLinks');
+      $form.find('input[name="field"]').prop('checked', false);
+
+      liveSearch.formChange();
+      expect(liveSearch.state).toEqual([{name: 'published_at', value: '2004'}]);
+      expect(liveSearch.updateLinks).toHaveBeenCalled();
```
andysellick: Presumably you're not explicitly checking the new href attributes on the links because you don't know what they'd be? If that's the case we could at least test that they are no longer equal to their initially set values.

app/presenters/advanced_search_result_set_presenter.rb
```diff
@@ -4,7 +4,6 @@ class AdvancedSearchResultSetPresenter < ResultSetPresenter
   def to_hash
     super
       .merge(applied_filters: applied_filters_or_all_subgroups)
-        .except(:atom_url)
```
andysellick: Comparing the deployment to live, it doesn't look like this version is updating the atom link in the head anymore...

`<link rel="alternate" type="application/atom+xml" title="ATOM" href="/news-and-communications.atom">`

Is this intentional?

app/views/finders/_results.mustache
```diff
@@ -1,11 +1,5 @@
-{{#atom_url}}
-  <div class="feed">
-    <span>Get updates to this list</span>
-    <a href="{{atom_url}}">feed</a>
-  </div>
-{{/atom_url}}
+<ul data-module="track-click" class="finder-results">
```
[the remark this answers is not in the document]
Seraphiana: oh, that'll be because it got removed in the merge by accident then the class got readded, I'll change the order back.

spec/javascripts/live_search_spec.js
Seraphiana: added new unit test that should provide more certainty

app/presenters/result_set_presenter.rb
Seraphiana: yeah, I believe so

spec/javascripts/live_search_spec.js
```diff
@@ -190,6 +192,15 @@ describe("liveSearch", function(){
       expect(liveSearch.updateResults).toHaveBeenCalled();
     });
 
+    it("should update email and atom subscription links when checkbox is changed", function(){
```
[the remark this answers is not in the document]
Seraphiana: I've added unit testing for the behaviours of the function itself, at "should replace links with new links when state changes"
I'm half tempted to rename this test, since it doesn't actually test that the links are updated, just that the function is wired up.
As for checking when other types of facet are changed, up to you. I feel like the tests would likely overlap existing testing, but I'm not familiar enough with the existing level of testing to judge that either way with certainty.

PR #884, opened 2019-02-14
Remove supertypes to document_types conversions

koetsier:
Revert translations of supertypes to content types in search

See: https://github.com/alphagov/finder-frontend/pull/606

Supertypes were translated to content types in the expectation that
supertypes would be deprecated in the search api.

This is actually quite complicated to do (for instance handling cases where
supergroups are 'other') and so we will keep indexing supertypes in
ElasticSearch.

See also: https://github.com/alphagov/rummager/pull/1426

Trello: https://trello.com/c/jInE7P59/362-remove-supertypes-to-documenttypes-conversions


features/support/env.rb
```diff
@@ -16,6 +16,7 @@
 require 'slimmer/test'
 
 GovukTest.configure
+Capybara.javascript_driver = :selenium_chrome
```
andysellick: Was this an attempt to get that failing test to pass? Seems like changing this could have consequences.

koetsier: yes, for debugging - this is a WIP

features/fixtures/statistics.json
```diff
@@ -25,7 +25,6 @@
   "details":{
     "document_noun":"statistic",
     "filter":{
-      "content_purpose_supergroup": "research_and_statistics"
```
bilbof: Does the stats supergroup finder no longer need this field?

features/support/rummager_url_helper.rb
```diff
@@ -48,57 +48,17 @@ def policy_search_fields
...
-      utaac_decision
-      world_location_news_article
-      world_news_story
-      written_statement
-    )
-
     base_search_params.merge(
       'fields' => news_and_communications_search_fields.join(','),
-      'filter_content_store_document_type' => supergroup_document_types,
+      'filter_content_purpose_supergroup' => 'news_and_communications',
```
bilbof: Really good change.

features/fixtures/statistics.json
koetsier: The test doesn't really - this was part of testing cps->content types conversion

PR #888, opened 2019-02-15
Dynamic values in email subscription titles

bilbof:
When a user signs up for an email subscription, we generate a subscriber list for the user with their selected filters. We also send email-alert-api a title for the subscriber list.

This adds more detail to the title if the filters are dynamic (filters without allowed fields set in the content item, such as organisations, people, and so on, which are populated at runtime from other sources). Examples of a title we generate below.

Before:

"News with 3 people and 2 organisations"

After:

"News with people of Harry Potter, Ron Weasley, and 1 other, and organisations of HM Treasury and Cabinet Office"

It is possible to display the actual values rather than the counts because we now
have registries - cached collections of people, organisations, world locations, etc. -
that we can use to lookup a beautified value for a given slug.

Authors: Felicity and Bill

https://trello.com/c/J6ZHmdgi/349-improve-email-signups-description

### DIFF
--- a/app/lib/registries/base_registries.rb
+++ b/app/lib/registries/base_registries.rb
@@ -5,7 +5,10 @@ module Registries
     def all
       @all ||= {
         'world_locations' => world_locations,
-        'part_of_taxonomy_tree' => topic_taxonomy
+        'all_part_of_taxonomy_tree' => topic_taxonomy,
+        'part_of_taxonomy_tree' => topic_taxonomy,
+        'people' => people,
+        'organisations' => organisations,
       }
     end
 
@@ -17,6 +20,14 @@ module Registries
 
     def topic_taxonomy
       @topic_taxonomy ||= TopicTaxonomyRegistry.new
+    end
+
+    def people
+      @people ||= PeopleRegistry.new
+    end
+
+    def organisations
+      @organisations ||= OrganisationsRegistry.new
     end
   end
 end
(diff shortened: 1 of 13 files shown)


app/lib/registries/base_registries.rb
```diff
@@ -5,7 +5,10 @@ class BaseRegistries
     def all
       @all ||= {
         'world_locations' => world_locations,
-        'part_of_taxonomy_tree' => topic_taxonomy
+        'all_part_of_taxonomy_tree' => topic_taxonomy,
+        'part_of_taxonomy_tree' => topic_taxonomy,
+        'people' => people,
+        'organisations' => organisations,
```
sihugh: I wasn't sure whether adding things here would cause them to be used to populate the facets in finders.  This might be fine, but might also not be ideal yet.

bilbof: After this change we will still default to the title value that rummager provides us with for people, organisations and world locations facets.

```
    label = value.fetch("title", find_facet_title_by_slug(slug, facet_key))
```

spec/helpers/registry_spec_helper.rb
```diff
@@ -0,0 +1,53 @@
+module RegistrySpecHelper
+  def stub_people_registry_request
+    stub_request(:get, "http://search.dev.gov.uk/search.json?count=1500&fields%5B%5D=slug&fields%5B%5D=title&filter_format=person&start=0")
```
koetsier: pedantic point: you could use  ``` .with(query: {"a" => ["b", "c"]}``` to make this clearer? If you use ```hash_including``` you could even omit unimportant parameters

bilbof: Thanks a lot, that's much nicer. Changed.

PR #890, opened 2019-02-15
Show all values for People, Organisations and World Locations regardless of applied filters

karlbaker02:
This PR updates finders to show all available _People_, _Organisations_ and _World Locations_ values in dynamic facets, regardless of whether these values are applicable to the currently selected facets.

The reason behind this is to ensure that we don't give the appearance of les options being available for these facets than there actually are, and also to ensure that we don't incorrectly constrain a user's search as they select and de-select filters - something which can happen right now, as the facets aren't reloaded on-the-fly with JS enabled but instead only on page load/reload.

Further information can be found at https://trello.com/c/Bc9UKt5W.

Solo: @karlbaker02

https://finder-frontend-pr-890.herokuapp.com/news-and-communications

### DIFF
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -105,7 +105,12 @@ private
   def augment_facets_with_dynamic_values(content_item, search_response)
     search_response.fetch("facets", {}).each do |facet_key, facet_details|
       facet = content_item['details']['facets'].find { |f| f['key'] == facet_key }
-      facet['allowed_values'] = allowed_values_for_facet_details(facet_key, facet_details) if facet
+
+      if registries.all.has_key?(facet_key) && facet
+        facet['allowed_values'] = allowed_values_from_registry(facet_key)
+      elsif facet
+        facet['allowed_values'] = allowed_values_for_facet_details(facet_key, facet_details)
+      end
     end
   end
 
@@ -113,6 +118,12 @@ private
     facet_details.fetch("options", {})
       .map { |f| f.fetch("value", {}) }
       .map { |value| present_facet_option(value, facet_key) }
+      .reject { |f| f["label"].blank? || f["value"].blank? }
+  end
+
+  def allowed_values_from_registry(facet_key)
+    registries.all[facet_key].values
+      .map { |_, results| present_facet_option(results, facet_key) }
       .reject { |f| f["label"].blank? || f["value"].blank? }
   end
 
--- a/app/lib/registries/organisations_registry.rb
+++ b/app/lib/registries/organisations_registry.rb
@@ -15,6 +15,10 @@ module Registries
       {}
     end
 
+    def values
+      organisations
+    end
+
   private
 
     def organisations_as_hash
@@ -30,7 +34,8 @@ module Registries
       params = {
         filter_format: 'organisation',
         fields: %w(title slug acronym),
-        count: 1500
+        count: 1500,
+        order: 'title'
       }
       response = Services.rummager.search(params)
       response['results']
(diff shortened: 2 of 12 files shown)


app/lib/finder_api.rb
```diff
@@ -116,6 +121,12 @@ def allowed_values_for_facet_details(facet_key, facet_details)
       .reject { |f| f["label"].blank? || f["value"].blank? }
   end
 
+  def allowed_values_from_registry(facet_key)
+    registries.all[facet_key].values
+      .map { |_, results| present_facet_option(results, facet_key) }
```
bilbof: Not sure if `_, ` is needed, or you can just do `map { |value| ... }`

app/lib/finder_api.rb
```diff
@@ -116,6 +121,12 @@ def allowed_values_for_facet_details(facet_key, facet_details)
...
+      .reject { |f| f["label"].blank? || f["value"].blank? }
```
bilbof: Might be worth doing `.reject { |option| option["slug"].blank? }`  before the `map`, as this should have the same effect as the above `reject`.

app/lib/registries/organisations_registry.rb
```diff
@@ -15,6 +15,10 @@ def organisations
       {}
     end
 
+    def values
```
bilbof: Might be worth renaming the organisations method to values (and in other registries).

app/lib/registries/organisations_registry.rb
```diff
@@ -30,7 +34,8 @@ def fetch_organisations_from_rummager
       params = {
         filter_format: 'organisation',
         fields: %w(title slug acronym),
-        count: 1500
+        count: 1500,
+        order: 'title'
```
bilbof: Clever, no need to sort later :)

app/lib/finder_api.rb
karlbaker02: I'll double check.

app/lib/registries/organisations_registry.rb
karlbaker02: I have a slight reservation that Rummager shouldn't perhaps be used for this sort of thing - it should provide the data and then the apps should do filtering - but given the scale of data and the fact that Elasticsearch is optimised for this sort of thing, it makes sense.

karlbaker02:
Fixing the tests...

app/lib/finder_api.rb
karlbaker02: I don't think that'd have _exactly_ the same effect - `option['title']` could still be populated, but would then be excluded, which doesn't happen in the current setup. This is also the logic that exists in `allowed_values_for_facet_details`.

app/lib/registries/organisations_registry.rb
karlbaker02: I will raise another PR for this (there will be quite a few changes) but also to rename the method from `values` to `all` and then use this to make `allowed_values_from_registry` clearer.

app/lib/finder_api.rb
```diff
@@ -105,7 +105,12 @@ def augment_content_item_details_with_results(content_item, search_response)
   def augment_facets_with_dynamic_values(content_item, search_response)
     search_response.fetch("facets", {}).each do |facet_key, facet_details|
       facet = content_item['details']['facets'].find { |f| f['key'] == facet_key }
-      facet['allowed_values'] = allowed_values_for_facet_details(facet_key, facet_details) if facet
+
+      if registries.all.has_key?(facet_key) && facet
+        facet['allowed_values'] = allowed_values_from_registry(facet_key)
+      elsif facet
+        facet['allowed_values'] = allowed_values_for_facet_details(facet_key, facet_details)
```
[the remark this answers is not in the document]
karlbaker02: Agree, there are definitely refactorings that can be made.

koetsier:
Looks good! Could maybe add a cucumber test that shows what this code does?

karlbaker02:
@koetsier Test added

PR #891, opened 2019-02-15
Add acronyms to organisations registry

bilbof:
This adds an additional field (acronyms) to the results
we fetch and cache from rummager for the organisations
registry.

Later we will use this extra field to allow users
to filter by looking for an acronym in the autocomplete
facet. This just adds them to the registry.

https://trello.com/c/Ql3rrz9O/374-allow-users-to-filter-orgs-using-abbreviations

### DIFF
--- a/app/lib/registries/organisations_registry.rb
+++ b/app/lib/registries/organisations_registry.rb
@@ -22,14 +22,14 @@ module Registries
         .reject { |result| result['slug'].empty? || result['title'].empty? }
         .each_with_object({}) { |result, orgs|
           slug = result['slug']
-          orgs[slug] = result.slice('title', 'slug')
+          orgs[slug] = result.slice('title', 'slug', 'acronym')
         }
     end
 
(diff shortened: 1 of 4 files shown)


sihugh:
I guess we should get this into prod and make sure the cache has been refreshed before we deploy anything that relies on it?

sihugh:
Or is this the thing that clashes with Karl's PR?

bilbof:
@sihugh we should deploy this before we merge the code that will use the acronyms. The code we write to use the acronyms should be able to handle cases where no acronym is provided (only ~half of organisations have an acronym).

It might be nice to add a rake task to clear the cache of set registries, what do you think?

This doesn't clash with Karl's PR (https://github.com/alphagov/finder-frontend/pull/890); I already merged the PR that created conflicts there (https://github.com/alphagov/finder-frontend/pull/888). We should make sure when reviewing Karl's PR that the cache structure is the same or the difference is handled in some way.

PR #892, opened 2019-02-18
Fli.* - display message if finder returns zero results

Seraphiana:
added a message to be displayed if a particular query returns zero results.

added to existing cucumber tests to validate absence or presence of this message.

story card: https://trello.com/c/uA82ETNg/368-display-a-message-when-a-finder-returns-0-results

Example page: https://finder-frontend-pr-892.herokuapp.com/news-and-communications?keywords=&level_one_taxon=ba951b09-5146-43be-87af-44075eac3ae9&level_two_taxon=&people%5B%5D=george-zambellas&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&order=updated-newest

### DIFF
--- a/app/views/finders/_results.mustache
+++ b/app/views/finders/_results.mustache
@@ -43,7 +43,6 @@
   <div class='no-results'>
     <h2 class='gem-c-heading'>There are no matching results.</h2>
     <p>Try making your search broader.</p>
-    {{results}}
   </div>
 {{/zero_results}}
 
--- a/app/views/finders/_results.mustache
+++ b/app/views/finders/_results.mustache
@@ -39,4 +39,12 @@
   {{/documents}}
 </ul>
 
+{{#zero_results}}
+  <div class='no-results'>
+    <h2 class='gem-c-heading'>There are no matching results.</h2>
+    <p>Try making your search broader.</p>
+    {{results}}
+  </div>
+{{/zero_results}}
+
 {{{next_and_prev_links}}}
--- a/features/finders.feature
+++ b/features/finders.feature
@@ -26,12 +26,14 @@ Feature: Filtering documents
     Given a collection of documents exist
     When I search documents by keyword
     Then I see all documents which contain the keywords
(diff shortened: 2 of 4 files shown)


features/finders.feature
```diff
@@ -26,12 +26,14 @@ Feature: Filtering documents
...
     When I search documents by keyword
     Then I see all documents which contain the keywords
+    And there is not a zero results message
 
   Scenario: Hiding keyword search
     Given no results
     When I view the finder with no keywords and no facets
     Then I see no results
     And there is no keyword search box
+    And there is a helpful message
```
bilbof: 'helpful' isn't too specific. Perhaps "And there is a no results message"

app/views/finders/_results.mustache
```diff
@@ -39,4 +39,12 @@
   {{/documents}}
 </ul>
 
+{{#zero_results}}
+  <div class='no-results'>
+    <h2 class='gem-c-heading'>There are no matching results.</h2>
+    <p>Try making your search broader.</p>
+    {{results}}
```
bilbof: What is the intent of `{{results}}` here?

Seraphiana: good catch, that's my mistake, will remove.

features/finders.feature
Seraphiana: renamed

app/views/finders/_results.mustache
```diff
@@ -39,4 +39,11 @@
   {{/documents}}
 </ul>
 
+{{#zero_results}}
+  <div class='no-results'>
+    <h2 class='gem-c-heading'>There are no matching results.</h2>
```
andysellick: Obviously this has been merged now, and this is a mustache template so you couldn't have pulled in a component, but we should try to avoid using component classes outside of a component. It would be better to use some styles from GOV.UK Frontend (for example https://design-system.service.gov.uk/styles/typography/) as that gets pulled in with the components gem.

PR #893, opened 2019-02-18
Adds "skip to results" link to improve keyboard navigation

aliuk2012:
When a user is using keyboard navigation (eg. using 'tab' key to skip to next focusable element) they should not have to go through all the filters before seeing the results.

skip-to-results

Test page: https://finder-frontend-pr-893.herokuapp.com/news-and-communications

Trello card: https://trello.com/c/HrwDU1do/387-improve-navigations-for-keyboard-users-and-screen-readers


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -446,3 +446,9 @@
     }
   }
 }
+
+/* Override class to bleed more into govuk-form-group class */
+.govuk-skip-link:focus {
```
andysellick: We shouldn't be adding glue code like this - can this be built into the component instead?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -446,3 +446,9 @@
...
+  position: relative !important;
```
andysellick: I assume there's a good reason for the `!important` but we should try to avoid this - currently this rule will apply to any `govuk-skip-link` in finder-frontend.

app/views/finders/_facet_collection.html.erb
```diff
@@ -8,7 +8,10 @@
     id: "finder-keyword-search",
     controls: "js-search-results-info"
   } if finder.show_keyword_search? %>
-
+    <%= render "govuk_publishing_components/components/skip_link", {
```
andysellick: Nitpick: indentation.

andysellick:
@aliuk2012 sorry, I've reviewed this and only just noticed it's a WIP. Apologies.

app/views/finders/_facet_collection.html.erb
aliuk2012: 👍 good spot though

app/assets/stylesheets/finder_frontend.scss
aliuk2012: I agree, the reason for using `!important` is that govuk frontend uses it and so we have to override it. Anyway, I've found a better way which doesn't need this override because the only reason I added this was to try add some spacing between the skip link and "Show brexit" checkbox.

app/assets/stylesheets/finder_frontend.scss
aliuk2012: We shouldn't need too add this to component, its just used to add spacing. but I've fixed that

features/step_definitions/filtering_steps.rb
```diff
@@ -526,3 +526,16 @@
     end
   end
 end
+
+And(/^I press (tab) key to navigate$/) do |key|
+  find_field('Search').send_keys key.to_sym
+end
+
+Then(/^I should (see|not see) a "Skip to results" link$/) do |can_be_seen|
+  visibility = can_be_seen === 'see'
```
sihugh: this should be an `==` - pesky javascript :-D

aliuk2012: 🙄 thats what happens when you've just been coding html,css & js for the past 3 months.

Well spotted @sihugh 👍

aliuk2012:
I'm not merging this just yet as I need to add some tracking to the skip links

selfthinker:
This definitely makes sense from an accessibility perspective. :+1:

PR #903, opened 2019-02-20
Restrict pinning to grouped display

steventux:
https://trello.com/c/pDDHPDzM/172-bug-pinned-items-appearing-within-finder-results-even-when-not-grouped-by-topic

Bugfix for promoted / pinned rendering in non-grouped results display.
Pinned items should only appear when rendering grouped results.

This PR contains a bit of refactoring to extract grouped display logic from the `ResultSetPresenter` which should all be limited to the penultimate commit.

#### Non grouped
screenshot from 2019-02-20 10-41-04

#### Grouped
screenshot from 2019-02-20 10-51-48


steventux:
@MuriloDalRi I've added more commits since initial review, would you mind re-reviewing please?

PR #904, opened 2019-02-20
Add metadata fields to statistics finder.

koetsier:
Add metadata fields to statistics finder.

Adds Document Type and Publishing orgs as metadata.
Adds Updated date and collection to 'published' statistics.
Adds Release date and cancelled/confirmed state to 'upcoming' stats

The search api only indexes 'updated date' and 'collection' metadata
on 'published' stats and 'release date' and 'state' on 'upcoming'
stats so no extra logic is needed.

See Trello: https://trello.com/c/XLdskaKe/382-make-stats-results-metadata-match-existing-metadata

image

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -48,7 +48,7 @@
     dt {
       display: none;
 
-      &.metadata-date-label {
+      &.metadata-label {
         @include inline-block;
       }
     }
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -163,7 +163,9 @@ private
   # Add a finder with the base path as a key and the finder name
   # without filetype as the value; example:
   # "/guidance-and-regulation" => "guidance_and_regulation"
-  FINDERS_IN_DEVELOPMENT = {}.freeze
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -48,7 +48,7 @@
     dt {
       display: none;
 
-      &.metadata-date-label {
+      &.metadata-date-label, &.metadata-text-label {
```
andysellick: If `metadata-date-label` and `metadata-text-label` have exactly the same styles, can I suggest merging them into one `metadata-label` (perhaps) ?

karlbaker02:
As a general question, do we want to make this more or less exactly like the Whitehall finder in not showing the labels for the metadata, except for the collection.

**Example**

_Whitehall_

screen shot 2019-02-26 at 10 51 33

_This PR_

screen shot 2019-02-26 at 10 51 25

koetsier:
The metadata in Whitehall is a bit different, but the existing code in Finder-frontend styles metadata like this. Maybe we need some design input?

andysellick:
If this is a question of style, go with the finder-frontend look. If this is a question of content, I'm not sure.

PR #907, opened 2019-02-22
Add filter to option select component

andysellick:
Improve the option select component by adding an option to provide a filter field, which shows/hides the checkboxes based on user input.

screen shot 2019-02-27 at 10 11 08

How it works:

- filtering is applied as a user types, with `setTimeout` used to provide a delay to avoid unnecessary processing
- filtering is done entirely in-page, no backend involvement, so reloading or returning to the page completely resets the filter
- string matching is case insensitive, ignores punctuation and ignores the difference between & and 'and'
- checked checkboxes are always shown regardless of what has been typed into the filter
- checkbox labels are processed and cached on page load to optimise filtering
- a hidden element is used to announce to screen readers the number of results that have been found with the filter, and also to provide context that something is happening when the filter is used

Test: https://finder-frontend-pr-907.herokuapp.com/component-guide/option-select/with_filter

Trello card: https://trello.com/c/ppFUeOhY/385-improve-option-select-functionality

### DIFF
--- a/app/assets/javascripts/components/option-select.js
+++ b/app/assets/javascripts/components/option-select.js
@@ -13,6 +13,9 @@
     this.$labels = this.$optionSelect.find(".govuk-checkboxes__item");
     this.$optionsContainer = this.$optionSelect.find('.options-container');
     this.$optionList = this.$optionsContainer.children('.js-auto-height-inner');
+    this.$allCheckboxes = this.$optionsContainer.find('.govuk-checkboxes__item');
+    this.$filter = this.$optionsContainer.find('input[name="filter"]');
+    this.checkedCheckboxes = [];
 
     this.attachCheckedCounter();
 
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/application.js
```diff
@@ -42,7 +42,4 @@ jQuery(function($) {
       templateDir: templateDir
     });
   }
-
-  var $buttons = $("label.block-label input[type='checkbox']");
-  new GOVUK.SelectionButtons($buttons);
```
aliuk2012: Why is this being removed?

aliuk2012:
Could you add the "How it works" to the commits description?

app/assets/javascripts/application.js
andysellick: Excellent question. Sorry, I should probably do this in a separate PR (in fact now you've mentioned it I think I will) - I noticed while I was working on this that this code was throwing an error in the console because `GOVUK.SelectionButtons` couldn't be found (turns out it was removed ages ago) so I took it out. I'll put it in a separate PR.

vanitabarrett:
This could probably be a separate piece of work, but would some kind of clear filter functionality be useful?

andysellick:
@vanitabarrett yes, that's one of the things we considered. For the moment though, I'm keeping this PR as simple as I can so we can get it in front of users and try it out. I'm collating other issues relating to this component in this card: https://trello.com/c/8IBlJjn1/465-fix-option-select-issues so I've just added that.

PR #911, opened 2019-02-25
Only use people that are associated with documents

sihugh:
Our person registry included more people than were associated with results which slowed down pages that used this large facet.  

This PR changes the query that populates the registry to just fetch people that are associated with documents in our search index.

This will mean reducing to 490 people rather than the 3368 we were rendering before.

Uses https://www.gov.uk/api/search.json?facet_people=1500,order:value.title&count=0
instead of: https://www.gov.uk/api/search.json?filter_format=person&fields=slug&fields=title&order=title

https://finder-frontend-pr-911.herokuapp.com/news-and-communications

https://trello.com/c/yp9p8lxj/438-scope-dynamic-facets-to-most-popular-options

### DIFF
--- a/app/lib/registries/people_registry.rb
+++ b/app/lib/registries/people_registry.rb
@@ -23,23 +23,22 @@ module Registries
 
     def people_as_hash
       GovukStatsd.time("registries.people.request_time") do
-        fetch_people_from_rummager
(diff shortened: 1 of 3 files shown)


karlbaker02:
We were using 1500 before, unless I'm mistaken?

PR #914, opened 2019-02-26
Put closed orgs at the bottom of the facet

sihugh:
We need to show closed organisations as they still have content tagged to them.  To help users navigate a long list, we're going to move them to the bottom.  At present they appear in a clump as the orgs list is sorted by title, and closed orgs start with "Closed organisation: ".

We also want to use a case insensitive sort because we have one org that starts with a lower case "j".
(Search for `jHub`)

https://finder-frontend-pr-914.herokuapp.com/news-and-communications

https://trello.com/c/leCxXdgK/367-closed-orgs-should-appear-at-the-bottom-of-the-org-autocomplete-facet

### DIFF
--- a/app/lib/registries/organisations_registry.rb
+++ b/app/lib/registries/organisations_registry.rb
@@ -25,6 +25,7 @@ module Registries
       GovukStatsd.time("registries.organisations.request_time") do
         fetch_organisations_from_rummager
           .reject { |result| result['slug'].empty? || result['title'].empty? }
+          .sort_by { |result| result['title'].sub("Closed organisation: ", "ZZ").upcase }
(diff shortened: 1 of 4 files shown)


app/lib/registries/organisations_registry.rb
```diff
@@ -25,6 +25,7 @@ def organisations_as_hash
       GovukStatsd.time("registries.organisations.request_time") do
         fetch_organisations_from_rummager
           .reject { |result| result['slug'].empty? || result['title'].empty? }
+          .sort_by { |result| result['title'].sub("Closed organisation: ", "ZZ").upcase }
```
koetsier: nice!

PR #922, opened 2019-02-28
Adds new Policy Papers and Consultation finder

aliuk2012:
## Introduces new Policy papers and consultations finder

As a user, when I view this finder it should be scoped to just policy papers and consultations and I should have the option to filter by:
- Policy papers
- Consultations (open)
- Consultations (closed)

-----

**Ticket:** https://trello.com/c/40IA4pPe/241-introduce-documenttype-facet-behaviour-on-policy-papers-consultations-finder

**Demo:** https://finder-frontend-pr-922.herokuapp.com/policy-papers-and-consultations

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,6 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require File.expand_path('../config/application', __FILE__)
+require File.expand_path('config/application', __dir__)
 
 FinderFrontend::Application.load_tasks
--- a/app/lib/filters/text_filter.rb
+++ b/app/lib/filters/text_filter.rb
@@ -1,7 +1,27 @@
 module Filters
   class TextFilter < Filter
     def value
-      Array(params)
+      Array(parsed_value)
+    end
+
+  private
+
+    def parsed_value
+      return if params.blank?
+
+      if multi_value?
+        option_lookup.select { |key, _| params.include? key }.values.flatten
+      else
+        params
+      end
+    end
+
+    def multi_value?
+      facet.has_key?('option_lookup')
+    end
+
+    def option_lookup
+      @option_lookup ||= facet['option_lookup']
     end
   end
 end
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -164,7 +164,8 @@ private
   # without filetype as the value; example:
   # "/guidance-and-regulation" => "guidance_and_regulation"
   FINDERS_IN_DEVELOPMENT = {
-    "/statistics" => "statistics",
+    "/policy-papers-and-consultations" => 'policy_and_engagement',
+    "/statistics" => "statistics"
   }.freeze
 
   def development_env_finder_json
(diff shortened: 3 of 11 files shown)


Rakefile
```diff
@@ -1,6 +1,6 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require File.expand_path('../config/application', __FILE__)
+require File.expand_path('config/application', __dir__)
```
karlbaker02: Is this required?

aliuk2012: Yes, it was an issue that was raised by govuk-lint-ruby. Although not strictly related to our work we did.

app/lib/filters/text_filter.rb
```diff
@@ -1,7 +1,27 @@
...
+  private
+
+    def parsed_value
+      return if params.blank?
+
+      if multi_value?
+        option_lookup.select { |key, _| params.include? key }.values.flatten
+      else
+        params
+      end
```
andysellick: I think this if statement could be written as...

```
return option_lookup.select { |key, _| params.include? key }.values.flatten if multi_value?

params
```

app/views/components/docs/option-select.yml
```diff
@@ -248,3 +248,26 @@ examples:
...
+      title: Groups of clothes
+      options_container_id: groups_of_clothes
+      options:
+      - value: hats
+        label: Hats
+        id: hats
+      - value: socks
+        label: socks
+        id: socks
+      option_lookup:
```
andysellick: What does `option_lookup` actually do in the component? I can't see any changes to the component itself other than to this yml file, and I checked out the branch and removed the `option_lookup` data from this file, and everything rendered exactly as it had before. Have all the changes been committed?

app/lib/filters/text_filter.rb
aliuk2012: I find @karlbaker02 implementation easy to read because if I skim through the code I can see that this method could return two possible things depending on a condition. I don't have read the whole line only to discover that its only likely to return that value if a condition is met.

karlbaker02: I would normally be with you @andysellick on this kind of defensive programming but I'm not sure what the Ruby convention is on that front - I've never seen it done elsewhere and am of a mind that it perhaps is a little easier to understand with the `if else`

app/views/components/docs/option-select.yml
karlbaker02: `option_lookup` maps a bunch of keys to the actual values they stand-in for. In `allowed_values` we use the _keys_ defined in `option_lookup` as the values which are then populated in each of the checkboxes. 

To go with an example, `closed_consultations` in for this policy finder maps to an array `['closed_consultation', 'consultation_outcome']`. In the component, the checkbox rendered with have the value `closed_consultation`, and when that checkbox is selected, it'll get mapped behind the scenes to the actual values which that key represents, so the search will be performed against doc types of `closed_consultation` and `consultation_outcome`.

app/lib/filters/text_filter.rb
andysellick: Well, if you both think so - I was just trying to reduce the code a little.

app/views/components/docs/option-select.yml
andysellick: This documentation is for the component, and this isn't a change to the component, it's a change to the data that you're passing to the component. It's misleading to have this documentation because it suggests that the component accepts something called `option_lookup` when in fact it doesn't.

I agree that documenting this somewhere is a good idea, but this isn't the place for it.

karlbaker02: This has now been removed.

karlbaker02:
All fixed now @andysellick and ready for review again 👍

PR #924, opened 2019-03-01
Option select fixes

andysellick:
Fix some issues with the option select component, specifically:

- replace the up/down arrow image with an SVG (won't appear in IE8 but have checked everything up and appears fine)
- add explicit styles to give the expand/collapse button the standard focus yellow outline
- rewrite the height calculation code to fix an issue where heights were not being calculated properly when the component was loaded inside a hidden div (on mobile, where facets are hidden by default)

Test:

- [Test link (general)](https://finder-frontend-pr-924.herokuapp.com/component-guide/option-select)
- [Test link for height fix (load on mobile)](https://finder-frontend-pr-924.herokuapp.com/news-and-communications?parent=&keywords=&level_one_taxon=&people%5B%5D=george-zambellas&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&order=updated-newest)

https://trello.com/c/8IBlJjn1/465-fix-option-select-issues


app/views/components/_option-select.html.erb
```diff
@@ -6,6 +6,8 @@
   <div class="container-head js-container-head">
     <div class="option-select-label" id="<%= title_id %>">
       <%= title %>
+      <svg version="1.1" viewBox="0 0 1024 1024" xmlns="http://www.w3.org/2000/svg" class="app-c-option-select__icon app-c-option-select__icon--up"><path d="m798.16 609.84l-256-256c-16.683-16.683-43.691-16.683-60.331 0l-256 256c-16.683 16.683-16.683 43.691 0 60.331s43.691 16.683 60.331 0l225.84-225.84 225.84 225.84c16.683 16.683 43.691 16.683 60.331 0s16.683-43.691 0-60.331z"/></svg>
+      <svg version="1.1" viewBox="0 0 1024 1024" xmlns="http://www.w3.org/2000/svg" class="app-c-option-select__icon app-c-option-select__icon--down"><path d="m225.84 414.16l256 256c16.683 16.683 43.691 16.683 60.331 0l256-256c16.683-16.683 16.683-43.691 0-60.331s-43.691-16.683-60.331 0l-225.84 225.84-225.84-225.84c-16.683-16.683-43.691-16.683-60.331 0s-16.683 43.691 0 60.331z"/></svg>
```
aliuk2012: Do we need svg? Could we not replace it with a css-only version using `:after` pseudo element?
GOV.UK Design System uses something similar for the "Back link" icon and this should have better browser support than using svg.

https://design-system.service.gov.uk/components/back-link/default/index.html
https://design-system.service.gov.uk/components/back-link/
https://css-tricks.com/snippets/css/css-triangle/

andysellick: We could, but technically they're a hack and making sure they work properly in all the browsers we want can be fiddly. I'd rather use SVG here because I know it will either work or not appear, and the support is more than we need.

andysellick:
@aliuk2012 I've added the additional tests I mentioned as a separate commit - once you're happy I can fix it up.

aliuk2012:
Could you fix the failing tests before I approve this PR?

PR #925, opened 2019-03-01
Adds new Statistics and Research finder

aliuk2012:
Adds the statistics and research finder.

screen shot 2019-03-08 at 14 08 54

Since the three options in the new facet are mutually exclusive they are represented by a new radio button facet. This means that one option from this facet is always set and cannot be turned off, so having a facet tag for this facet makes no sense - if you removed it, it would immediately reappear as the default option.

This code change includes support for any facets that do not require a facet tag, not just this one. In order to be accessible any facets without a facet tag now have their details output into the `visually-hidden` element that also contains the 'sort by' information.

Ticket: https://trello.com/c/pNXeZLrW/403-introduce-documenttype-facet-behaviour-on-stats-finder

Demo: https://finder-frontend-pr-925.herokuapp.com/statistics

### DIFF
--- a/app/lib/filters/radio_filter.rb
+++ b/app/lib/filters/radio_filter.rb
@@ -9,9 +9,19 @@ module Filters
     def parsed_value
       return if params.blank?
 
-      JSON.parse params
-    rescue JSON::ParserError
-      params
+      if multi_value?
+        option_lookup.select { |key, _| params.include? key }.values.flatten
+      else
+        params
+      end
+    end
+
+    def multi_value?
+      facet.has_key?('option_lookup')
+    end
+
+    def option_lookup
+      @option_lookup ||= facet['option_lookup']
     end
   end
 end
--- a/app/lib/filters/radio_filter.rb
+++ b/app/lib/filters/radio_filter.rb
@@ -1,27 +1,7 @@
 module Filters
   class RadioFilter < Filter
     def value
-      Array(parsed_value)
-    end
-
-  private
-
-    def parsed_value
-      return if params.blank?
-
-      if multi_value?
-        option_lookup.select { |key, _| params.include? key }.values.flatten
-      else
-        params
-      end
-    end
-
-    def multi_value?
-      facet.has_key?('option_lookup')
-    end
-
-    def option_lookup
-      @option_lookup ||= facet['option_lookup']
+      parsed_value
     end
   end
 end
--- a/app/lib/filters/radio_filter.rb
+++ b/app/lib/filters/radio_filter.rb
@@ -1,0 +1,17 @@
+module Filters
+  class RadioFilter < Filter
+    def value
+      Array(parsed_value)
+    end
+
+  private
+
+    def parsed_value
+      return if params.blank?
+
+      JSON.parse params
+    rescue JSON::ParserError
+      params
+    end
+  end
+end
--- a/app/presenters/screen_reader_filter_description_presenter.rb
+++ b/app/presenters/screen_reader_filter_description_presenter.rb
@@ -5,7 +5,7 @@ class ScreenReaderFilterDescriptionPresenter
   end
 
   def present
-    text = (facets_without_tags + sort_options).compact.join(", ")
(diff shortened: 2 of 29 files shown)


app/lib/filters/radio_filter.rb
```diff
@@ -0,0 +1,27 @@
+module Filters
+  class RadioFilter < Filter
```
bilbof: The code here has been copied from TextFilter. Would it make sense to just extend `TextFilter` and not have anything here?

app/presenters/screen_reader_filter_description_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+    "<span class='visually-hidden'>#{text}</span>"
+  end
+
+private
+
+  attr_reader :filters, :sort_options
+
+  def facets_without_tags
+    description = filters.each_with_object([]) do |filter, facets_description|
+      if filter.hide_facet_tag?
```
bilbof: I'd suggest a `reject` before this. That way you don't need the `compact` call later.

app/presenters/screen_reader_filter_description_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+    ""
+  end
+
+  def facet_without_tag_default_option(filter)
+    default_option = filter.allowed_values
+                         &.detect { |option| option['default'] }
+    return '' if default_option.nil?
+
+    default_option.fetch('label', '')
```
bilbof: Perhaps we could do this, which is a bit clearer I think.

```
# return label of selected option, else return label of default option, else return nil
def filter_label(filter)
    selected_option = filter.allowed_values.find { |option| option['value'] == filter.value }
    return selected_option['label'] unless selected_option.nil?

    default_option = filter.allowed_values.find { |option| option['default'] }
    return default_option['label'] unless default_option.nil?

    nil
end

# in facets_without_tags
label = filter_label(filter)
facets_description << "#{filter.preposition} #{label}" unless label.nil?
```

It might even be worth adding the `filter_label` method to the facet model, what do you think? It's a bit confusing that the `filters` here are actually facets.

features/fixtures/statistics.json
```diff
@@ -102,41 +105,52 @@
...
         "filterable": true
       },
       {
         "key": "world_locations",
         "name": "World location",
         "preposition": "in",
-        "type": "autocomplete",
+        "type": "text",
         "display_as_result_metadata": true,
         "filterable": true
```
aliuk2012: could you add `"show_option_select_filter": true` to this file?

This key has been added to govuk-content-schema so it should not break anything and the functionality will be added as part of https://github.com/alphagov/finder-frontend/pull/933

features/fixtures/statistics.json
```diff
@@ -102,41 +105,52 @@
...
+      },
       {
         "key": "organisations",
         "name": "Organisation",
         "short_name": "Organisation",
         "preposition": "from",
-        "type": "autocomplete",
+        "type": "text",
         "display_as_result_metadata": true,
         "filterable": true
```
aliuk2012: could you add `"show_option_select_filter": true` to this file?

This key has been added to govuk-content-schema so it should not break anything and the functionality will be added as part of https://github.com/alphagov/finder-frontend/pull/933

app/presenters/screen_reader_filter_description_presenter.rb
oscarwyatt: I thought about adding it to the model but it felt a bit eggy, I think its six of one, AFAIK this logic only appears here so I'm minded to keep it here 🤷‍♂️

oscarwyatt:
> Just two minor things, if it could be added to this pr then it means we dont have to worry about updating this file later

Separate commit

features/fixtures/statistics.json
aliuk2012: @oscarwyatt You seemed to have missed off this second `"show_option_select_filter"` in your commit

PR #931, opened 2019-03-07
Change the default sorting after a keyword search to 'by relevance' in the business readiness finder

leenagupte:
Trello: https://trello.com/c/xyDkF16R

# What's changed

* When a user searches by keywords, the option to sort by Relevance is added to the Sort by drop-down list.
* When a user searches by keywords, the default order of the content switches to Sort by: Relevance.
* Pinned content does not appear at the top of results sorted by relevance
* Pinned content does not have special styling when results are sorted by relevance

# Why

Allowing users to sort by relevance on a keyword search means that we can start tagging content as best bets for a term thereby increasing their relevancy score in elastic search.  

Paired with @MahmudH

### DIFF
--- a/features/fixtures/business_readiness.json
+++ b/features/fixtures/business_readiness.json
@@ -85,6 +85,10 @@
         "key": "-popularity"
       },
       {
+        "name": "Relevance",
+        "key": "-relevance"
+      },
+      {
         "name": "Most recent",
         "key": "-public_timestamp"
       },
--- a/features/fixtures/business_readiness.json
+++ b/features/fixtures/business_readiness.json
@@ -33,6 +33,23 @@
         "web_url": "https://www.gov.uk/find-eu-exit-guidance-business/email-signup"
       }
     ],
+    "ordered_related_items": [
+      {
+        "api_path": "/api/content/guidance/regulations-and-standards-after-brexit",
+        "base_path": "/guidance/regulations-and-standards-after-brexit",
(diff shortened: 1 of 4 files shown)


vanitabarrett:
I've noticed that if you search (so the sort option changes to Relevance), then change the sort order (e.g: to Topic), and then try and go back to Relevance, it won't let you ('relevance' is greyed out). Looks like this also happens on News and Communications too though.

features/fixtures/business_readiness.json
```diff
@@ -33,6 +33,353 @@
         "web_url": "https://www.gov.uk/find-eu-exit-guidance-business/email-signup"
       }
     ],
+    "ordered_related_items": [
+      {
+        "api_path": "/api/content/guidance/the-food-and-drink-sector-and-preparing-for-eu-exit",
```
vanitabarrett: Do we need all the details here for the ordered related items? Would it work with just the minimum? Only because it makes the json file even longer and a bit difficult to read.

vanitabarrett:
I've left some comments but nothing blocking. Happy to approve 👍

features/fixtures/business_readiness.json
leenagupte: I'll shrink them down to just two.

leenagupte: @vanitabarrett I've amended and fixed up.

PR #933, opened 2019-03-07
Allow option_select component to show/hide filter field

aliuk2012:
By default the option_select component will not display the filter field.
The reason for this is that we wanted to control which option_select facet
allowed users to filter the checkboxes. There maybe a scenario where the
facet only has a few checkboxes and therefore it wouldn't make sense to
filter the options out (For example Policy finders doc type filter which
only has 3 options)

Ticket: https://trello.com/c/sEGGP7dF/470-turn-on-option-select-filters-for-finders
Demo:
- [x] **All Content**: https://finder-frontend-pr-933.herokuapp.com/all-content
- [x] **News & Comms**: https://finder-frontend-pr-933.herokuapp.com/news-and-communications (https://github.com/alphagov/rummager/pull/1458 needs to be merged)
- [x] **Services**: https://finder-frontend-pr-933.herokuapp.com/services
- [x] **Guidance & regulation**: https://finder-frontend-pr-933.herokuapp.com/guidance-and-regulation
- [x] **Transparency & FOI**: https://finder-frontend-pr-933.herokuapp.com/transparency-and-freedom-of-information-releases

### DIFF
--- a/spec/components/option_select_spec.rb
+++ b/spec/components/option_select_spec.rb
@@ -96,6 +96,15 @@ describe 'components/_option-select.html.erb', type: :view do
     expect(rendered).to have_selector('.app-c-option-select__count')
   end
 
+  it "does not show a filter control" do
+    arguments = option_select_arguments
+    arguments[:show_filter] = false
+    render_component(arguments)
+
+    expect(rendered).to have_no_selector('.app-c-option-select .gem-c-input[name="option-select-filter"]')
+    expect(rendered).to have_no_selector('.app-c-option-select__count')
+  end
+
   def expect_label_and_checked_checkbox(label, id, value)
     expect_label_and_checkbox(label, id, value, true)
   end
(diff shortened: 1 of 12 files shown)


spec/components/option_select_spec.rb
```diff
@@ -96,6 +96,16 @@ def option_select_with_tracking_arguments
     expect(rendered).to have_selector('.app-c-option-select__count')
   end
 
+  it "does not show a filter control" do
+    arguments = option_select_arguments
+    arguments[:show_filter] = false
+    render_component(arguments)
+
+    expect(rendered).to have_no_selector('.app-c-option-select .gem-c-input[name="option-select-filter"]')
+    print rendered
```
andysellick: Debug?

aliuk2012: Good spot

PR #935, opened 2019-03-08
fixes "disabled" Relevance sort option

aliuk2012:
By default "Relevance" sort option is disabled when JS has been disabled.
When the livesearch initialises it remembers the state of "Relevance" sort
option and then inserts the option back into the sort list in its original
state. This means that the option is disabled and users cannot reselect it
should they want to.
______________________
## Before

sort-bug-before

## After
sort-bug-after

_____________
Ticket: https://trello.com/c/H4P3VKtH/492-bugfix-relevance-sort-option-is-disabled-incorrectly
Demo: https://finder-frontend-pr-935.herokuapp.com/news-and-communications

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -217,7 +217,7 @@
 
   LiveSearch.prototype.insertRelevanceOption = function insertRelevanceOption() {
     var adjacentOption = this.$orderSelect.children("option").eq(this.$relevanceOrderOptionIndex);
-
+    this.$relevanceOrderOption.removeAttr('disabled');
     adjacentOption.before(this.$relevanceOrderOption);
   };
 
--- a/spec/javascripts/live_search_spec.js
+++ b/spec/javascripts/live_search_spec.js
@@ -34,10 +34,12 @@ describe("liveSearch", function(){
   };
 
   beforeEach(function () {
+    sortList = '<select id="order" class="js-order-results" data-relevance-sort-option="relevance"><option>Test 1</option><option value="relevance" disabled>Relevance</option>';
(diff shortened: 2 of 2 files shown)


vanitabarrett:
@aliuk2012 Should the 'Relevance' option be appearing on first page load (without a search)? It's disabled, but I'm wondering if it's clear to users why.

aliuk2012:
good question @vanitabarrett. I'm not 100% sure why it does what it does. @andysellick do you know the history of the sort?

bilbof:
I think that relevance isn't 'relevant' unless you enter keywords. From ES docs, with only sort/filters applied:

> Documents will be returned in effectively random order, and each document will have a score of zero

Behind the scenes in finder-frontend, selecting 'sort by relevance' omits a sort option from rummager see (app/lib/search_query_builder.rb:101), so the order will be 'effectively random'.

The reasoning behind not providing relevance as an option is that it wouldn't have any meaningful effect (it wouldn't make things more relevant).

Selecting 'show relevant results first' with no keywords entered might even confuse users as the documents would be sorted in random order.

But if the UX of disappearing/reappearing sort options is confusing then I don't see a problem with removing the special logic around this option, as it would make things simpler :) Another option would be to display a note explaining this when no keywords are entered and the user selects 'relevance'. What do you think?

andysellick:
Yeah, we could just make the option not be present unless you've entered keywords. If that's simpler.

andysellick:
I've just noticed that if you add a keyword then remove a keyword then the 'relevance' sort option is removed, so yeah, might be good to make this consistent by only having 'relevance' present in the sort options if a keyword is entered.

PR #937, opened 2019-03-11
Have toggle option for facets on mobile only if facets are present

andysellick:
@sihugh noticed that on https://www.gov.uk/prepare-eu-exit/going-and-being-abroad the 'show more search options' link is visible on mobile, even though there are no more search options. This change fixes that.

Screen Shot 2019-03-11 at 10 40 54

Example where there this link should be present: /news-and-communications

Example where it shouldn't be present: /prepare-eu-exit/going-and-being-abroad


aliuk2012:
Could you write a test for this?

PR #938, opened 2019-03-11
Rename select facet to option select facet

andysellick:
We have a facet called a select facet, which is actually an option select facet. 

A `select` is a HTML element that allows users to choose one option from a drop down list. It looks like this:

Screen Shot 2019-03-11 at 15 19 55

An [option select](https://finder-frontend.herokuapp.com/component-guide/option-select) is a custom built GOV.UK component that looks like this:

Screen Shot 2019-03-11 at 15 20 58

This is a bit confusing. This is especially confusing since we also have a facet called a dropdown select facet, which is actually a select.

This PR renames the select facet to the option select facet in order to make it clearer what it actually is.


bilbof:
Technically this change looks OK. Perhaps you could define what an option select is, and why it is different from a select?

PR #939, opened 2019-03-11
display abbreviation alongside any facet that has an abbreviation

Seraphiana:
this changes how any individual facet entry displays: 
if a value has an abbreviation, it will be displayed as {value} ({abbr})
otherwise it will be as before.

https://trello.com/c/Ql3rrz9O/374-allow-users-to-filter-orgs-using-abbreviations


Seraphiana:
some examples of slightly sketchy data:
District Valuer Services (DVS) (DVS)
Civil Service (the civil service)

PR #940, opened 2019-03-12
[WIP] Add styling for top result

vanitabarrett:
Trello: https://trello.com/c/rzhhJI54/252-add-styling-to-high-relevance-content-on-business-readiness-finder-and-internal-search

To see what the styling will look like, add the 'document--top' class to a result list item.

Screen Shot 2019-03-12 at 10 16 42

**To-Do**
- Create a function that only shows this styling on the top, most relevant result per search (actual logic to be defined).
- Make sure this only applies to the business finder, not other finders
- Get the description to display for top items only.
- Make sure the top item is only shown when finder is sorted by Relevance.

### DIFF
--- a/app/presenters/result_set_presenter.rb
+++ b/app/presenters/result_set_presenter.rb
@@ -49,10 +49,15 @@ class ResultSetPresenter
     }.reject(&:empty?)
   end
 
+  def is_top_result?(result)
+    ## Logic here that compares all the results and decides if a result is the top one.
+    true
+  end
+
   def documents
     results.each_with_index.map do |result, index|
       {
-        document: SearchResultPresenter.new(result).to_hash,
+        document: SearchResultPresenter.new(result, is_top_result?(result)).to_hash,
         document_index: index + 1
       }
     end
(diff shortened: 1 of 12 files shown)


app/presenters/result_set_presenter.rb
```diff
@@ -49,13 +49,25 @@ def selected_filter_descriptions
     }.reject(&:empty?)
   end
 
+  def is_top_result?(score_a, score_b)
+    (score_a / score_b) > 1
```
DilwoarH: We need to speak to Sam Dub about the scoring rule

app/views/finders/_document.mustache
```diff
@@ -1,4 +1,5 @@
-<li class="document">
+<li class="document{{#document.top_result}} document--top{{/document.top_result}}">
+  <p>{{#document.top_result}}Most relevant result{{/document.top_result}}</p>
```
DilwoarH:
```suggestion
  {{#document.top_result}}<p>Most relevant result</p>{{/document.top_result}}
```

DilwoarH: The reason why I am suggesting this is because if there is no top results, it will add an empty `<p>` tag

MahmudH:
Please refer to https://github.com/alphagov/finder-frontend/pull/974

PR #943, opened 2019-03-12
Adds in routes for upcoming and published stats

SamJamCul:
Adds additional redirect routes from the old statistics finder, converting the upcoming and published pages into equivalent radio button selections. There are some tests that will currently fail though, so this should not be merged yet!

Relates to https://trello.com/c/XxjGvJ63/493-handle-redirects-from-the-stats-finder

### DIFF
--- a/app/controllers/redirection_controller.rb
+++ b/app/controllers/redirection_controller.rb
@@ -1,28 +1,41 @@
 class RedirectionController < ApplicationController
   def announcements
-    parameters = { keywords: params['keywords'],
-                  level_one_taxon: params['taxons'].try(:first),
-                  level_two_taxon: params['subtaxons'].try(:first),
-                  people: params['people'],
-                  organisations: params['departments'],
-                  world_locations: params['world_locations'],
-                  public_timestamp: { from: params['from_date'], to: params['to_date'] }.compact.presence }
     respond_to do |format|
-      format.html { redirect_to(finder_path('search/news-and-communications', params: parameters)) }
-      format.atom { redirect_to(finder_path('search/news-and-communications', params: parameters, format: :atom)) }
+      format.html { redirect_to(finder_path('search/news-and-communications', params: convert_common_parameters)) }
+      format.atom { redirect_to(finder_path('search/news-and-communications', params: convert_common_parameters, format: :atom)) }
     end
   end
 
   def publications
-    parameters = { keywords: params['keywords'],
-                  level_one_taxon: params['taxons'].try(:first),
-                  level_two_taxon: params['subtaxons'].try(:first),
-                  organisations: params['departments'],
-                  world_locations: params['world_locations'],
-                  public_timestamp: { from: params['from_date'], to: params['to_date'] }.compact.presence }
     respond_to do |format|
-      format.html { redirect_to(finder_path('search/all', params: parameters)) }
-      format.atom { redirect_to(finder_path('search/all', params: parameters, format: :atom)) }
+      format.html { redirect_to(finder_path('search/all', params: convert_common_parameters)) }
+      format.atom { redirect_to(finder_path('search/all', params: convert_common_parameters, format: :atom)) }
     end
+  end
+
+  def published_statistics
+    respond_to do |format|
+      format.html { redirect_to(finder_path('search/statistics', params: convert_common_parameters)) }
+      format.atom { redirect_to(finder_path('search/statistics', params: convert_common_parameters, format: :atom)) }
+    end
+  end
+
+  def upcoming_statistics
+    respond_to do |format|
+      format.html { redirect_to(finder_path('search/statistics', params: convert_common_parameters.merge(content_store_document_type: :statistics_upcoming))) }
+      format.atom { redirect_to(finder_path('search/statistics', params: convert_common_parameters.merge(content_store_document_type: :statistics_upcoming), format: :atom)) }
+    end
+  end
+
+private
+
+  def convert_common_parameters
+    { keywords: params['keywords'],
+      level_one_taxon: params['taxons'].try(:first) || params['topics'].try(:first),
+      level_two_taxon: params['subtaxons'].try(:first),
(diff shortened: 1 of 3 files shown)


SamJamCul:
> Looks good! `published_statistics` and `upcoming_statistics` has some duplicated code, so if you can find a way to reuse that stuff that would be nice.

Yeah will do, I think there's actually quite a lot of duplication that I can get rid of for all the redirection controller functions. 

The failing tests are currently intended, as the statistics finder hasn't been published yet - once it's out, I'll move this from a draft PR to a proper one

app/controllers/redirection_controller.rb
```diff
@@ -25,4 +25,27 @@ def publications
...
+      format.atom { redirect_to(finder_path('search/statistics', params: parameters, format: :atom)) }
+    end
+  end
+
+  def upcoming_statistics
+    parameters = { keywords: params['keywords'],
+                  level_one_taxon: params['topics'].try(:first),
+                  organisations: params['organisations'],
+                  content_store_document_type: :statistics_upcoming,
+                  public_timestamp: { from: params['from_date'], to: params['to_date'] }.compact.presence }
```
oscarwyatt: As this logic is nearly identical to `published_statistics` I'd suggest moving the common logic to a separate method eg 
```
def convert_statistics_parameters
{ keywords: params['keywords'],
                  level_one_taxon: params['taxons'].try(:first),
                  organisations: params['departments'],
                  public_timestamp: { from: params['from_date'], to: params['to_date']
}.compact.presence }
end
```
and then doing something like this in upcoming statistics
```
parameters = convert_statistics_parameters.merge({ content_store_document_type: :statistics_upcoming})
```

which keeps things a little neater

spec/controllers/redirection_controller_spec.rb
```diff
@@ -62,4 +62,57 @@
       expect(response).to redirect_to finder_path('search/all', format: :atom, params: { keywords: %w[one two] })
     end
   end
+
+  describe '#published_statistics' do
```
oscarwyatt: This sparks joy

oscarwyatt:
Sterling work 👍 just one code thing and one git tidiness thing -> unless there is a good reason to have 8 commits, I'd suggest squashing them into one

app/controllers/redirection_controller.rb
SamJamCul: Alright, I've kind of ended up implementing the opposite of this, where every other method except upcoming_statistics uses a common_parameters method to handle stuff they're all using. This is because the upcoming stats page uses it's own special parameter names for some reason, calling things 'topics' rather than 'taxons' and 'departments' rather than 'organisations'.

But the main efficiency is there, and it'll be much easier to add more methods now it doesn't have to get retyped every time. 

I'll find out from someone how to make it all one big commit.

app/controllers/redirection_controller.rb
```diff
@@ -1,28 +1,53 @@
...
       format.html { redirect_to(finder_path('search/all', params: parameters)) }
       format.atom { redirect_to(finder_path('search/all', params: parameters, format: :atom)) }
     end
   end
+
+  def published_statistics
+    parameters = convert_common_parameters
+    respond_to do |format|
+      format.html { redirect_to(finder_path('search/statistics', params: parameters)) }
+      format.atom { redirect_to(finder_path('search/statistics', params: parameters, format: :atom)) }
```
bilbof: You can just use `convert_common_parameters` here, then you don't need to create a new variable 'parameters'.

app/controllers/redirection_controller.rb
```diff
@@ -1,28 +1,53 @@
...
   def publications
-    parameters = { keywords: params['keywords'],
-                  level_one_taxon: params['taxons'].try(:first),
-                  level_two_taxon: params['subtaxons'].try(:first),
-                  organisations: params['departments'],
-                  world_locations: params['world_locations'],
-                  public_timestamp: { from: params['from_date'], to: params['to_date'] }.compact.presence }
+    parameters = convert_common_parameters.merge(
+      level_two_taxon: params['subtaxons'].try(:first),
+      world_locations: params['world_locations']
```
bilbof: I would recommend moving `people`, `level_two_taxon` and `world_locations` into the `convert_common_parameters` method.

If param keys like `world_locations` are unlikely to be used more in different ways, even if they are not used by all finders there's no harm in putting `world_locations` into the translated common params method at the bottom. You can always call `.compact` on the hash to remove keys with no values.

app/controllers/redirection_controller.rb
bilbof: What you could do is also move this stuff to the params method. Where you have multiple keys being used, like organisations / departments both being the same thing, you could do this:

```
organisations: params['organisations'] || params['departments'],
```

app/controllers/redirection_controller.rb
SamJamCul: Hey brill, I didn't know about .compact :D

SamJamCul: Alrighty, refactored so it's super condensed now

app/controllers/redirection_controller.rb
```diff
@@ -1,28 +1,41 @@
...
+private
+
+  def convert_common_parameters
+    { keywords: params['keywords'],
+      level_one_taxon: params['taxons'].try(:first) || params['topics'].try(:first),
+      level_two_taxon: params['subtaxons'].try(:first),
+      organisations: params['departments'] || params['organisations'],
+      people: params['people'],
+      world_locations: params['world_locations'],
+      public_timestamp: { from: params['from_date'], to: params['to_date'] }.compact.presence }
```
bilbof: You'll need to call `.compact` on the final `}` otherwise you'll have empty keys being passed in.

SamJamCul: Cool, made the change now

bilbof: :shipit:

PR #944, opened 2019-03-12
Map values for rummager

peteglondon:
# What
* Map the query sent to rummager to `filter_facet_values` queries when facets are selected.
* Map the `facet_values` array returned by rummager back to metadata used by `finder_frontend`.
* Filter by `filter_facet_groups` to replace `appear_in_find_eu_exit_guidance_business_finder`.
* Handle the new finder structure where `details -> facets` is replaced by `links -> facet_group -> links -> facets -> links -> facet_values`
* Add a cucumber scenario for the business finder Q&A and make Q&A compatible with the new finder structure.
* Update the fixtures for all business finder scenarios to use the new finder structure. 
# Why
We need to map the facets the user selects in the finder to the query Rummager expects Eg.

```
https://www.gov.uk/search?filter_facet_values=42ce66de-04f3-4192-bf31-8394538e0734
https://www.gov.uk/search?filter_facet_groups=42ce66de-04f3-4192-bf31-8394538e0734
```

We also need to map the response we'll get back from Rummager into a format that matches how finders currently work (mapping the metadata into the format used to group by and render the metadata).

Trello: https://trello.com/c/Sd7omEFH/263-mapping-rummager-request-response-in-finder-frontend


peteglondon:
[Mapping Rummager request/response in Finder Frontend](https://trello.com/c/Sd7omEFH/263-mapping-rummager-request-response-in-finder-frontend)

app/lib/facets_from_facet_group_extractor.rb
```diff
@@ -0,0 +1,36 @@
+class FacetsFromFacetGroupExtractor
+  def initialize(content_item)
+    @facets = content_item['links']['facet_group'].first['links']['facets']
```
steventux: I know we test for the existence of facet links in the calling `FacetExtractor` class but this line makes me wonder whether the two are correctly coupled. This constructor makes an assumption about the data it's passed, wouldn't it be better to let the class initialisation handle missing data.

Even a more guarded
```ruby
@facets = content_item['links'].fetch('facet_group', []).first.fetch('links', {})['facets']
```
might be less brittle.

Also I think when we're at the point where we're chaining so much data lookup there's an argument for handling only the data we need, is it possible to pass only `content_item['links']['facet_group']` to the constructor?

PR #946, opened 2019-03-13
[Do not merge] Option 1: Fiddle tags for business finder

leenagupte:
Trello: https://trello.com/c/0rDl6oiK

## Motivation

The way email signups work for the business readiness finder is
changing. This is because the finder works differently to other finders.
Rather than narrowing the results, the finder expands them, doing an
"or" query rather than an "and". That is not how email-alert-api works.
It creates an "and" query from the facets it's passed.

This means that most users subscribed to the business readiness finders
were not receiving updates.

## What's changed

The content-item is being updated in a related commit to remove `email_filter_facets`, so users won't be forced to select a facet to subscribe to.

We need to pass
```
{ "appear_in_find_eu_exit_guidance_business_finder" => { "any" => %W(yes) } }
```
in the tags for the business readiness finder to create the subscription.

Ideally we would add `appear_in_find_eu_exit_guidance_business_finder` to the email-signup content item, however we tried many combinations of adding the tag and nothing worked.

### What didn't work

``` 
email_filter_by: appear_in_find_eu_exit_guidance_business_finder
```
(both with and without the subscription_title_prefix whitespace)
This returned `all-government-publishing-updates` as the subscription
topic_id

```
filter: "appear_in_find_eu_exit_guidance_business_finder"
```
(both with and without the subscription_title_prefix whitespace)
We had to add it to the content schemas. This also returned
`all-government-publishing-updates` as the subscription topic_id

```
facet_id: appear_in_find_eu_exit_guidance_business_finder
filter_key: "appear_in_find_eu_exit_guidance_business_finder"
filter_value: "yes"
facet_name: appear_in_find_eu_exit_guidance_business_finder
```
(both with and without the subscription_title_prefix whitespace)
This also returned `all-government-publishing-updates` as the
subscription topic_id

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -63,6 +63,7 @@ private
       default_attributes: { filter: default_filters, reject: default_rejects },
       subscription_list_title_prefix: content['details']['subscription_list_title_prefix'],
       available_choices: signup_presenter.choices,
+      signup_content_id: signup_content_id,
     )
   end
 
@@ -88,5 +89,9 @@ private
 
   def default_rejects
     content['details'].fetch('reject', {})
+  end
+
+  def signup_content_id
+    content['content_id']
   end
 end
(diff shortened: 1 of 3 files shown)


leenagupte:
Superseded by PR #954

PR #947, opened 2019-03-13
Converts q parameter into keywords to work with Google Chrome

oscarwyatt:
This means Google Chrome users will be able to
tab into a search of GOV.UK and our new finders
will be able to understand this.

Trello: https://trello.com/c/hWcX9Feo/502-setup-q-as-a-synonym-for-keywords-in-finders-m


bilbof:
Some failing integration tests, but otherwise looks good!

PR #949, opened 2019-03-13
Change Google Chrome config to search with new finder

oscarwyatt:
This tells Google Chrome that if users tab when typing 'gov.uk' to
use the new finder and the correct parameter

Trello: https://trello.com/c/Rr6BrjXi/453-configure-chrome-address-bar-search-to-know-where-our-site-search-is-s

Relies on #947 - DO NOT MERGE until that is deployed

### DIFF
--- a/app/views/search/opensearch.xml
+++ b/app/views/search/opensearch.xml
@@ -2,5 +2,5 @@
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/" xmlns:moz="http://www.mozilla.org/2006/browser/search/">
     <ShortName>GOV.UK</ShortName>
     <Description>Search GOV.UK - The best place to find government services and information.</Description>
-    <Url type="text/html" template="https://www.gov.uk/search?q={searchTerms}"/>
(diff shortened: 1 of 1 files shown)


app/views/search/opensearch.xml
```diff
@@ -2,5 +2,5 @@
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/" xmlns:moz="http://www.mozilla.org/2006/browser/search/">
     <ShortName>GOV.UK</ShortName>
     <Description>Search GOV.UK - The best place to find government services and information.</Description>
-    <Url type="text/html" template="https://www.gov.uk/search?q={searchTerms}"/>
+    <Url type="text/html" template="https://www.gov.uk/search/all?keywords={searchTerms}"/>
```
bilbof: Should this be `all-content`?

bilbof: Actually, no you're right - it turns out I missed something in the spec for https://trello.com/c/SjudMnoT/499-update-urls-across-all-7-finders 😄 thanks!

PR #954, opened 2019-03-14
Pass business readiness finder tag as hidden field

leenagupte:
Trello: https://trello.com/c/0rDl6oiK

Supersedes: PR #946

## Motivation

The way email signups work for the business readiness finder is
changing. This is because the finder works differently to other finders.
Rather than narrowing the results, the finder expands them, doing an
"or" query rather than an "and". That is not how email-alert-api works.
It creates an "and" query from the facets it's passed.

This means that most users subscribed to the business readiness finders
were not receiving updates.

## What's changed

The assumption is that "appear_in_find_eu_exit_guidance_business_finder" will be added to the email-signup content item:

```
    "email_filter_by": "appear_in_find_eu_exit_guidance_business_finder",
    "email_filter_name": "appear_in_find_eu_exit_guidance_business_finder",
    "email_signup_choice": [
      {
        "key": "yes",
        "radio_button_name": "yes",
        "prechecked": true
      }
    ],
```

As "email_signup_choice" has to be populated to get the value of "yes", the checkbox form will be rendered.  Additional code has been be added to hide the checkbox field, but still pass in the facet values to EmailAlertSignupAPI.

### DIFF
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -28,6 +28,20 @@ class SignupPresenter
     choices? && choices_formatted.any?
   end
 
+  def hidden_choices
+    hidden_choices = choices.map do |choice|
+      if ignore_facet?(choice["facet_id"])
+        choice['facet_choices'].map do |facet_choice|
+          {
+            name: "filter[#{choice['facet_id']}][]",
+            value: facet_choice["key"],
+          }
+        end
+      end
+    end
+    hidden_choices.flatten
+  end
+
   def choices?
     multiple_facet_choice_data.present? || single_facet_choice_data[0]["facet_choices"].present?
   end
@@ -96,5 +110,9 @@ private
     return nil unless email_filter_name
 
     (email_filter_name["plural"] || email_filter_name).capitalize
+  end
+
+  def ignore_facet?(facet_id)
+    %W(appear_in_find_eu_exit_guidance_business_finder).include?(facet_id)
   end
 end
--- a/app/views/email_alert_subscriptions/new.html.erb
+++ b/app/views/email_alert_subscriptions/new.html.erb
@@ -10,8 +10,15 @@
 <div class="outer-block">
   <div class="signup-content">
     <%= form_tag email_alert_subscriptions_path(subscriber_list_params: subscriber_list_params), class: 'signup-choices form' do %>
-
-      <% if @signup_presenter.can_modify_choices? %>
+      <% if @signup_presenter.hidden_choices.any? %>
+        <%= render partial: 'govuk_publishing_components/components/title', locals: {
+          title: @signup_presenter.name,
+          context: "Email alert subscription"
+        } %>
+        <% @signup_presenter.hidden_choices.each do |hidden_choice| %>
+          <%= hidden_field_tag hidden_choice[:name], hidden_choice[:value] %>
+        <% end %>
+      <% elsif @signup_presenter.can_modify_choices? %>
         <%= render "govuk_publishing_components/components/checkboxes", {
           name: nil,
           heading: @signup_presenter.name,
--- a/features/finders.feature
+++ b/features/finders.feature
@@ -234,10 +234,12 @@ Feature: Filtering documents
     When I use a checkbox filter and another disallowed filter
     Then I can sign up to email alerts for allowed filters
 
+  @javascript
   Scenario: Subscribing to email alerts for business readiness finder
     When I view the business readiness finder
     And I create an email subscription
-    Then I see the email subscription page  
+    Then I see the email subscription page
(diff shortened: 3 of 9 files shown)


app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -11,7 +11,15 @@
...
     <%= form_tag email_alert_subscriptions_path(subscriber_list_params: subscriber_list_params), class: 'signup-choices form' do %>
 
-      <% if @signup_presenter.can_modify_choices? %>
+      <% if @signup_presenter.hidden_choices.any? %>
+        <%= render partial: 'govuk_publishing_components/components/title', locals: {
+          title: @signup_presenter.name,
+          context: "Email alert subscription"
+        } %>
+        <% @signup_presenter.hidden_choices do |choice| %>
+          <input type="hidden" id="#{<%=choice[:id]%>}" name=="#{<%=choice[:name]%>}", value="#{<%=choice[:value]%>}">
```
steventux: This can be done with a tag helper:
```ruby
<%= hidden_field_tag choice[:name], choice[:value], { id: choice[:id] } %>
```

vanitabarrett:
@leenagupte I've made the following changes to fix the logic. I deployed to integration and it seems to work!

- Pass `facet_id` instead of `facet_name` through to `ignore_facet?`
- Instead of trying to render a hidden input, just wrap a checkbox in the `js-hidden` class.
- Updated the email signup content item so it doesn't refer to news and communication (I assume we'll have to move this file into govuk-app-deployment-secrets).

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -10,23 +10,28 @@
...
+        </div>
       <% elsif @signup_presenter.can_modify_choices? %>
         <%= render "govuk_publishing_components/components/checkboxes", {
-          name: nil,
-          heading: @signup_presenter.name,
-          hint_text: "Select the email alerts you need.",
-          items: @signup_presenter.choices_formatted,
-          is_page_heading: true,
-        } %>
+              name: nil,
```
steventux: Is this indentation intentional?

leenagupte:
@vanitabarrett Thanks for sorting this out! Yes we'll need to add the updated content item to govuk-app-deployment-secrets, and update the developer docs. I'm just going to complete the feature we started and remove the rogue `puts page.body` that I left in there.

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -10,8 +10,21 @@
...
     <%= form_tag email_alert_subscriptions_path(subscriber_list_params: subscriber_list_params), class: 'signup-choices form' do %>
-
-      <% if @signup_presenter.can_modify_choices? %>
+      <% if @signup_presenter.hidden_choices.any? %>
+        <%= render partial: 'govuk_publishing_components/components/title', locals: {
+          title: @signup_presenter.name,
+          context: "Email alert subscription"
+        } %>
+        <div class="js-hidden">
+          <%= render "govuk_publishing_components/components/checkboxes", {
```
steventux: Is there a reason why this isn't a hidden field? Using `js-hidden` to hide the elements will show 2 unintuitive checkboxes when js is disabled. Also there's no point in constructing the `hidden_choices` hash(es) if we aren't using them with these checkboxes.
I think we can make this a hidden field with
```ruby
<% hidden_choices.each do |hidden_choice| %>
  <%= hidden_field_tag hidden_choice[:name], hidden_choice[:value] %>
<% end %>
```

app/presenters/signup_presenter.rb
```diff
@@ -28,6 +28,18 @@ def can_modify_choices?
     choices? && choices_formatted.any?
   end
 
+  def hidden_choices
+    choices.select do |choice|
```
steventux: This method is calling select but the [return type of the block isn't a boolean](https://apidock.com/ruby/Hash/select) which probably isn't intended.
It's also only used to detect a hidden choice, the hash data we populate here is never used.
I suggest we map the choices into a name and value pair we can use as hidden field attribute values.
ie

```ruby
def hidden_choices
  choices.map do |choice|
    if ignore_facet?(choice["facet_id"])
      choice['facet_choices'].map do |facet_choice|
        {
          name: "filter[#{choice['facet_id']}][]",
          value: facet_choice["key"],
        }
      end
    end
  end.flatten
end
```

This should work with

```ruby
<% hidden_choices.each do |hidden_choice| %>
  <%= hidden_field_tag hidden_choice[:name], hidden_choice[:value] %>
<% end %>
```

as described below.

leenagupte:
@steventux `hidden_field_tag` is a much neater solution. Thanks. I've made the change and fixed up.

PR #958, opened 2019-03-14
Statistics finder email signup page

bilbof:
https://trello.com/c/mMPcqr2x/509-default-finder-radio-button-should-always-be-applied
https://trello.com/c/5oag3R25/288-publish-statistics-finder-from-rummager-s

This is the final step to enable us to publish the new stats finder.

Changes:
- Adds email signup page for testing
- Bugfix: the Radio facet & filter will now use a provided default value when unset
- Bugfix: the query strings for email alerts and atom feeds will be correct on first page load.

Reviewing by commit will be easier - see commit messages.

### DIFF
--- a/app/lib/filters/radio_filter.rb
+++ b/app/lib/filters/radio_filter.rb
@@ -1,7 +1,36 @@
 module Filters
   class RadioFilter < Filter
     def value
-      parsed_value
+      return default_value unless acceptable_param?
+
+      return option_lookup_values(params) if multi_value?
+
+      Array(params)
+    end
+
+  private
+
+    def default_value
+      return option_lookup_values(default_allowed_value) if multi_value?
+
+      Array(default_allowed_value)
+    end
+
+    def option_lookup_values(val)
+      option_lookup.select { |key, _| val.include? key }.values.flatten
+    end
+
+    def default_allowed_value
+      @default_allowed_value ||= facet['allowed_values'].find(Proc.new { {} }) { |option| option['default'] }
+      @default_allowed_value['value']
+    end
+
+    def acceptable_param?
+      params.present? && params.is_a?(String) && param_is_part_of_allowed_values
+    end
+
+    def param_is_part_of_allowed_values
+      facet['allowed_values'].find { |option| option['value'] == params }.present?
     end
   end
 end
--- a/app/lib/filters/radio_filter.rb
+++ b/app/lib/filters/radio_filter.rb
@@ -11,6 +11,8 @@ module Filters
   private
 
     def default_value
+      return [] if default_allowed_value.blank?
+
       return option_lookup_values(default_allowed_value) if multi_value?
 
       Array(default_allowed_value)
--- a/app/lib/filters/radio_filter.rb
+++ b/app/lib/filters/radio_filter.rb
@@ -30,7 +30,7 @@ module Filters
     end
 
     def param_is_part_of_allowed_values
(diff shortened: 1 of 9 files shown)


spec/lib/filters/radio_filter_spec.rb
```diff
@@ -43,39 +94,73 @@
...
+
+      context "when no option is selected" do
+        it "should return the corresponding default values from the option_lookup" do
+          expect(radio_filter.value).to eq(%w(guidance))
+        end
+      end
+
+      context "when a disallowed value is provided" do
+        let(:params) { "does_not_exist" }
+        it "should return the corresponding default values from the option_lookup" do
```
karlbaker02: Is it worth again here splitting out into a default value being provided and not being provided as you have above?

karlbaker02:
Is it also worth adding an explicit test for a radio with previously selected value that is then reset to `nil` if for example it is set to `nil` via the facet (is this actually a behaviour that the radio facet should/will support)?

app/models/radio_facet.rb
```diff
@@ -1,4 +1,8 @@
 class RadioFacet < FilterableFacet
+  def value
+    @value || default_value['value']
```
karlbaker02: Where is `@value` defined? There's a `value` on the `FilterableFacet` base class, but I don't think it's this?

app/lib/filters/radio_filter.rb
```diff
@@ -1,7 +1,36 @@
...
+
+      Array(default_allowed_value)
+    end
+
+    def option_lookup_values(val)
+      option_lookup.select { |key, _| val.include? key }.values.flatten
+    end
+
+    def default_allowed_value
+      @default_allowed_value ||= facet['allowed_values'].find(Proc.new { {} }) { |option| option['default'] }
```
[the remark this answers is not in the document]
bilbof: It sets a default return value of `{}` rather than the default `nil`. This is instead of doing `|| {}` afterwards. https://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-find

app/lib/filters/radio_filter.rb
```diff
@@ -1,7 +1,36 @@
...
+      @default_allowed_value['value']
+    end
+
+    def acceptable_param?
+      params.present? && params.is_a?(String) && param_is_part_of_allowed_values
+    end
+
+    def param_is_part_of_allowed_values
+      facet['allowed_values'].find { |option| option['value'] == params }.present?
```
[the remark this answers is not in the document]
bilbof: Yeah we can, thanks!

app/models/radio_facet.rb
bilbof: It's quite hard to see. It's set within `FacetCollection` here:

https://github.com/alphagov/finder-frontend/blob/master/app/models/facet_collection.rb#L15

This gets called in the `FinderPresenter`:

https://github.com/alphagov/finder-frontend/blob/master/app/presenters/finder_presenter.rb#L16

This is how values from the params are translated into their corresponding facets.

bilbof:
Currently there is no case where a user can reset the radio button to nil. I don't think that the facet needs to support this yet, though it may support it now.

spec/lib/filters/radio_filter_spec.rb
bilbof: Yes good idea. And actually, this unearthed a bug - see the last commit. It was unexpected behaviour for there to be no default value and an option_lookup. Thanks for suggesting this.

PR #959, opened 2019-03-14
Use release_timestamp for atom feed entries

bilbof:
If public_timestamp is not available we will try to use release_timestamp. This is only useful for upcoming stats which don't have a public timestamp, but do have a release timestamp.

https://trello.com/c/8D4Y1ye5/522-fix-upcoming-stats-atom-feed

View the change at: `/search/statistics?parent=&keywords=&level_one_taxon=&content_store_document_type=statistics_upcoming&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&order=updated-newest`

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,6 +1,7 @@
 class Document
   attr_reader :title, :public_timestamp, :is_historic, :government_name,
-              :content_purpose_supergroup, :document_type, :organisations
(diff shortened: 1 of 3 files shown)


bilbof:
I've added a do not merge label as I'm awaiting approval from a product perspective; it's possible we don't want email/atom feeds for upcoming statistics.

PR #960, opened 2019-03-15
Toggle all content facets

andysellick:
Adds a toggle link to the all content finder so that all facets (apart from keyword search) are hidden by default. Other finders should not be affected.

Screen Shot 2019-03-15 at 08 16 25

This is a little confusing as the behaviour on mobile for all finders should be to hide the facets like this. Initial page load states should be as follows:

Finder | Element | Desktop state | Mobile state | JS disabled state
-------|---------|---------------|--------------|-----------------
All content | Facet toggle link | Show | Show | Hide
All content | Facets | Hide | Hide | Show
Others | Facet toggle link | Hide | Show | Hide
Others | Facets | Show | Hide | Show

Trello card: https://trello.com/c/Fy8KPoWV/501-all-content-show-hide-facets?menu=filter&filter=label:Finders

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -379,23 +379,34 @@
 
 .facet-toggle {
   display: none;
+  margin-bottom: $gutter;
 }
 
 .js-enabled {
   .facet-toggle {
     display: inline-block;
-    margin-bottom: $gutter;
+  }
 
-    @include media(tablet) {
+  .facet-toggle__content {
+    display: block;
+
+    &.facet-toggle__content--hide {
       display: none;
     }
   }
 
-  .facet-toggle__content--hidden {
-    display: none;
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -379,23 +379,34 @@
...
+        display: none;
+      }
+    }
 
-    @include media(tablet) {
-      display: block;
+    .facet-toggle__content.facet-toggle__content--hide {
+      @include media(tablet) {
+        display: block;
+      }
```
aliuk2012: Do we need so much nesting? its making it really hard to read and understand what each class is doing. Perhaps we should group the media queries inside the class names so we don't have a lot of the same class name dotted around the file.

We could also use govuk design systems media query mixins (https://design-system.service.gov.uk/get-started/updating-your-code/#media-queries)

PR #964, opened 2019-03-18
Refactor FinderApi

bilbof:
This is a proposed refactor of `FinderApi`, a class that fetches and provides data for the finder presenters, views, controllers, etc. Reviewing by commit will be easier.

Changes:

- adds a ContentItem model, responsible for fetching and providing content items (logic moved from FinderApi)
- Refactor and moves augmentation code from `FinderApi` to the presenters. Manipulating a content item hash can be hard to debug, this should make things simpler.

It doesn't go the whole way and remove the `allowed_options` code and other content item manipulation from FinderApi and AdvancedSearchFinderApi. I hope it will make that easier, if we do want to do that.

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -23,6 +23,10 @@ private
     @content ||= fetch_content_item(request.path)
   end
 
+  def finder_content_item
+    @finder_content_item ||= fetch_content_item(finder_base_path)
+  end
+
   def signup_presenter
     @signup_presenter ||= SignupPresenter.new(content, params)
   end
@@ -66,14 +70,8 @@ private
     )
   end
 
-  def finder
-    @finder ||= FinderPresenter.new(fetch_content_item(finder_base_path))
-  end
-
   def finder_format
-    return nil unless finder.filter
-
-    finder.filter['document_type']
+    finder_content_item.dig('details', 'filter', 'document_type')
   end
 
   def email_signup_attributes
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -53,7 +53,7 @@ private
   end
 
   def fetch_content_item(content_item_path)
-    FinderApi.new(content_item_path, {}).content_item
+    ContentItem.new(content_item_path).as_hash
   end
 
   def email_alert_signup_api
--- a/app/lib/advanced_search_finder_api.rb
+++ b/app/lib/advanced_search_finder_api.rb
@@ -5,8 +5,9 @@ class AdvancedSearchFinderApi < FinderApi
     raise_on_missing_taxon_param
 
     filter_params[TAXON_SEARCH_FILTER] = taxon["content_id"] if taxon
-    content_item_with_results = augment_content_item_with_results
-    augment_content_item_links_with_taxon(content_item_with_results, taxon)
+    augment_content_item_with_results
+    augment_content_item_links_with_taxon(content_item, taxon)
+    content_item
   end
 
   def taxon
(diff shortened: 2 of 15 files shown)


app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -66,14 +70,11 @@ def email_alert_signup_api
     )
   end
 
-  def finder
-    @finder ||= FinderPresenter.new(fetch_content_item(finder_base_path))
-  end
-
   def finder_format
-    return nil unless finder.filter
+    filter = finder_content_item['details']['filter']
```
koetsier: could 'dig' work here?, i.e. ```finder_content_item.dig('details', 'filter', 'document_type')```

app/presenters/finder_presenter.rb
```diff
@@ -58,7 +59,16 @@ def summary
   end
 
   def pagination
-    content_item['details']['pagination']
+    documents_per_page = content_item['details']['default_documents_per_page']
+    start_offset = search_results['start']
```
koetsier: could put a guard here instead: ```return nil unless documens_per_page``` - saves having to get ```start_offset``` and ```total_results```

app/lib/advanced_search_finder_api.rb
```diff
@@ -5,12 +5,8 @@ def content_item_with_search_results
     raise_on_missing_taxon_param
 
     filter_params[TAXON_SEARCH_FILTER] = taxon["content_id"] if taxon
-    search_response = fetch_search_response(content_item)
-    content_item_with_taxon_links = augment_content_item_links_with_taxon(
-      content_item,
-      taxon
-    )
-    augment_content_item_with_results(content_item_with_taxon_links, search_response)
+    content_item_with_results = augment_content_item_with_results
```
koetsier: Maybe either call this ```augmented_content_item``` and don't bother with the assignment, just use it directly in the statement below.

On the other hand I like explicitly including the arguments so it is clear what this methods dependencies are

app/lib/advanced_search_finder_api.rb
```diff
@@ -26,13 +22,20 @@ def augment_content_item_links_with_taxon(content_item, taxon)
     content_item
   end
 
-  def augment_facets_with_dynamic_values(content_item, _search_response)
-    augment_facets_with_dynamic_subgroups(content_item) if supergroups.any?
+  def augment_content_item_with_results
+    item_hash = content_item
```
koetsier: see previous comment

app/lib/advanced_search_finder_api.rb
```diff
@@ -26,13 +22,20 @@ def augment_content_item_links_with_taxon(content_item, taxon)
...
+    item_hash['details']['results'] = search_results.fetch("results")
+    augment_facets_with_dynamic_values(item_hash)
+    item_hash
```
koetsier: this mutates``` @content_item``` and may complicate matters

app/lib/finder_api.rb
```diff
@@ -72,7 +65,7 @@ def relevance_scores_exist?(results)
     results.all? { |result| result['es_score'].present? }
   end
 
-  def fetch_search_response(content_item)
+  def fetch_search_response
```
koetsier: maybe keep the dependency on content_item - I'd be tempted to add filter_params as well

app/lib/finder_api.rb
```diff
@@ -85,28 +78,15 @@ def fetch_search_response(content_item)
     end
   end
 
-  def augment_content_item_with_results(content_item, search_response)
-    content_item = augment_content_item_details_with_results(content_item, search_response)
-    augment_facets_with_dynamic_values(content_item, search_response)
-    content_item
+  def augment_content_item_with_results
+    item_hash = content_item
```
koetsier: see comments above

koetsier:
Couple of comments. Great step in the right direction! 👍

app/lib/finder_api.rb
```diff
@@ -85,28 +78,15 @@ def fetch_search_response(content_item)
...
-      search_response.fetch('total')
-    )
-
-    content_item
-  end
-
-  def augment_facets_with_dynamic_values(content_item, search_response)
-    search_response.fetch("facets", {}).each do |facet_key, facet_details|
-      facet = content_item['details']['facets'].find { |f| f['key'] == facet_key }
+  def augment_facets_with_dynamic_values(item)
```
koetsier: ```item``` is a bit of a non-descript name. Maybe content_item_hash?

bilbof:
Thanks @koetsier, I amended the final commit to include the changes you recommended.

PR #968, opened 2019-03-18
Fix multiple Analytics events for checkbox facets

oscarwyatt:
They were firing three times per click, which would skew our analysis.
This prevents live-search and the built in checkbox component events
firing, thus allowing the TrackClick module in static to take over.
(This is defined in static/app/assets/javascripts/modules/track-click.js)

Trello: https://trello.com/c/7vQb8Yww/434-bug-fix-issue-where-use-of-brexit-checkbox-fires-multiple-ga-events

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -33,7 +33,8 @@
 
       this.$form.on('change', 'input[type=checkbox], input[type=radio], select',
         function(e) {
-          if (!e.suppressAnalytics) {
+          // Checkboxes fire their own GA tracking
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -33,7 +33,8 @@
 
       this.$form.on('change', 'input[type=checkbox], input[type=radio], select',
         function(e) {
-          if (!e.suppressAnalytics) {
+          // Checkboxes fire their own GA tracking
+          if (!e.suppressAnalytics && $(e.target).attr("type") !== "checkbox") {
```
aliuk2012: Could we maybe add a test for this bug to ensure it never happens again?

PR #971, opened 2019-03-19
Remove organisation metadata for mainstream content

koetsier:
Organisation metadata should not be shown for mainstream
content.

Trello: https://trello.com/c/GlohunbM/519-remove-org-metadata-from-mainstream-results-in-all-content-and-services-finders

image

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -47,6 +47,20 @@ private
 
   attr_reader :link, :rummager_document, :finder, :description
 
+  def is_mainstream_content?
+    %w(completed_transaction
+       local_transaction
+       calculator
+       smart_answer
+       simple_smart_answer
+       place
+       licence
+       step_by_step
+       transaction
+       answer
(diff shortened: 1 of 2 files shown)


app/models/document.rb
```diff
@@ -47,6 +47,20 @@ def promoted
...
+       local_transaction
+       calculator
+       smart_answer
+       simple_smart_answer
+       place
+       licence
+       step by step
+       transaction
+       answer
+       guide).include?(@document_type)
```
bilbof: Is there anywhere this is stored centrally? Like Govuk Document Types?

koetsier: couldn't find it anywhere myself

bilbof: Should step by step be underscored?

koetsier: you're right. serves me for copy-pasting from trello! good spot

PR #972, opened 2019-03-19
Change stats fixture URL

bilbof:
https://trello.com/c/VMPLPT1X/538-change-stats-finder-url


app/models/content_item.rb
```diff
@@ -40,10 +40,11 @@ def fetch_content_item
   # without filetype as the value; example:
   # "/guidance-and-regulation" => "guidance_and_regulation"
   FINDERS_IN_DEVELOPMENT = {
+    "/search/all" => "all_content",
```
sihugh: Al's updated these to use `/review-search/....` in a current PR. We _really_ don't want them overriding the live ones if we can help it

sihugh: It was also causing his tests to fail because it was using the FINDERS_IN_DEVELOPMENT instead of the webmock fixtures

PR #973, opened 2019-03-19
Adds in routes for upcoming and published stats

SamJamCul:
Small fix needed to make sure parameters aren't passed over if empty! I somehow managed to merge the version that didn't have that final .compact :/


SamJamCul:
> 👍

Thanks, didn't get an email about you doing this :/

PR #974, opened 2019-03-19
Add top result in documents with styling

MahmudH:
When the business finder is visited, and a keyword search is made, resulting in order of most relevance, we check if the top result is a best bet. This comparison is made between the first and second document, with the difference in es_score. If the first document's es_score is much larger than the second document's es_score, then top_result of true is added to the first document. When this occurs, the document has special styling for prominence. 

Before:
Screen Shot 2019-03-19 at 16 04 02

After:
Screen Shot 2019-03-19 at 16 10 41

Trello: https://trello.com/c/rzhhJI54


app/presenters/result_set_presenter.rb
```diff
@@ -78,6 +90,22 @@ def signup_links
 
   attr_reader :view_context
 
+  def no_top_results(documents)
```
leenagupte: The Rails convention for methods that return true/false is add a `?` to the end of the name. 

Also, I think for readability, this function should have a "truthy" name, i.e. `highlight_top_result?`.
It would make the code easier to read too:

e.g. 
```
documents.length > 2 && 
sort_option && 
sort_option["key"].eq?("-relevance") &&
finder.slug == "/find-eu-exit-guidance-business"
```

app/presenters/result_set_presenter.rb
```diff
@@ -49,13 +49,25 @@ def selected_filter_descriptions
     }.reject(&:empty?)
   end
 
+  def is_top_result?(score_a, score_b)
```
leenagupte: I think this needs to be renamed too because the first results is always the "top" results when ordered by `-relevance`

app/models/document.rb
```diff
@@ -16,6 +16,7 @@ def initialize(rummager_document, finder)
     @is_historic = rummager_document.fetch(:is_historic, false)
     @government_name = rummager_document.fetch(:government_name, nil)
     @es_score = rummager_document.fetch(:es_score, nil)
+    @top_result = rummager_document.fetch(:top_result, false)
```
leenagupte: Is `top_result` returned by Rummager?

app/models/document.rb
```diff
@@ -1,7 +1,7 @@
 class Document
   attr_reader :title, :public_timestamp, :is_historic, :government_name,
               :content_purpose_supergroup, :document_type, :organisations,
-              :release_timestamp, :es_score
+              :release_timestamp, :es_score, :top_result, :description
```
leenagupte: Why has description been made public?

app/presenters/result_set_presenter.rb
```diff
@@ -78,6 +90,22 @@ def signup_links
...
+    documents.length < 2 ||
+      !sort_option ||
+      !sort_option["key"].eql?("-relevance") ||
+      finder.slug != "/find-eu-exit-guidance-business"
+  end
+
+  def with_top_result(documents)
+    if (documents[0][:document][:es_score].present? && documents[1][:document][:es_score].present?)
+      documents[0][:document][:top_result] = is_top_result?(documents[0][:document][:es_score], documents[1][:document][:es_score])
+      documents[0][:document][:summary] = results[0].description if documents[0][:document][:top_result]
```
leenagupte: Could this logic be applied inside the `map` in the documents method?

spec/models/document_spec.rb
```diff
@@ -129,4 +130,24 @@
       expect(instance.es_score).to eq 0.005
     end
   end
+
+  describe "top_result" do
```
leenagupte: I think this test needs to be moved to `result_set_presenter_spec.rb`.

app/views/finders/_document.mustache
```diff
@@ -1,4 +1,5 @@
-<li class="document">
+<li class="document{{#document.top_result}} document--top{{/document.top_result}}">
+    {{#document.top_result}}<p>Most relevant result</p>{{/document.top_result}}
```
DilwoarH: Sam said this should be removed

app/presenters/search_result_presenter.rb
```diff
@@ -25,6 +25,7 @@ def to_hash
       promoted_summary: promoted_summary,
       show_metadata: show_metadata,
       es_score: es_score,
+      top_result: false
```
leenagupte: Why is this needed?

app/presenters/result_set_presenter.rb
```diff
@@ -98,7 +98,7 @@ def no_top_results(documents)
   end
 
   def with_top_result(documents)
-    if (documents[0][:document][:es_score].present? && documents[1][:document][:es_score].present?)
+    if documents[0][:document][:es_score].present? && documents[1][:document][:es_score].present?
```
leenagupte: Can you fix up this lint error to the previous commit?

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -393,20 +393,18 @@
...
         promoted_summary: nil,
         show_metadata: false,
-        es_score: 0.1,
         es_score: 900.0,
         )
     end
 
     before(:each) do
-      allow(presenter_with_top_result).to receive(:sort_option).and_return({ "key" => "-relevance" })
+      allow(presenter_with_top_result).to receive(:sort_option).and_return("key" => "-relevance")
```
leenagupte: Ditto this lint error, can you fix it up?

features/finders.feature
```diff
@@ -220,6 +220,13 @@ Feature: Filtering documents
     Then I see most relevant order selected
     And I do not see results with pinned items
 
+  @javascript
+  Scenario: Adding top result in business finder
+    When I view the business readiness finder
+    And I fill in some keywords
+    Then I see most relevant order selected
+    And I see results with top result
```
leenagupte: Could you add another step that tests tracking to this test?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -66,6 +61,18 @@
   }
 }
 
+.document--top {
+  border: 1px solid $grey-3;
```
aliuk2012: Could you replace `$grey-3` with `govuk-colour("grey-3")` we are trying to move more towards govuk-frontend and move away for govuk-elements https://design-system.service.gov.uk/styles/colour/

features/fixtures/business_readiness_filtered_results.json
```diff
@@ -58,7 +58,7 @@
             "other-eu"
           ],
           "index": "detailed",
-          "es_score": null,
+          "es_score": 10000.0,
```
leenagupte: Do both fixture files need to have es_scores? For some searches (A-Z etc) the es_score is always null.

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -66,6 +61,18 @@
...
+  padding: govuk-spacing(6);
+  margin-bottom: govuk-spacing(3);
+
+  .document__link {
+    @include bold-24;
```
aliuk2012: could you replace this with `@include govuk-font(24, $weight: bold)`

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -66,6 +61,18 @@
...
+    display: block;
+    margin-top: $gutter-one-third;
```
aliuk2012: $gutter, only use for the gaps in between grid columns

Try and use GOVUK spacing or responsive spacing https://design-system.service.gov.uk/styles/spacing/

app/models/document.rb
MahmudH: We need to add the description to the top result, therefore needed it publicly available to use in ResultSetPresenter.

app/presenters/result_set_presenter.rb
```diff
@@ -78,6 +90,22 @@ def signup_links
 
   attr_reader :view_context
 
+  def no_top_results(documents)
+    documents.length < 2 ||
+      !sort_option ||
+      !sort_option["key"].eql?("-relevance") ||
+      finder.slug != "/find-eu-exit-guidance-business"
```
leenagupte: It's possible that we might want to roll this pattern out to other finders in the (near) future, so perhaps we should separate out the slug check to either another method, or check if the finder slug is in an allowed list.

app/models/document.rb
leenagupte: Maybe add that to the commit message?

app/models/document.rb
MahmudH: No it isn't so I will change this 👍

app/views/finders/_document.mustache
DilwoarH: Sam wants to keep this in now :) lool

features/fixtures/business_readiness_filtered_results.json
MahmudH: I have removed es_score out of business_readiness_results.json

app/presenters/result_set_presenter.rb
```diff
@@ -16,6 +16,7 @@ def initialize(finder, filter_params, view_context)
     @total = finder.results.total
     @filter_params = filter_params
     @view_context = view_context
+    @top_result_finder_whitelist = ["/find-eu-exit-guidance-business"]
```
leenagupte: You could use a global variable here as the value never changes.

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -335,6 +335,94 @@
     end
   end
 
+  describe "#top_results?" do
```
leenagupte: There isn't a method in ResultSetPresenter called `top_results?` so this should be renamed.

app/models/document.rb
steventux: I think finder frontend exposes `summary` which is usually the underlying document description.
https://github.com/alphagov/finder-frontend/blob/05a65874c6633787f487cebdaa7af8ad7d71bad3/app/models/document.rb#L35

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -335,6 +335,94 @@
...
+    subject(:presenter_with_top_result) { ResultSetPresenter.new(finder, filter_params, view_context) }
+
+    let(:filter_params) { { keywords: 'test', order: '-relevance' } }
+    let(:results) do
```
leenagupte: Do you need this? `results` is being reassigned on line 411, so I don't think this is being used.

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -335,6 +335,94 @@
...
+    let(:document_with_lower_es_score) do
+      double(
+        Document,
+        title: 'Investigation into the distribution of road fuels in parts of Scotland',
+        path: 'slug-2',
+        metadata: [
+          { id: 'case-state', name: 'Case state', value: 'Open', type: 'text', labels: %W(open) },
+          { id: 'opened-date', name: 'Opened date', value: '2006-7-14', type: 'date' },
+          { id: 'case-type', name: 'Case type', value: 'CA98 and civil cartels', type: 'text', labels: %W(ca98-and-civil-cartels) },
+          { id: 'case-type', name: 'Case type', value: 'CA98 and civil cartels', type: 'text', labels: %W(ca98-and-civil-cartels) }
```
leenagupte: `case-type` has been defined twice.

app/presenters/result_set_presenter.rb
steventux: We really shouldn't pepper finder-frontend with too many paths like this (usually the same eu-exit finder path).
There's a method on finder which (I added here https://github.com/alphagov/finder-frontend/blob/master/app/presenters/finder_presenter.rb#L345) which also contains this path. As @leenagupte says we should define this as a constant once somewhere where it can referenced easily in many methods/places.

spec/presenters/result_set_presenter_spec.rb
leenagupte: Is all of this metadata need for the test?

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -335,6 +335,94 @@
...
+
+    context "has top result true" do
+      let(:results) do
+        ResultSet.new(
+          [document_with_higher_es_score, document_with_lower_es_score],
+          total
+        )
+      end
+
+      it 'creates a new document for each result' do
```
leenagupte: I think this should be renamed. Something like `sets top_result to true if...`.

Could you also add a test that checks `top_result` isn't set?

app/presenters/result_set_presenter.rb
```diff
@@ -49,10 +50,27 @@ def selected_filter_descriptions
...
+
+  def get_documents_with_top_result(documents = [])
+    with_top_result(documents) if highlight_top_result?(documents)
+    documents
+  end
+
   def documents
     results.each_with_index.map do |result, index|
+      doc = SearchResultPresenter.new(result).to_hash
+      if highlight_top_result?(results) && index === 0 && result.es_score && results[index+1].es_score
```
leenagupte: Should the check work the other other way around? i.e. `if index === 0` do all the things?

app/presenters/result_set_presenter.rb
```diff
@@ -63,8 +63,8 @@ def get_documents_with_top_result(documents = [])
   def documents
     results.each_with_index.map do |result, index|
       doc = SearchResultPresenter.new(result).to_hash
-      if highlight_top_result?(results) && index === 0 && result.es_score && results[index+1].es_score
-        if is_best_bet?(result.es_score, results[index+1].es_score)
+      if highlight_top_result?(results) && index === 0 && result.es_score && results[index + 1].es_score
+        if is_best_bet?(result.es_score, results[index + 1].es_score)
```
leenagupte: Can you fix up both of these lint errors to where they first occurred?

features/fixtures/business_readiness_filtered_results.json
leenagupte: I think you should leave the score as `null` rather than taking it out completely as that's what the data from rummager looks like.

app/presenters/result_set_presenter.rb
leenagupte: I agree with @steventux. Also FinderPresenter is being passed to ResultSetPresenter as `finder` so if we make the `eu_exit_finder` method public, ResultSetPresenter will be able to access it:

https://github.com/alphagov/finder-frontend/blob/master/app/controllers/finders_controller.rb#L50

app/presenters/result_set_presenter.rb
```diff
@@ -49,10 +50,27 @@ def selected_filter_descriptions
     }.reject(&:empty?)
   end
 
+  def is_best_bet?(score_a, score_b)
+    # We found the average score on the top 500 searches and found that 7 was the most suitable number
+    (score_a / score_b) > 7
+  end
+
+  def get_documents_with_top_result(documents = [])
```
leenagupte: This method is being called by anything, so it can be removed.

app/presenters/result_set_presenter.rb
```diff
@@ -76,7 +94,18 @@ def signup_links
...
+  attr_reader :view_context, :top_result_finder_whitelist
+
+  def highlight_top_result?(documents)
+    documents.length >= 2 &&
+      sort_option &&
+      sort_option["key"].eql?("-relevance") &&
+      whitelist_finder_for_top_result(finder)
+  end
+
+  def whitelist_finder_for_top_result(finder)
```
leenagupte: You don't need to pass in `finder`, it can already be accessed.

app/presenters/result_set_presenter.rb
leenagupte: `results` can be read from anywhere in the class, so it doesn't need to be passed in.

app/presenters/result_set_presenter.rb
```diff
@@ -49,10 +50,27 @@ def selected_filter_descriptions
     }.reject(&:empty?)
   end
 
+  def is_best_bet?(score_a, score_b)
```
leenagupte: This method can also access `results` so we don't need to pass anything to it.

The checks for whether  an `es_score` exists could be moved here as well. Something like:

```
score_a = results[0].es_score
score_b = results[1].es_score

# We found the average score on the top 500 searches and found that 7 was the most suitable number
if score_a.present? && score_b.present?
  (score_a / score_b) > 7
end
```

In Ruby convention you don't need the `is_` in the method name as that's covered by the `?`.
Also, this method should be private.

app/presenters/result_set_presenter.rb
leenagupte: I think by moving most of the checks to `highlight_top_result?` you could make this block much easier to read.

The `documents` method could just check `index === 0 && highlight_top_result?`

Then `highlight_top_result?` could do all of the work:

```
    whitelist_finder_for_top_result? &&
      results.length >= 2 &&
      is_best_bet? &&
      sort_option.present? &&
      sort_option["key"].eql?("-relevance")
```

app/models/document.rb
```diff
@@ -1,7 +1,7 @@
 class Document
   attr_reader :title, :public_timestamp, :is_historic, :government_name,
               :content_purpose_supergroup, :document_type, :organisations,
-              :release_timestamp, :es_score
+              :release_timestamp, :es_score, :description
```
leenagupte: As @steventux mentioned in another comment, I don't think you need to expose `description` as `summary` is already available.

app/presenters/result_set_presenter.rb
```diff
@@ -76,7 +81,22 @@ def signup_links
 
 private
 
-  attr_reader :view_context
+  attr_reader :view_context, :top_result_finder_whitelist
```
leenagupte: I don't think `top_result_finder_whitelist` exists anymore?

spec/models/document_spec.rb
```diff
@@ -208,6 +208,8 @@
           value: 'other-value-1'
         )
       end
+    end
+  end
```
leenagupte: What happened here? Can you fix it up?

app/models/document.rb
MahmudH: summary always comes back as null, so have to get the description from the document and add it to top result

spec/models/document_spec.rb
MahmudH: Not sure why it wasn't there at the beginning but it is needed to end the tests

app/presenters/result_set_presenter.rb
MahmudH: Removed 👍

app/models/document.rb
MahmudH: The summary always comes back with null so I had to use the document description

app/models/document.rb
leenagupte: That's because `show_summaries` isn't set in the content_item. However, I think that if we set that to true, the summary will show for every list item.

The only concern I have is that `summary` returns a truncated description. We might want to do the same thing.

app/presenters/result_set_presenter.rb
```diff
@@ -51,8 +51,13 @@ def selected_filter_descriptions
 
   def documents
     results.each_with_index.map do |result, index|
+      doc = SearchResultPresenter.new(result).to_hash
+      if  index === 0 && highlight_top_result?
+        doc[:top_result] = true
+        doc[:summary] = results[0].description
```
leenagupte: I think this can just be `result.description`

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -66,6 +61,18 @@
   }
 }
 
+.document--top {
+  border: 1px solid govuk-colour('grey-2');
+  padding: govuk-spacing(6);
+  margin-bottom: govuk-spacing(3);
+
+  .document__link {
```
aliuk2012: Does this have to be nested inside `.document--top`?  One of the benefits of using BEM is not having to be so specific have have nested classes.

app/assets/javascripts/live_search.js
```diff
@@ -150,6 +150,16 @@
         });
       });
     }
+
+    var $results = $('.finder-results');
+    if ($results.length > 0) {
+      if ($results.find('.document--top .document__link').length == 1) {
```
aliuk2012: if you refactor the `.document__link` in the scss file then you could refactor this to `.find('.document__link--top')`

aliuk2012: `$results.find('.document--top .document__link')` seems to be used quite a few times. Could I suggest you cache the elements  eg `var $topResultsLink = $results.find('.document__link--top')`

app/assets/javascripts/live_search.js
```diff
@@ -150,6 +150,16 @@
...
+        var trackingAction = $results.find('.document--top .document__link').attr('data-track-action');
+        trackingAction += 'r';
+        $results.find('.document--top .document__link').attr('data-track-action', trackingAction);
```
aliuk2012: if you refactor the `.document__link` in the scss file then you could refactor this to `.find('.document__link--top')` or better yet use the cached results

app/assets/javascripts/live_search.js
```diff
@@ -150,6 +150,16 @@
         });
       });
     }
+
+    var $results = $('.finder-results');
+    if ($results.length > 0) {
+      if ($results.find('.document--top .document__link').length == 1) {
+        var trackingAction = $results.find('.document--top .document__link').attr('data-track-action');
```
aliuk2012: Use the cached lookup

app/presenters/finder_presenter.rb
```diff
@@ -278,6 +278,12 @@ def all_content_finder?
     self.content_item['content_id'] == 'dd395436-9b40-41f3-8157-740a453ac972'
   end
 
+  # FIXME: This should be removed once we have a way to determine
+  # whether to display metadata in the finder definition
+  def eu_exit_finder
+    slug == "/find-eu-exit-guidance-business"
```
aliuk2012: Would it not be better to use the `content_id` instead of the slug?  This will also match how `all_content_finder?` works. Could I also suggest adding `?` to the end of the method name to follow ruby conventions?

aliuk2012: If we going to make this method public could we also write a test for it?

aliuk2012:
Looking good from frontend POV. Only two issues relating to the eu_exit_finder 
 Presenter method.

spec/models/document_spec.rb
```diff
@@ -222,4 +224,31 @@
...
+      truncated_description = "The government has many departments."
+
+      subject(:instance_with_description) { described_class.new({ title: "Y", link: "/y", description: description }, finder) }
+      subject(:instance_without_description) { described_class.new({ title: "Y", link: "/y" }, finder) }
+
+      it 'should have truncated description' do
+        expect(instance_with_description.truncated_description).to eq(truncated_description)
+      end
+
+      it 'should have truncated description' do
```
leenagupte: "not" is missing from the test name.

app/presenters/finder_presenter.rb
```diff
@@ -278,6 +278,12 @@ def all_content_finder?
     self.content_item['content_id'] == 'dd395436-9b40-41f3-8157-740a453ac972'
   end
 
+  # FIXME: This should be removed once we have a way to determine
+  # whether to display metadata in the finder definition
+  def eu_exit_finder
```
leenagupte: As per @aliuk2012's suggestion it would be nice to add a `?` to the end of the method name so that it follows Rails conventions.

app/models/document.rb
```diff
@@ -47,9 +47,14 @@ def promoted
...
 
+  def truncated_description
+    # This truncates the description at the end of the first sentence
+    description.gsub(/\.\s[A-Z].*/, '.') if description.present?
+  end
+
 private
 
-  attr_reader :link, :rummager_document, :finder, :description, :facet_content_ids
+  attr_reader :link, :rummager_document, :finder, :facet_content_ids, :description
```
leenagupte: If you use a tool like [Fork](https://git-fork.com/) it's possible to discard changes, rather than trying to revert them manually.

spec/presenters/finder_presenter_spec.rb
```diff
@@ -322,6 +322,17 @@ def sort_option(label, value, disabled: false, selected: false)
     end
   end
 
+  describe "#eu_exit_finder" do
```
leenagupte: Typo: The test name is missing a `?`

PR #976, opened 2019-03-20
All-Content: Update page heading to "Search"

aliuk2012:
The All Content finder should have the keywords field at the top of the page instead of in the facets sidebar. All other finders should still have the keywords search field in their facets sidebar

Example:
image

____

Ticket: https://trello.com/c/LznDjL1r/505-all-content-update-page-heading-to-search
Demo:
All Content: https://finder-frontend-pr-976.herokuapp.com/review-search/all
News & Comms: https://finder-frontend-pr-976.herokuapp.com/search/news-and-communications

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -410,3 +410,12 @@
   }
 }
 
+.finder-frontend-content {
+  .app-c-search-page-heading {
+    @include govuk-font(48, $weight: bold);
+    @include govuk-responsive-margin(2, "top");
+    @include govuk-responsive-margin(4, "bottom");
+    color: govuk-colour("black");
+  }
+}
+
--- a/app/views/finders/_facet_collection.html.erb
+++ b/app/views/finders/_facet_collection.html.erb
@@ -1,13 +1,18 @@
-<div class="filter-form" id="keywords">
-  <%= render "govuk_publishing_components/components/input", {
-    label: {
-      text: "Search"
-    },
-    name: "keywords",
-    value: @results.user_supplied_keywords,
-    id: "finder-keyword-search",
-    controls: "js-search-results-info"
-  } if finder.show_keyword_search? %>
+<div class="filter-form">
+  <% if finder.show_keyword_search? and !finder.all_content_finder? %>
+    <div id="keywords">
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: "Search"
+        },
+        name: "keywords",
+        value: @results.user_supplied_keywords,
+        id: "finder-keyword-search",
+        controls: "js-search-results-info"
+      }  %>
+    </div>
+  <% end %>
+
   <div class="govuk-form-group" data-module="track-click" data-track-category="filterClicked"
       data-track-action="skip-Link" data-track-label="">
     <%= render "govuk_publishing_components/components/skip_link", {
--- a/app/views/finders/show.html.erb
+++ b/app/views/finders/show.html.erb
@@ -21,43 +21,62 @@
   <%= render 'govuk_publishing_components/components/government_navigation', active: finder.government_content_section %>
 <% end %>
 
-<header>
-  <div class="title-and-metadata">
-    <%= render partial: 'govuk_publishing_components/components/title', locals: {
-      context: finder.human_readable_finder_format,
-      title: finder.name
-    } %>
-    <% if finder.page_metadata.any? %>
-      <%= render 'govuk_publishing_components/components/metadata', page_metadata(finder.page_metadata) %>
-    <% end %>
+<form method="get" action="<%= finder.slug %>" class="js-live-search-form">
+  <input type="hidden" name="parent" value="<%= @parent %>">
 
-    <% if finder.summary %>
-      <div class="summary">
-        <%= render 'govuk_publishing_components/components/govspeak', content: finder.summary.html_safe %>
+  <header>
+    <div class="title-and-metadata">
+      <% if finder.all_content_finder? %>
+        <% label_text = capture do %>
+          <h1 class="app-c-search-page-heading">Search<span class="govuk-visually-hidden"> all content</span></h1>
+        <% end %>
+        <div id="keywords">
+          <%= render "govuk_publishing_components/components/search", {
+              inline_label: false,
+              label_text: label_text,
+              name: "keywords",
+              hide_search_button: true,
+              value: @results.user_supplied_keywords,
+              id: "finder-keyword-search",
+              controls: "js-search-results-info"
+          } %>
+        </div>
+      <% else %>
+        <%= render partial: 'govuk_publishing_components/components/title', locals: {
+            context: finder.human_readable_finder_format,
+            title: finder.name
+        } %>
+      <% end %>
+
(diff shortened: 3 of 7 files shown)


app/views/finders/_facet_collection.html.erb
```diff
@@ -7,7 +7,7 @@
     value: @results.user_supplied_keywords,
     id: "finder-keyword-search",
     controls: "js-search-results-info"
-  } if finder.show_keyword_search? %>
+  } if finder.show_keyword_search? and finder.name != 'All content' %>
   <div class="govuk-form-group" data-module="track-click" data-track-category="filterClicked"
```
andysellick: Nitpick: can we get a newline in here? I keep working on this file and briefly thinking the `if` on line 10 as relating to the code below.

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -410,3 +410,12 @@
   }
 }
 
+.finder-frontend-content {
+  .app-c-search-page-heading {
```
andysellick: Ideally we shouldn't have glue code like this for components but I can see why it was necessary. I think we should raise a separate PR to add some styling to the component to handle this.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -46,7 +46,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     function getInput(removeFilterName, removeFilterValue, removeFilterFacet) {
       var selector = (!!removeFilterName) ? " input[name='" + removeFilterName + "']" : " [value='" + removeFilterValue + "']";
 
-      return $('#' + removeFilterFacet).find(selector);
+      return $(selector);
```
andysellick: Can you explain this change?

app/views/finders/show.html.erb
```diff
@@ -21,75 +21,95 @@
...
+          <% end %>
+          <%= render "govuk_publishing_components/components/search", {
+              inline_label: false,
+              label_text: label_text,
+              name: "keywords",
+              hide_search_button: true,
+              value: @results.user_supplied_keywords,
+              id: "finder-keyword-search",
+              controls: "js-search-results-info"
+          } %>
```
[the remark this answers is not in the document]
aliuk2012: I totally agree, we should extend the search component and it would also help with not having to add the extra styling added in this PR

app/views/finders/show.html.erb
```diff
@@ -21,75 +21,95 @@
...
+      <div class="logo">
+        <%= image_tag finder.logo_path, id: "logo-image" %>
+      </div>
     <% end %>
 
-    <% if finder.summary %>
-      <div class="summary">
-        <%= render 'govuk_publishing_components/components/govspeak', content: finder.summary.html_safe %>
+    <% if finder.related.any? %>
+      <div class='related-links'>
```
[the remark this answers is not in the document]
aliuk2012: Good catch!!  I didn't spot it but it looks like this has been hanging around since Aug 2014 😱 (86b5a772dfabf6027f8afbd451cc170d796af905)

app/assets/stylesheets/finder_frontend.scss
aliuk2012: I agree, I've mention before we should extend the search component to allow it to be used as a page heading and that change would mean that we could remove this code.

app/assets/javascripts/modules/remove-filter.js
aliuk2012: It just works! 😛 

I went back to try remember what the issue was but discovered an existing issue with "keywords". I remembered that there seemed to be an issue with removing keywords using the facet labels and the reason was that it could not find the div with id "keywords".

I've fixed undone this change and fixed the bigger issue of keywords div wrapping the whole left sidebar. Keywords div now only wraps the search field. See my last commit

aliuk2012:
@vanitabarrett would you mind giving this a review? @andysellick has had to go off but he has done an initial review and I've addressed all his issues. We need to get this out as soon as possible.

app/assets/stylesheets/finder_frontend.scss
vanitabarrett: Is there a card or something that can be linked to from here? So people who come across this comment can see if this change to the search component has been made yet or not.

app/views/finders/show.html.erb
```diff
@@ -56,7 +72,9 @@
 </header>
 
 <div class="grid-row">
-  <form method="get" action="<%= finder.slug %>" class="js-live-search-form">
+  <% if finder.name != 'All content' %>
```
vanitabarrett: `if finder.name != 'All content'` appears in quite a few places. Can we define it as a function in one place, e.g: the finder presenter?

app/views/finders/_facet_collection.html.erb
```diff
@@ -7,7 +7,8 @@
     value: @results.user_supplied_keywords,
     id: "finder-keyword-search",
     controls: "js-search-results-info"
-  } if finder.show_keyword_search? %>
+  } if finder.show_keyword_search? and finder.name != 'All content' %>
```
vanitabarrett: `finder.name != 'All content'` isn't an ideal check, as someone could change the name in the finder config, not realising it would break this logic. Ideally this would be set somewhere in the finder config, but we could also check against the content id as this is less likely to change.

app/views/finders/show.html.erb
aliuk2012: Good idea, I'll see what I can do

app/views/finders/_facet_collection.html.erb
aliuk2012: Thanks for suggestion the content id, that seems like a good alternative. As you say anyone could change the name.

aliuk2012:
@vanitabarrett Would you mind reviewing my last commit? It should contain a new presenter method which uses the `content_id`.

Thanks for flagging the issue with the 2nd commit. My plan is to squash 1st and 2nd commit. The local contents slugs uses a different one from production '/review-search/` so it shouldn't have conflicts or issues or it being released into production and causing confusion. I think the 3rd and 4th commit should still be separate.

spec/presenters/finder_presenter_spec.rb
```diff
@@ -248,6 +248,17 @@ def sort_option(label, value, disabled: false, selected: false)
     end
   end
 
+  describe "#all_content_finder?" do
+    fit 'returns true if finder is not "All Content"' do
```
vanitabarrett: This should say "returns true if finder is All Content"

aliuk2012:
🤔 I might just squash the lot into single commit

spec/presenters/finder_presenter_spec.rb
```diff
@@ -248,6 +248,17 @@ def sort_option(label, value, disabled: false, selected: false)
     end
   end
 
+  describe "#all_content_finder?" do
+    fit 'returns true if finder is "All Content"' do
```
andysellick: What does `fit` do?

aliuk2012: Its a Fit bit (https://www.fitbit.com/) but only for specs :trollface: 

it only used to run specific 'focused' tests.

Good spot.

PR #984, opened 2019-03-20
Fix facet tracking events being logged more than once

aliuk2012:
## This issue covered two different scenarios. 

1. All form elements were double logging tracking events.
2. Brexit checkbox facet was logging two events of its own.

## Solution 1:

There was condition on live search which always evaluated to true. The method then went on to call the tracking method in addition to the original tracking event. To solve this issue, I removed the condition and the function call.

## Solution 2:

**TL;DR - govuk_publishing_components and Static's Module.TrackClick were responsible for the double logging.**

Checkbox facets were adding `data-module="track-click"` html attribute. This is a hook for Static's Module.TrackClick.

govuk_publishing_component's checkbox component was also firing off its own event. So to resolve this issue I created this PR https://github.com/alphagov/govuk_publishing_components/pull/801 which uses the events `suppressAnalytics` flag to stop tracking event being called. 

Ticket: https://trello.com/c/7vQb8Yww/434-bug-fix-issue-where-use-of-brexit-checkbox-fires-multiple-ga-events

**Demo:** 
https://finder-frontend-pr-984.herokuapp.com/search/all
https://finder-frontend-pr-984.herokuapp.com/search/news-and-communications
https://finder-frontend-pr-984.herokuapp.com/search/research-and-statistics

### DIFF
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -5,6 +5,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict';
 
   GOVUK.Modules.RemoveFilter = function () {
+    var onChangeSuppressAnalytics = {
+      type: "change",
+      suppressAnalytics: true
+    }
+
     this.start = function (element) {
       $(element).on('click', '[data-module="remove-filter-link"]', toggleFilter);
     };
@@ -19,8 +24,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var removeFilterFacet = $el.data('facet');
 
       var $input = getInput(removeFilterName, removeFilterValue, removeFilterFacet);
-      clearFacet($input, removeFilterValue, removeFilterFacet);
       fireRemoveTagTrackingEvent(removeFilterValue, removeFilterFacet);
+      clearFacet($input, removeFilterValue, removeFilterFacet);
     }
 
     function clearFacet($input, removeFilterValue, removeFilterFacet) {
@@ -30,7 +35,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       if (inputType == 'checkbox') {
         $input.prop("checked", false);
-        $input.trigger('change');
+        $input.trigger(onChangeSuppressAnalytics);
       }
       else if (inputType == 'text' || inputType == 'search') {
         var splitKeywords = currentVal.split(" ");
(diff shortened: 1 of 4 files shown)


aliuk2012:
@vanitabarrett would you mind reviewing these changes as I know you have done a lot of work around this part of the code base and just to be sure its not breaking anything on Business finders

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -34,8 +34,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       }
       else if (inputType == 'text' || inputType == 'search') {
         $input.val(currentVal.replace(removeFilterValue, '').replace(/\s+/g,' ').trim()).trigger({
-          type: "change",
-          suppressAnalytics: true
```
vanitabarrett: Removing this is causing the analytics event for removing a facet tag to fire twice (which is why `suppressAnalytics` was introduced). We still need to do some sort of check to make sure we don't fire events twice in this scenario.

andysellick:
I'm trying this out in the review app and it seems like GA is still firing twice whenever a checkbox is checked or a select element is changed (however this is definitely an improvement - on master I can see 4 events when the Brexit checkbox is ticked).

Is this right?

vanitabarrett:
@andysellick Are these 2 events identical? There should be 2 non-identical events firing per checkbox change (one fires for the checkbox and one for the page view). I can only see duplicate events when a facet tag is removed.

andysellick:
Ah, yes sorry I'm getting confused. Thanks.

app/assets/javascripts/modules/remove-filter.js
aliuk2012: So I was chatting with Haur about it and we initially agreed that it was fine for it to log two events:facetTagRemoved and filterRemoved (i.e one of clicking on the facetTag  and the other for any filter changes). 

However this afternoon we were going through all the various facets and realised there is an inconsistency with this approach between removing text in the keywords, clearing select list or unticking checkboxes. 

So I'll try and introduce this again see if there is a better solution.

aliuk2012: I've reverted these changes as you have recommended.  What I have done was ensure that we use `suppressAnalytics` more across the components and the events.

I think before it was only being used for text inputs but it should of been used for checkboxes/radios and any other form elements that are changed when a user clicks on "Remove facet label"

PR #985, opened 2019-03-21
Exclude brexit terms from keyword search

leenagupte:
Trello: https://trello.com/c/kklgC5ov

## What's changed
Remove stop words from the keyword search made to rummager. This will return fewer, more relevant results.

N.B. If your keyword search only contains stop words, the search behaves as if you haven't searched by any keywords and returns all results. 

### Before
Screen Shot 2019-03-21 at 16 02 09

Screen Shot 2019-03-21 at 16 03 09

### After
Screen Shot 2019-03-21 at 16 02 17

Screen Shot 2019-03-21 at 16 03 16

### DIFF
--- a/app/lib/search_query_builder.rb
+++ b/app/lib/search_query_builder.rb
@@ -141,7 +141,7 @@ private
   def remove_stopwords
     keywords = params["keywords"].split(' ')
     keywords.delete_if do |keyword|
-      stopwords.include?(keyword.gsub(/\W/, ''))
+      stopwords.include?(keyword.downcase.gsub(/\W/, ''))
     end
     keywords.join(' ')
   end
--- a/app/lib/search_query_builder.rb
+++ b/app/lib/search_query_builder.rb
@@ -129,7 +129,13 @@ private
   end
 
   def keywords
-    remove_stopwords if params["keywords"].present?
+    return remove_stopwords if remove_stopwords?
+
+    params["keywords"].presence
+  end
+
+  def remove_stopwords?
+    params["keywords"].present? && ["/find-eu-exit-guidance-business"].include?(finder_content_item["base_path"])
   end
 
   def remove_stopwords
--- a/app/lib/search_query_builder.rb
+++ b/app/lib/search_query_builder.rb
@@ -141,7 +141,7 @@ private
   def remove_stopwords
     keywords = params["keywords"].split(' ')
     keywords.delete_if do |keyword|
-      stopwords.include?(keyword)
+      stopwords.include?(keyword.gsub(/\W/, ''))
     end
     keywords.join(' ')
   end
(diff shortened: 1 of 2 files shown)


spec/lib/search_query_builder_spec.rb
```diff
@@ -172,6 +173,88 @@
...
+            'facets' => facets,
+            'filter' => filter,
+            'reject' => reject,
+            'default_order' => default_order,
+            'default_documents_per_page' => 10
+          }
+        }
+      }
+
+      it "should not include stopwords in search" do
```
sihugh: Is it worth testing one of the contractions here too?  Like `shouldn't` or something else with an apostrophe

leenagupte: @sihugh I've updated the test here: 7cbbf34a9bf4896f7aa2a6fdac29b8c9598c1919

app/lib/search_query_builder.rb
```diff
@@ -203,4 +211,26 @@ def facet_params
...
+  def stopwords
+    %w(
+     a about above after again against all am an and any are arent as at
+     be because been before being below between both but by
+     cant cannot could couldnt
+     did didnt do does doesnt doing dont down during
+     each
+     few for from further
+     had hadnt has hasnt have havent having he hed hell hes her here heres hers herself him himself his how hows
+     i id ill im ive if in into is isnt it its its itself
```
vanitabarrett: `its` appears twice here

app/lib/search_query_builder.rb
```diff
@@ -203,4 +211,26 @@ def facet_params
...
+     lets
+     me more most mustnt my myself
+     no nor not of off on once only or other ought our ours ourselves out over own
+     same shant she shed shell shes should shouldnt so some such
+     than that thats the their theirs them themselves then there theres these they theyd theyll theyre theyve this those through to too
+     under until up
+     very
+     was wasnt we wed well were weve were werent what whats when whens where wheres which while who whos whom why whys with wont would wouldnt
```
vanitabarrett: `were` appears twice here

leenagupte: 🤦‍♀️ Copy and paste error. Will fix.

app/lib/search_query_builder.rb
leenagupte: 🤦‍♀️ Copy and paste error. Will fix.

leenagupte: @vanitabarrett Fixed and fixed up

app/lib/search_query_builder.rb
leenagupte: @vanitabarrett Fixed and fixed up

PR #988, opened 2019-03-21
Fix facet tag removal for multiple search terms

vanitabarrett:
Trello: https://trello.com/c/udoZSC9q/331-bug-fix-facet-tag-removal

This commit fixes a bug with facet tag removal for text searches. Facet tag removal was always removing the first instance of that keyword, even if it was part of another word, e.g: "therefore the". In this case, removing "the" would result in "refore the". This commit fixes that.

Approaches:

- Surrounding the keyword with a space on both sides before looking it up in the input value. However, the first and last value won't have spaces on both sides
- Adding an index to the facet tag so it can be looked up and removed by index. I got this partially implemented by adding a data attribute in KeywordFacet. However, this seemed to disappear each time a facet tag was removed. To get this solution working, I think we'd need to spend more time looking into how the facet tags are generated and/or refreshed.
- Splitting the input value by space and looping through it, replacing the current word with "" when it matched the facet tag being removed. However, this meant if a keyword existed in the search term multiple times, it was removed in all instances.
- **Implemented approach:** expanded on the above approach by exiting the loop early when the keyword was found. This means if a keyword exists multiple times in a search time, it is only removed once.

Given the following search:
Screen Shot 2019-03-21 at 16 19 19

trying to remove "the":

## Before
Screen Shot 2019-03-21 at 16 19 12

## After
Screen Shot 2019-03-22 at 10 46 34

Heroku: https://finder-frontend-pr-988.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -33,7 +33,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         $input.trigger('change');
       }
       else if (inputType == 'text' || inputType == 'search') {
-        $input.val(currentVal.replace(removeFilterValue, '').replace(/\s+/g,' ').trim()).trigger({
+        var splitKeywords = currentVal.split(" ");
+
+        for (var i = 0; splitKeywords.length; i++) {
+          if (splitKeywords[i].toString() === removeFilterValue.toString()) {
+            splitKeywords.splice(i, 1);
+            break;
+          }
+        }
+
+        var newVal = splitKeywords.join(" ").trim();
+
+        $input.val(newVal).trigger({
           type: "change",
           suppressAnalytics: true
         });
--- a/README.md
+++ b/README.md
@@ -66,8 +66,7 @@ $ bundle exec rake
 ```
 
 The application has jasmine tests, which can be accessed at `/specs` when the application is running in development mode. These are also run when `rake`, above, is run.
-To run JavaScript tests separately: 
-`bundle exec rake jasmine:test`
+To run JavaScript tests separately: `bundle exec rake spec:javascript`
 
 [govuk-content-schemas]: https://github.com/alphagov/govuk-content-schemas
 [content_schema_examples]: https://github.com/alphagov/finder-frontend/blob/master/lib/govuk_content_schema_examples.rb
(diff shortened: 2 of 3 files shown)


vanitabarrett:
**Note: looking into this again as it still doesn't work as expected**

vanitabarrett:
@andysellick Could you please take another look? I've pushed a solution which exists the loop early once a keyword is found. I think it works... 🤞 Other tried approaches documented in the PR description.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -33,7 +33,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        var splitKeywords = currentVal.split(" ");
+
+        for (var i = 0; splitKeywords.length; i += 1) {
+          if (splitKeywords[i].toString() === removeFilterValue.toString()) {
+            splitKeywords.splice(i, 1);
+            break;
+          }
+        }
+
+        var newVal = splitKeywords.join(" ")
```
andysellick: Nitpick: semi-colon.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -33,7 +33,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+        $input.val(newVal.trim()).trigger({
```
andysellick: Nitpick: I'd add the `trim` onto the end of line 45, just to make this line only about setting a value.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -33,7 +33,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         $input.trigger('change');
       }
       else if (inputType == 'text' || inputType == 'search') {
-        $input.val(currentVal.replace(removeFilterValue, '').replace(/\s+/g,' ').trim()).trigger({
+        var splitKeywords = currentVal.split(" ");
+
+        for (var i = 0; splitKeywords.length; i += 1) {
```
andysellick: Interesting, I'd write this as `i++`, I've not seen it done this way before. Works fine though.

vanitabarrett: Doh, I've got too used to Ruby not having `++` or `--`.

PR #990, opened 2019-03-22
Fix option select styling

andysellick:
Fixes some missing padding and the colour of the button/border on hover, that was lost in a recent refactor.

Before:

Screen Shot 2019-03-22 at 09 52 55

After:

Screen Shot 2019-03-22 at 09 53 09

### DIFF
--- a/app/assets/stylesheets/components/_option-select.scss
+++ b/app/assets/stylesheets/components/_option-select.scss
@@ -101,6 +101,7 @@
 }
 
 .app-c-option-select__container-inner {
+  padding: $gutter-one-quarter;
   padding-top: govuk-spacing(1);
 }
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/components/_option-select.scss
```diff
@@ -110,6 +110,7 @@
 
   .govuk-form-group {
     padding-top: govuk-spacing(2);
+    margin-bottom: $gutter-one-third;
```
vanitabarrett: Should we be using `govuk-spacing(2)` here?

andysellick: Probably, but as this is one of three uses of gutter in this file, I think we should look at it later.

PR #994, opened 2019-03-22
Add tracking to progressive disclosure link for facets

karlbaker02:
This PR adds tracking to the progressive disclosure link for facets, which previously had no tracking. It does this by bumping the `govuk_publishing_components` gem, which now includes tracking within the `GemToggle` module.

Trello: https://trello.com/c/CmGuEPyK


app/views/finders/_facet_collection.html.erb
```diff
@@ -23,7 +23,7 @@
 
   <% if facet_collection.filters.any? %>
     <div data-module="gem-toggle" data-toggle-class="facet-toggle__content--hide" <%= "class=facet-toggle--only-on-mobile" unless finder.hide_facets_by_default %>>
-      <a href="#" class="facet-toggle" data-controls="facet-wrapper" data-expanded="false" data-toggled-text="- <%= t("finders.facet_disclosure.fewer", default: "Show fewer search options") %>">
+      <a href="#" class="facet-toggle" data-controls="facet-wrapper" data-expanded="false" data-track-category="filterClicked" data-track-action="accordian" data-toggled-text="- <%= t("finders.facet_disclosure.fewer", default: "Show fewer search options") %>">
```
andysellick: Nitpick: `accordion`

features/finders.feature
```diff
@@ -288,12 +288,19 @@ Feature: Filtering documents
...
     And I should not see a "Show more search options" link
     Then I view the all content finder
     And I should see a "Show more search options" link
 
+  @javascript
+  Scenario: Facets should expand when clicking on the "Show more search options" link on all content
+    When I view the all content finder
+    And I should see a "Show more search options" link
+    Then I click to expand all facets
+    And I should see a "Show fewer search options" link
```
andysellick: While we're at it, is it worth checking to see that after the link is clicked, the content we expect is visible?

features/finders.feature
```diff
@@ -288,12 +288,19 @@ Feature: Filtering documents
...
+  Scenario: Facets should not be hidden by default on finders except all content
     When I view the research and statistics finder
     And I should not see a "Show more search options" link
     Then I view the all content finder
     And I should see a "Show more search options" link
 
+  @javascript
+  Scenario: Facets should expand when clicking on the "Show more search options" link on all content
+    When I view the all content finder
+    And I should see a "Show more search options" link
```
andysellick: This and the other test calls like this are currently failing as 'undefined'.

app/views/finders/_facet_collection.html.erb
karlbaker02: Thanks

features/finders.feature
karlbaker02: That would create a dependency on the facets that we display for the finder. It's not a very big thing and I'm happy to go either way on this.

andysellick: That's a good point. Is it possible to check that any one of several likely facets is visible? So the test would pass if a checkbox OR a select box OR a text input is suddenly visible in the space? You could even do a generic 'if there's any kind of input in this space' check?

karlbaker02: I've met you halfway in my latest commit 🙂

features/finders.feature
karlbaker02: Shouldn't be. Let's see if the new build passes

PR #995, opened 2019-03-22
Remove overriding development finders

sihugh:
We don't want the finders that we have been using for developing new features to override those that we've defined in content store.

It'll be really confusing to troubleshoot, and people may make unwitting changes to the development finders, not being aware that they might actually be in use on the production system.


sihugh:
This is being done elsewhere

PR #996, opened 2019-03-25
Changes to facet tag sentence fragments

bilbof:
Two changes here, both relating to facet tag preposition text. Please see individual commits for detail.

Screen Shot 2019-03-25 at 10 43 20

https://trello.com/c/kT5AMe3i/569


bilbof:
Thanks, @karlbaker02! Yes sorry I didn't update the Trello card. I spoke with Si and the spec was updated to have the word "Updated" as the preposition.

PR #998, opened 2019-03-25
Adds in lib object for publications routes

SamJamCul:
Adds a big list of special behaviour for the publications redirections.
Depending on the publication_filter_option that's selected, there's a
variety of redirections, which this addresses.

Tests are also included for each of these special redirects, as well as
the default path to search/all when nothing special should happen.

https://trello.com/c/8iWtTLXi/514-redirect-publications-finder-to-finder-frontend-m

### DIFF
--- a/app/controllers/redirection_controller.rb
+++ b/app/controllers/redirection_controller.rb
@@ -1,4 +1,9 @@
+require 'publications_routes'
+
 class RedirectionController < ApplicationController
+  include PublicationsRoutes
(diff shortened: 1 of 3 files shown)


app/controllers/redirection_controller.rb
```diff
@@ -1,4 +1,9 @@
+require 'publications_routes'
```
bilbof: Is requiring the file necessary?

PR #1001, opened 2019-03-26
Prevent users from subscribing to email alerts of upcoming statistics

bilbof:
This contains two changes to email signup api, split into two commits:

1. Remove validation queries before we create an email subscription - `email-alert-api` should handle this
2. Add support for `filter_values` of `facet_choices` in email signup JSONs. 

The first commit will make things clearer. It removes code that isn't needed anymore (see commit message).

The second commit implements (https://trello.com/c/qBt4q04L/548), in order to hide the option to see upcoming stats in the email alert options. See Trello card for more context (and the commit message for the implementation details). After the change (see below pic), Statistics (upcoming) won't be an option on the email signup page:

Screen Shot 2019-03-26 at 12 32 32


lib/email_alert_signup_api.rb
```diff
@@ -62,12 +62,26 @@ def massaged_attributes
...
+    chosen_options = Array(value)
+    # filter_values can be set if we want to send an array of values to
+    # the email-alert-api. E.g. document group: research, translates to
+    # to several document types being sent to rummager / email-alert-api.
+    chosen_facet_choices = choice.fetch('facet_choices', []).select { |option|
+      chosen_options.include?(option['key'])
+    }
+
+    chosen_facet_choice_values = chosen_facet_choices.map { |option| option['filter_values'] }.flatten
+
```
oscarwyatt: This took me a while to figure out, I'd suggest (but really, only suggest rather than demand) that this might be slightly easier to comprehend
```
    chosen_facet_choice_values = choice.fetch('facet_choices', []).map { |option|
      if chosen_options.include?(option['key'])
        option['filter_values']
      end
    }

    chosen_facet_choice_values.flatten!
    chosen_facet_choice_values.compact!
```

oscarwyatt:
@aliuk2012 I chatted to Si about this and made a commit to improve the description:
757fe62c5c03d0a9bf7d5070318da9e950822cd1

aliuk2012:
@oscarwyatt Thanks for addressing the issue. Still not ideal to have that extra radio button for only two checkboxes but I think the copy changes definitely helps clarify why that extra option is present.

PR #1003, opened 2019-03-26
Adds hidden clearable facet tag for manuals

oscarwyatt:
Creates new facet type called HiddenClearableFacet which is a facet that allows
users to clear but not amend or filter by that value - thus it only
shows a facet tag and not a component to amend it

This has the handy side effect of allowing users to search manuals

Trello: https://trello.com/c/CkBPvmeH/532-allow-hidden-facets-to-appear-in-facet-tags-if-used

### DIFF
--- a/app/lib/facet_query_builder.rb
+++ b/app/lib/facet_query_builder.rb
@@ -27,11 +27,11 @@ private
   attr_reader :facets
 
   def dynamic_facets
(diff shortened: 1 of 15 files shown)


karlbaker02:
To confirm that I had a chat with @oscarwyatt about this PR and the one thing left to do is fix the issue around the `Manual` facet disappearing from facet tags if you change the facets selected.

PR #1005, opened 2019-03-26
Clean up finder api

Seraphiana:
We don't want the finders that we have been using for developing new features to override those that we've defined in content store.

It'll be really confusing to troubleshoot, and people may make unwitting changes to the development finders, not being aware that they might actually be in use on the production system.


Seraphiana:
> Looks great except that `app/assets/javascripts/templates.js` has sneaked in there somehow which I'm not sure is intentional?

Should be removed now

sihugh:
👌

PR #1007, opened 2019-03-27
Option select filter add search icon

andysellick:
Make some changes to the option select component filter option.

- visually hide the label on the checkbox filter and add a search icon
- screenreader users should have the exact same experience as before the label was hidden
- looks like this:

Screen Shot 2019-03-27 at 12 43 48

https://trello.com/c/N16toILy/572-fix-minor-option-select-issues

### DIFF
--- a/app/assets/stylesheets/components/_option-select.scss
+++ b/app/assets/stylesheets/components/_option-select.scss
@@ -107,10 +107,17 @@
 
 .app-c-option-select__filter {
   display: none;
+  padding-top: govuk-spacing(2);
+  margin-bottom: govuk-spacing(2);
+}
 
-  .govuk-form-group {
-    padding-top: govuk-spacing(2);
-    margin-bottom: govuk-spacing(2);
+.app-c-option-select__filter-input {
+  padding-left: 33px;
+  background: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 36 36' width='40' height='40'%3E%3Cpath d='M25.7 24.8L21.9 21c.7-1 1.1-2.2 1.1-3.5 0-3.6-2.9-6.5-6.5-6.5S10 13.9 10 17.5s2.9 6.5 6.5 6.5c1.6 0 3-.6 4.1-1.5l3.7 3.7 1.4-1.4zM12 17.5c0-2.5 2-4.5 4.5-4.5s4.5 2 4.5 4.5-2 4.5-4.5 4.5-4.5-2-4.5-4.5z' fill='currentColor'%3E%3C/path%3E%3C/svg%3E") govuk-colour("white") no-repeat -5px -3px;
+  @include govuk-font(19);
+
+  @include govuk-media-query($from: tablet) {
+    @include govuk-font(16);
   }
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_option-select.scss
```diff
@@ -107,10 +107,17 @@
 
 .app-c-option-select__filter {
   display: none;
+  padding-top: govuk-spacing(2);
+  margin-bottom: govuk-spacing(2);
+}
+
+.app-c-option-select__filter-input {
+  padding-left: 33px;
```
vanitabarrett: Does this need to be 33px? Can it be 30px instead, and therefore use a govuk-spacing variable?

app/assets/stylesheets/components/_option-select.scss
```diff
@@ -107,10 +107,17 @@
...
+  background: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 36 36' width='40' height='40'%3E%3Cpath d='M25.7 24.8L21.9 21c.7-1 1.1-2.2 1.1-3.5 0-3.6-2.9-6.5-6.5-6.5S10 13.9 10 17.5s2.9 6.5 6.5 6.5c1.6 0 3-.6 4.1-1.5l3.7 3.7 1.4-1.4zM12 17.5c0-2.5 2-4.5 4.5-4.5s4.5 2 4.5 4.5-2 4.5-4.5 4.5-4.5-2-4.5-4.5z' fill='currentColor'%3E%3C/path%3E%3C/svg%3E") govuk-colour("white") no-repeat -5px -3px;
+  @include govuk-font(19);
 
-  .govuk-form-group {
-    padding-top: govuk-spacing(2);
-    margin-bottom: govuk-spacing(2);
+  @include media(tablet) {
```
vanitabarrett: I'm aware we're trying to move to GOV.UK Frontend. Can we use the GOV.UK Frontend media query mixin instead? https://govuk-design-system-preview.netlify.com/get-started/updating-your-code/

app/views/components/_option-select.html.erb
```diff
@@ -3,20 +3,31 @@
...
-      label: {
-        text: "Filter " + title
-      },
-      name: "option-select-filter",
-      controls: checkboxes_id,
-      describedby: checkboxes_count_id
-    })
+<% if show_filter %>
+  <%
+    filter_id ||= "input-#{SecureRandom.hex(4)}"
```
vanitabarrett: Can this be defined at the top of the file?

app/assets/stylesheets/components/_option-select.scss
andysellick: Yeah, I didn't like that either. I'm trying to line up the text in the input with the text of the labels beneath - it looks neater to me.

app/views/components/_option-select.html.erb
andysellick: I put it in here to save a tiny bit of processing - unless there is a filter, we don't need the `filter_id` variable.

andysellick:
@vanitabarrett thanks for the review, have commented. On a general level, do you think it's okay to visually hide the label like this? Does the icon explain it enough? Also @selfthinker

app/views/components/_option-select.html.erb
vanitabarrett: Good point, sorry I think I misread and didn't realise this was the option select component, not a filter component 🤦‍♀️

vanitabarrett:
@andysellick I think it's ok - the only concern would be whether it's clear enough what you're searching when you type. Maybe something to test in user research.

PR #1009, opened 2019-03-27
Add expected class to option-select component

bilbof:
`.options-container` is expected, and required by smokey:

See smokey/features/step_definitions/search_steps.rb

This will resolve critical alerts on production.


bilbof:
Closing in favour of https://github.com/alphagov/smokey/pull/490

PR #1014, opened 2019-03-28
Add A/B test to show answer box in business finder

DilwoarH:
## What
We should set up an a/b test for answer boxes as documented [here](
https://docs.google.com/spreadsheets/d/1voQzdoGAFO9Tnvl7Xq4ahLEAyGtkeAtvTC26SxEP6rE/edit?ts=5c94d616#gid=985650788).

We should make sure we have the necessary tracking in place to measure the results.

## Why
https://trello.com/c/S1FUejEn/96-%F0%9F%A5%85-goal-3-enable-users-to-get-an-answer-quickly

PR: https://github.com/alphagov/govuk-cdn-config/pull/139

## Custom dimensions
I've now set up a 2nd custom dimension in Google Analytics.

The previous one is A/B Brexit finder (69), a session custom dimension (it will persist for the entire visit) and it will contain the values A, B or C. These will indicate which version of the test a user will see.

The new custom dimension is Relevant result shown (83). This should be set to 'Yes' if a  search in the finder search box results in a 'relevant result'. (The null label is 'No'.) This is set at a hit/pageview level, since a user may do several searches in their session.

Both custom dimensions only should be set on https://www.gov.uk/find-eu-exit-guidance-business.*

https://trello.com/c/lwQK2Pi7/336-set-up-a-b-test-for-answer-boxes

### DIFF
--- a/app/controllers/concerns/finder_top_result_ab_testable.rb
+++ b/app/controllers/concerns/finder_top_result_ab_testable.rb
@@ -1,0 +1,32 @@
+module FinderTopResultAbTestable
+  CUSTOM_DIMENSION = 69
+
+  def self.included(base)
+    base.helper_method(
+      :finder_top_result_variant,
+      :show_top_result?
+    )
+    base.after_action :set_test_response_header
+  end
+
+  def finder_top_result_test
+    @finder_top_result_test ||= GovukAbTesting::AbTest.new(
+      "FinderAnswerABTest",
+      dimension: CUSTOM_DIMENSION,
+      allowed_variants: %w(A B),
+      control_variant: "A"
+    )
+  end
+
+  def finder_top_result_variant
+    @finder_top_result_variant ||= finder_top_result_test.requested_variant(request.headers)
+  end
+
+  def set_test_response_header
+    finder_top_result_variant.configure_response(response)
+  end
+
+  def show_top_result?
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/finder_top_result_ab_testable.rb
```diff
@@ -0,0 +1,28 @@
...
+  def self.included(base)
+    base.helper_method(
+      :finder_top_result_test_variant,
+      :show_top_result?
+    )
+    base.after_action :set_test_response_header
+  end
+
+  def finder_top_result_test
+    @finder_top_result_variant ||= GovukAbTesting::AbTest.new(
```
leenagupte: Linter is upset because the instance variable name is not the same as the method name.

app/controllers/concerns/finder_top_result_ab_testable.rb
```diff
@@ -0,0 +1,28 @@
...
+      "FinderAnswerABTest",
+      dimension: CUSTOM_DIMENSION,
+      allowed_variants: %w(A B),
+      control_variant: "A"
+    )
+  end
+
+  def finder_top_result_test_variant
```
leenagupte: The name is a bit too similar to `finder_top_result_test`, perhaps drop the `test` from the name?

PR #1015, opened 2019-03-28
Redirect site search to all content finder

karlbaker02:
This PR redirects users going to site search to the all content finder. Specifically, it redirects all requests to the `.json` endpoint to the all content finder, and redirects regular user requests to the all content finder if the user searches for any keywords.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -7,29 +7,14 @@ class SearchController < ApplicationController
     search_params = SearchParameters.new(params)
 
     @content_item = Services.cached_content_item("/search")
-    if search_params.no_search? && params[:format] != "json"
-      render(action: 'no_search_term') && return
+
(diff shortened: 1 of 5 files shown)


karlbaker02:
On your comment @oscarwyatt, I have checked and do not believe that a user would come to site search with both an organisation _and_ a manual set. 

For example, if you go to https://www.gov.uk/search?q=test&filter_organisations%5B%5D=department-for-transport, you'll see results for the keyword search `test` filtered by org `department-for-transport`. 

If you then go to the highway code manual and search for keyword `test` inside the manual, you'll arrive at https://www.gov.uk/search?filter_manual%5B%5D=%2Fguidance%2Fthe-highway-code&q=test, at which point you can't filter by organisation (if you add `&filter_organisations%5B%5D=department-for-transport` to the end of the query string, you get the same set of results).

What this means: I'll add in a test for manuals which is _separate_ to the test for organisations, and will not add in a test with both of them together.

PR #1016, opened 2019-03-28
Remove toolkit button sass

andysellick:
Removes the last dependency on `govuk_frontend_toolkit`, for button styles. Most of the buttons in finder-frontend had already been converted to use the button component already. 

This change involves a visual change on the subscription page. This button is now using the button component:

**Before**

Screen Shot 2019-03-28 at 14 12 41

**After**

Screen Shot 2019-03-28 at 14 12 51

This change unfortunately required a custom wrapper on the text above in order to provide margin bottom. Hopefully this will be made obsolete at some point if the govspeak component is given margin bottom.

Note: in replacing this button I noticed that the `submit_tag` rails command adds a data attribute of `data-disable-with` to the outputted HTML. I don't know if this is important but to be safe I've configured the button component to include this attribute.

Example links:

- https://finder-frontend-pr-1016.herokuapp.com/search/news-and-communications/email-signup
- https://finder-frontend-pr-1016.herokuapp.com/search/news-and-communications/ (turn off JS)

I don't think there's any other buttons, but happy to be proved wrong.

### DIFF
--- a/app/views/email_alert_subscriptions/new.html.erb
+++ b/app/views/email_alert_subscriptions/new.html.erb
@@ -37,10 +37,15 @@
       <% end %>
 
       <% if @signup_presenter.body %>
-        <%= render 'govuk_publishing_components/components/govspeak', content: @signup_presenter.body.html_safe %>
+        <div class="govspeak-wrapper">
+          <%= render 'govuk_publishing_components/components/govspeak', content: @signup_presenter.body.html_safe %>
+        </div>
       <% end %>
 
-      <%= submit_tag 'Create subscription', class: "button" %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Create subscription",
+        margin_bottom: true
+      } %>
     <% end %>
 
   </div>
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -1,7 +1,4 @@
-// govuk_frontend_toolkit helpers
-@import "design-patterns/buttons";
 @import "not-ie-conditional";
-
 @import "reset";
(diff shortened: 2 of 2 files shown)


app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -37,10 +37,18 @@
...
+          <%= render 'govuk_publishing_components/components/govspeak', content: @signup_presenter.body.html_safe %>
+        </div>
       <% end %>
 
-      <%= submit_tag 'Create subscription', class: "button" %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Create subscription",
+        margin_bottom: true,
+        data_attributes: {
+          "disable-with": "Create subscription"
```
sihugh: I'm not sure this is going to make any difference because we don't include the javascript bits that hook into this.

When it does work, the button text will change to the `disable-with` value, and the button will be disabled to prevent double clicks.  I'm not sure we're too worried about that here anyway, because we're not actually signing people up to anything at this point.

A reasonable walkthrough of the rails-ujs stuff is here: https://m.patrikonrails.com/a-definitive-guide-to-railss-unobtrusive-javascript-adapter-ef13bd047fff

aliuk2012: If we did want to disable it the govuk-frontend does include JavaScript to do that https://github.com/alphagov/govuk-frontend/releases/tag/v2.8.0  and more specifically https://github.com/alphagov/govuk-frontend/pull/1018.

andysellick: Great, thanks both, have removed this. Was being paranoid.

PR #1022, opened 2019-04-01
Improve option select component accessibility

andysellick:
Following an accessibility review of the option select component, a number of fixes are required.

- change the legend given to the included checkboxes from `Please select all that apply` to the title of the option select, e.g. `Countries`
- change the eventlistener on the filter input from `change keyup` to `keyup`. An intermittent bug was occurring where deselecting a checkbox that didn't match a filter triggered that eventlistener, causing the checkbox to immediately disappear
- change the visually hidden text associated with the filter from `x options found` to `x options found, y selected`
- move the `x selected` visible count in the head of the component outside of the button element, as this was confusing some screen readers

Trello card: https://trello.com/c/8reo7ncp/592-fix-option-select-accessibility-issues

### DIFF
--- a/app/assets/javascripts/components/option-select.js
+++ b/app/assets/javascripts/components/option-select.js
@@ -53,8 +53,6 @@
       });
     }
 
-    this.attachCheckedCounter();
-
     // Performance in ie 6/7 is not good enough to support animating the opening/closing
     // so do not allow option-selects to be collapsible in this case
     var allowCollapsible = (typeof ieVersion == "undefined" || ieVersion > 7) ? true : false;
@@ -79,6 +77,8 @@
         this.setupHeight();
       }
     }
+
+    this.attachCheckedCounter();
   }
 
   OptionSelect.prototype.cleanString = function cleanString(text) {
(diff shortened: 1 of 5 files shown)


vanitabarrett:
Would it be possible to add some more spacing to the top of the selected count? On focus of the option select dropdown button, it's very close.

Screen Shot 2019-04-03 at 11 29 24

andysellick:
@vanitabarrett yeah, I agonised a bit over that. It would mean pushing everything down a bit, which would make it look a little odd when not focused.

Off topic, I wonder if having the 'x selected' text there is the right solution anyway? I get that we want to show that something's been selected, I just wonder if there's a better way of doing it. Especially now that the filter's been added, which includes its own text equivalent to that for screen reader users. Might be good to combine them.

andysellick:
@vanitabarrett I've added in a little more spacing as suggested.

selfthinker:
I've re-tested with NVDA. The changes are definitely a good improvement. :+1:

The only thing I've found is not related to accessibility. The "2 selected" doesn't have the right font. It's fine on the [single page](https://finder-frontend-pr-1022.herokuapp.com/component-guide/option-select/with_filter) but not on the [overview](https://finder-frontend-pr-1022.herokuapp.com/component-guide/option-select) or [preview page](https://finder-frontend-pr-1022.herokuapp.com/component-guide/option-select/with_filter/preview).

andysellick:
@selfthinker good spot. I think that's a quirk of the component guide - seems to be fine on the actual site.

PR #1023, opened 2019-04-01
Refactor of facets

oscarwyatt:
*Set the value as part of the initializer in Facets instead of through setter
*Handle setting value of facets in subclasses of Facets because they mean different things in those subclasses.
*Set the values in FacetCollection as part of the initializer
*Remove unused code in FacetTagPresenter and simplify code

Almost entirely by @koetsier with minor fix from @oscarwyatt

### DIFF
--- a/app/models/checkbox_facet.rb
+++ b/app/models/checkbox_facet.rb
@@ -20,6 +20,10 @@ class CheckboxFacet < FilterableFacet
     is_checked?
   end
 
+  def has_value?
+    is_checked?
+  end
+
   def value
     facet['value'] || true
   end
--- a/app/models/date_facet.rb
+++ b/app/models/date_facet.rb
@@ -14,6 +14,10 @@ class DateFacet < FilterableFacet
     present_values.any?
   end
 
+  def has_value?
+    value.present? && value.values.any?(&:present?)
+  end
+
 private
 
   def value_fragments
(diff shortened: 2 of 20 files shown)


koetsier:
It doesn't seem to have the original PR in it?

app/models/checkbox_facet.rb
```diff
@@ -20,6 +20,10 @@ def has_filters?
     is_checked?
   end
 
+  def has_value?
```
koetsier: You could also have a ```query_hash``` method that returns either an empty hash if none exists or a hash with key->value pair.

In the method ```FinderPresenter#alert_query_string``` you can then simply combine them, so that all logic is contained in the facet models.

What do you think?

oscarwyatt: I did think about that, in the hope that I could then get rid of the `has_filters?` method too but unfortunately that's used elsewhere. I hemmed and hawed a bit and in the end decided that I felt slightly happier about `has_value` as being a clear method that just does one thing, (although so would `query_hash`) so.... I dunno, it's a bit six of one really. If you have strong feelings, I'll bow to your superior tallness

koetsier:
Handled these changes as part of https://github.com/alphagov/finder-frontend/pull/1033

PR #1030, opened 2019-04-05
Send default attributes to email alert api as tags

bilbof:
**Note:** Should be deployed after https://github.com/alphagov/email-alert-api/pull/833

Trello: https://trello.com/c/IUEMbTEA

We need to subscribe users to arbitrary groups of document types
rather than supergroups. In this case, for news and comms, we
want to subscribe users to groups of documents (news and speeches).

Rather than sending a content_purpose_supergroup or other separate
parameter, this change means we will start sending all default fields
as part of the tags. The change to email-alert-api means that these
new grouping tags will have an effect.

Changes in this commit:
- Removed all default reject filter logic. Its not used.
- Added default filter subgroups to news and comms finder and signup fixture
- Moved default filters into the tags we send to email-alert-api
- Moved EmailAlertTitleBuilder logic out of EmailAlertSignupApi
- Refactored EmailAlertSignupApi a bit to make building of tags easier to understand

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -1,4 +1,5 @@
 require 'email_alert_signup_api'
+require 'email_alert_title_builder'
 
 class EmailAlertSubscriptionsController < ApplicationController
   layout "finder_layout"
@@ -38,16 +39,17 @@ private
   end
 
   def at_least_one_filter_chosen?
-    chosen_options.any?(&:present?)
+    applied_filters.any?(&:present?)
   end
 
   def has_default_filters?
-    default_filters.present? && default_filters.any? || default_rejects.present? && default_rejects.any?
+    default_filters.present? && default_filters.any?
   end
 
-  def chosen_options
+  def applied_filters
     params
-      .permit("filter" => {})['filter']
+      .permit("filter" => {})
+      .dig('filter')
       .to_h
       .merge(
         filter_params.fetch('subscriber_list_params', {})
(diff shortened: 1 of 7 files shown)


app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -60,29 +62,27 @@ def fetch_content_item(content_item_path)
 
   def email_alert_signup_api
     EmailAlertSignupAPI.new(
-      email_alert_api: Services.email_alert_api,
-      attributes: email_signup_attributes,
-      default_attributes: { filter: default_filters, reject: default_rejects },
-      subscription_list_title_prefix: content['details']['subscription_list_title_prefix'],
-      available_choices: signup_presenter.choices,
+      applied_filters: applied_filters,
```
koetsier: I like the refactor here, but I'm thinking it might be a good idea (at some point) to move all logic out of this controller and into the EmailAlertSignupAPI.

app/models/content_item.rb
```diff
@@ -46,7 +46,8 @@ def fetch_content_item
   # ***  OVER CONTENT STORE REQUESTS  ***
   # *************************************
   FINDERS_IN_DEVELOPMENT = {
-    "/review-search/all" => 'all_content'
+    "/review-search/all" => 'all_content',
+    "/review-search/news-and-communications" => 'news_and_communications',
```
koetsier: Is this right? or a leftover debug thing?

bilbof: It'll allow you to view the changes to the pages in the Heroku app and in development, but won't be available to the public.

app/controllers/email_alert_subscriptions_controller.rb
bilbof: It might also be worth considering `email-alert-frontend`. Email signup logic will probably grow in complexity so sticking it all in one app away from search / finders might be worth it.

PR #1034, opened 2019-04-05
Remove updated options from sort

karlbaker02:
This PR removes updated options from sort options when statistics upcoming is selected

This commit removes the options `Updated (newest)` and `Updated (oldest)` from the sort options when the user chooses to view `Statistics (upcoming)`. If the user later chooses to view `Statistics (published)` or `Research`, the options are added back.

In addition, the current default sort option is `Updated (newest)`; this presents us a problem after removing it for `Statistics (upcoming)` if we need to default back to showing it, either because:
- the user previously had a keyword search with a sort option of `Relevancy`, and then removed that keyword search, or
- the user previously sorted by `Upcoming (newest)` or `Upcoming (oldest)` and then chose to view `Statictics (upcoming)`

If for either of the above reasons, we need to default to an option, we will default to `Release date (latest)`.

This PR also fixes a bug which exists in finders when a user with no JS filters by keyword, chooses the Relevancy sort option and then removes the keyword. Previously, Relevancy would have stayed as the chosen sort option, even though it was disabled. This PR updates that behaviour to choose `Updated (newest)`.

Still to-do:
- Should be switching to sort option `Updated (newest)`, but instead is `Most viewed`.
- Add / update tests.
- Figure out how to codify the fallback sort options so that we don't have to hardcode them.

Trello: https://trello.com/c/IjADqrbb

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -26,7 +26,11 @@
     this.atomHref = this.$atomLink.attr('href');
     this.$orderSelect = this.$form.find('.js-order-results');
     this.$relevanceOrderOption = this.$orderSelect.find('option[value=' + this.$orderSelect.data('relevance-sort-option') + ']');
+    this.$updatedNewestOrderOption = this.$orderSelect.find('option[value=updated-newest]');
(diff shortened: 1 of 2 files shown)


karlbaker02:
May need to discuss / pair with you @andysellick for the still to-do part of the card.

PR #1038, opened 2019-04-10
Add content security policy

rubenarakelyan:
This commit adds a content security policy header using `govuk_app_config`.


elliotcm:
@rubenarakelyan This PR is coming up as old in our channel, is it still needed?

rubenarakelyan:
Closing for now since this is blocked on work to remove the Hogan framework which is causing the tests to fail.

PR #1044, opened 2019-04-12
Use the single search end point when appropriate

sihugh:
**This is a really simple code change, but results in a lot of changes to test fixtures**

We're currently using the batch_search endpoint for all finders queries.  We shouldn't pack and unpack query results if we don't need to as it adds overhead to every step and makes URLs, responses and code harder to understand. 

The Brexit business finder was the only thing that needed to use batch_search.

This PR means that batch_search is only invoked if there is more that one query specified.  This is indicated in the finder_api_spec tests by the `batch_search_filter_params` and the fact that the second facet has the combine mode of `or`.

A lot of the changes are reducing test fixtures indentation because they're no longer wrapped in an extra `results` array.  Reviewing with whitespace disabled is a lot friendlier

- http://finder-frontend-pr-1044.herokuapp.com/search/all
- http://finder-frontend-pr-1044.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1044.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1044.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1044.herokuapp.com/find-eu-exit-guidance-business


barrucadu:
I like this, as batch searches look like they're slower in all cases than normal searches:

Screen Shot 2019-04-29 at 11 35 35

Though maybe the finder queries are just more complex than the non-finder queries, so by making this change we'll see search "slow down" to how batch search is currently performing.

PR #1046, opened 2019-04-15
Send facet label as label to GA when removing a facet

bilbof:
Currently we send the value of the facet, which can be an id. This
send the label of the facet value. E.g. Brexit rather than the brexit
content id.

Trello: https://trello.com/c/jGJHf4um/619.


sihugh:
💇‍♂️ will be delighted.  Can you let him know?

PR #1052, opened 2019-04-17
Add alternative styling option to option select

andysellick:
This PR contains the following:

- add an option to the option select component to make it look like the new expander component, which it will sit alongside in the facets down the left of a finder
- modify the JS of the option select so that the element that houses the 'x selected' text is only present if that text is present (solves a long standing layout bug)
- change the expander component so there's no 'jump' on page load when the JS kicks in and hides the content, instead, use classes and `js-enabled` to show/hide the content
- clean up the option select CSS and markup, makes it a bit cleaner and and solves a similar 'jump' problem on page load
- changes the focus state of the component to surround it rather than just put the outline on the title (button)

These changes are preparatory work ahead of implementing an accordion style wrapper for the facets in the 6 new supergroup finders and the all content finder.

Demo URL: https://finder-frontend-pr-1052.herokuapp.com/component-guide

Trello cards: 

- https://trello.com/c/NS594ugt/644-make-option-select-facet-look-like-accordion
- https://trello.com/c/RwdoOufs/669-implement-facet-redesign
- https://trello.com/c/bDJmf7Sq/609-build-the-new-facet-accordion

### DIFF
--- a/app/assets/javascripts/components/expander.js
+++ b/app/assets/javascripts/components/expander.js
@@ -11,9 +11,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
 
     var openOnLoad = this.$module.getAttribute('data-open-on-load') === 'true' ? true : false
 
-    if (!openOnLoad) {
-      this.collapseContent()
-    }
     this.replaceTitleWithButton(openOnLoad)
 
     this.$module.toggleContent = this.toggleContent.bind(this)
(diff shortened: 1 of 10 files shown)


andysellick:
@injms @sihugh I think it might be a good idea to have a whole separate view for this redesign, just for the 6 supergroup finders and the all content finder - probably also with separate CSS. I think managing two different looks will be easier that way.

In the meantime, we should delete commit 8dd9998dc65b338e677b1243fb55a5c6d01d51c9 and check that this PR hasn't affected any of the other finders.

PR #1053, opened 2019-04-17
Research and statistics filtering

koetsier:
The research and statistics finder has a radio button with the following options:

'Statistics (upcoming)'
'Statistics (published)'
'Research'

Upcoming statistics are filtered using the format and release_timestamp field. This is difficult to do using the existing RadioButton facet/filter and so we introduce a new ResearchAndStatistics facet/filter that can filter using the correct Search-api queries.

Trello: https://trello.com/c/mkf4eUXF/593-bug-stats-finder-displaying-wrong-number-of-published-stats-docs
https://trello.com/c/QpM8Vhwv/590-bug-stats-finder-is-only-displaying-a-few-upcoming-stats-docs

### DIFF
--- a/app/lib/filter_query_builder.rb
+++ b/app/lib/filter_query_builder.rb
@@ -12,8 +12,6 @@ class FilterQueryBuilder
     end
   end
 
-
-
 private
 
   attr_reader :facets, :user_params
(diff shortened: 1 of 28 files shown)


app/lib/filters/research_and_statistics_filter.rb
```diff
@@ -0,0 +1,25 @@
+module Filters
+  class ResearchAndStatisticsFilter < Filter
```
bilbof: Nice work here.

app/models/filters.rb
```diff
@@ -0,0 +1,29 @@
+module Filters
+  def self.research_and_statistics_filters
+    [
+      {
+        'key' => 'upcoming_statistics',
+        'label' => 'Statistics (upcoming)',
+        'filter' => {
+          'release_timestamp' => "from:#{Date.today.iso8601}",
+          'format' => %w(statistics_announcement)
```
bilbof: There's an argument for moving more of the data in the finder content item into this app.
The finders are completely tied to the finder-frontend - stats finder is a good example.
There's not a good argument for putting this stuff into rummager.
Perhaps we need an application in between.. `finder-api`? 😃

app/lib/filter_query_builder.rb
```diff
@@ -7,17 +7,17 @@ def initialize(facets:, user_params:)
   end
 
   def call
-    filters.select(&:active?).reduce({}) { |query, filter|
-      query.merge(filter.key => filter_value(query, filter))
-    }
+    filters.select(&:active?).map(&:query_hash).inject({}) do |query, filter_hash|
+      query.merge(filter_hash) { |_, v1, v2| Array(v1) + Array(v2) }
```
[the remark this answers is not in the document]
koetsier: value1 and value2, this block is executed if there are keys in both hashes to be merged

PR #1061, opened 2019-04-23
Update business email links

leenagupte:
Trello: https://trello.com/c/hKKTZdoj

## What
Changes the `topic_id` passed in create subscription page from `find-eu-exit-guidance-for-your-business-appear-in-find-eu-exit-guidance-business-finder` to the content_id of the [facet_group](https://github.com/alphagov/content-tagger/blob/master/lib/data/find-eu-exit-guidance-business.yml).

## Why
The new facet_group is replacing `appear-in-find-eu-exit-guidance-business-finder` as the filter for determining whether a piece of content appears on the business finder. Users need to be able to create an email subscription based on the same criteria, or they won't receive any emails.

There is another [PR](https://github.com/alphagov/email-alert-api/pull/838) that covers migrating the existing email subscriptions.

### DIFF
--- a/lib/email_alert_signup_api.rb
+++ b/lib/email_alert_signup_api.rb
@@ -20,10 +20,29 @@ private
   end
 
   def subscriber_list_options
-    {
-      "tags" => tags,
-      "title" => subscriber_list_title,
-    }
+    options = if facet_groups?
+                {
+                  "links" => facet_groups,
+                }
+              else
+                {
+                  "tags" => tags,
+                }
+              end
+
+    options.merge("title" => subscriber_list_title)
+  end
+
+  def facet_groups?
+    facets.any? { |facet| facet["facet_id"] == "facet_groups" }
+  end
+
+  def facet_groups
+    facet_groups = facets.map do |facet|
+      facet["facet_choices"]["key"]
+    end
+
+    { "facet_groups" => { any: facet_groups } }
   end
 
   def tags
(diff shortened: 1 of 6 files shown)


lib/email_alert_signup_api.rb
```diff
@@ -20,10 +20,29 @@ def subscriber_list
...
+              else
+                {
+                  "tags" => tags,
+                }
+              end
+
+    options.merge("title" => subscriber_list_title)
+  end
+
+  def facet_groups?
```
oscarwyatt: I can only find one usage of this, line 23 - to my mind such a short method that isn't used anywhere else could be included in `subscriber_list_options`. Alternatively, rename it to `has_facet_groups?`  or similar as it's more descriptive and avoids having two methods (`facet_groups` and `facet_groups?`) that are differentiated only by a question mark and could easily be confused (that's the only bit that I'm really worried about)

leenagupte: @oscarwyatt The convention in Ruby is to just add a `?` to the name of the thing you're checking if you're returning a boolean, the `has_` or `is_` is implied. 

I split out the code into to two smaller methods to make `subscriber_list_options` easy to read, so that anyone new coming to the code knows what `subscriber_list_options` is doing without having to delve too deeply.

oscarwyatt: 👍

PR #1063, opened 2019-04-24
Open first facet box by default

ChrisBAshton:
Trello: https://trello.com/c/5ZwqcRWh 
Test: https://finder-frontend-pr-1063.herokuapp.com/find-eu-exit-guidance-business

## What
- opens first facet box by default - but only if another facet isn't already selected.
- implemented this logic by adding a `ensure_minimum_one_open_facet` method, which wraps around the `FacetCollection` and ensures that there is at least one open facet.
- updates integration tests that relied on the old behaviour
- adds new integration tests that capture the new behaviour explicitly
- removes unused facet logic ('facets with more than 10 options collapsed by default') and removes the corresponding tests.
- renames `close_facet?` to `open_on_load?` & updates application to use this new API (i.e. flip from `true` to `false` and vice versa)
- removes some logic from 'views/finders/_option_select_facter.html.erb' and delegates responsibility to the model.

## Why
- we can't look at a facet in isolation and determine whether or not it should be expanded. We need to look at the other facets, and _only if none of them are selected_ do we open the first facet box.
- therefore I've had to add a method higher up the chain than simply the `OptionSelectFacet` class.
- The rename of `close_facet?` to `open_on_load?` is for readability (much easier not having to deal with double negatives!).
- The logic added to `FacetCollection` runs irrespective of the facet type. If `open_on_load?` is missing, the application errors - so I've had to stub this `open_on_load?` method in the top-level `Facet` class. It defaults to `true`.
- Removes unused facet logic ('facets with more than 10 options collapsed by default') as currently ALL facets are collapsed by default (i.e. this code doesn't appear to be doing anything!) 
- I took the opportunity to remove business logic from the view as this isn't best practice.

## Screenshots

|Current behaviour | New behaviour introduced by this PR | Behaviour when something selected (unchanged by this PR - have manually verified & added an integration test to explicitly test this) |
|-------------------|--------------------------------------|----|
|current|new default|with one selected (current   new)|

### DIFF
--- a/app/models/facet_collection.rb
+++ b/app/models/facet_collection.rb
@@ -21,10 +21,7 @@ class FacetCollection
   end
 
   def ensure_minimum_one_open_facet(facets_to_ensure)
-    at_least_one_facet_selected = (facets_to_ensure.map { |f|
-      # TODO figure out where to move this method so that it only applies to `OptionSelectFacet`. Then can remove the `defined?` check.
-      (defined? f.close_facet?) ? !f.close_facet? : false
-    }).any?
+    at_least_one_facet_selected = facets_to_ensure.any?(&:open_on_load?)
     unless at_least_one_facet_selected
       facets_to_ensure.first.open_facet!
     end
--- a/app/models/facet_collection.rb
+++ b/app/models/facet_collection.rb
@@ -23,7 +23,9 @@ class FacetCollection
   def ensure_minimum_one_open_facet(facets_to_ensure)
     at_least_one_facet_selected = facets_to_ensure.any?(&:open_on_load?)
     unless at_least_one_facet_selected
-      facets_to_ensure.first.open_facet!
+      if defined? facets_to_ensure.first.open_facet!
+        facets_to_ensure.first.open_facet!
+      end
     end
     facets_to_ensure
   end
--- a/app/models/facet_collection.rb
+++ b/app/models/facet_collection.rb
@@ -7,9 +7,9 @@ class FacetCollection
 
   def initialize(facet_hashes, values)
     stringified_values_hash = values.stringify_keys
-    @facets = facet_hashes.map { |facet_hash|
+    @facets = ensure_minimum_one_open_facet(facet_hashes.map { |facet_hash|
       FacetParser.parse(facet_hash, stringified_values_hash)
-    }
+    })
   end
 
   def to_partial_path
@@ -18,6 +18,17 @@ class FacetCollection
 
   def filters
     facets.select(&:filterable?)
+  end
+
+  def ensure_minimum_one_open_facet(facets_to_ensure)
+    at_least_one_facet_selected = (facets_to_ensure.map { |f|
+      # TODO figure out where to move this method so that it only applies to `OptionSelectFacet`. Then can remove the `defined?` check.
+      (defined? f.close_facet?) ? !f.close_facet? : false
+    }).any?
+    unless at_least_one_facet_selected
+      facets_to_ensure.first.open_facet!
+    end
+    facets_to_ensure
   end
 
   def metadata
--- a/app/models/facet_collection.rb
+++ b/app/models/facet_collection.rb
@@ -21,13 +21,16 @@ class FacetCollection
   end
 
   def ensure_minimum_one_open_facet(facets_to_ensure)
-    at_least_one_facet_selected = facets_to_ensure.any?(&:open_on_load?)
-    unless at_least_one_facet_selected
+    unless at_least_one_facet_selected? facets_to_ensure
       if defined? facets_to_ensure.first.open_facet!
         facets_to_ensure.first.open_facet!
       end
     end
     facets_to_ensure
+  end
+
+  def at_least_one_facet_selected?(facets_to_ensure)
+    facets_to_ensure.any?(&:open_on_load?)
   end
 
   def metadata
--- a/app/views/finders/_option_select_facet.html.erb
+++ b/app/views/finders/_option_select_facet.html.erb
@@ -1,12 +1,11 @@
 <% cache_if option_select_facet.cacheable?, option_select_facet.cache_key do %>
-  <% close_facet = option_select_facet.unselected? && ((option_select_facet_counter > 0) || option_select_facet.close_facet?) %>
   <%= render partial: 'components/option-select', locals: {
     :key => option_select_facet.key,
     :title => option_select_facet.name,
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => close_facet,
+    :closed_on_load => option_select_facet.close_facet?,
     :show_filter => option_select_facet.show_option_select_filter
   }
   %>
--- a/app/views/finders/_option_select_facet.html.erb
+++ b/app/views/finders/_option_select_facet.html.erb
@@ -5,7 +5,7 @@
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => !option_select_facet.open_on_load?,
+    :closed_on_load => option_select_facet.closed_on_load?,
     :show_filter => option_select_facet.show_option_select_filter
   }
   %>
--- a/app/views/finders/_option_select_facet.html.erb
+++ b/app/views/finders/_option_select_facet.html.erb
@@ -5,7 +5,7 @@
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => option_select_facet.close_facet?,
+    :closed_on_load => !option_select_facet.open_on_load?,
     :show_filter => option_select_facet.show_option_select_filter
   }
   %>
--- a/features/finders.feature
+++ b/features/finders.feature
@@ -297,6 +297,18 @@ Feature: Filtering documents
     And I should see all world locations in the world location facet
 
   @javascript
+  Scenario: Business readiness finder with no facets selected
+    When I view the business readiness finder with no preselected facets
+    Then only the first facet should be expanded
+    And all other facets should be closed by default
+
+  @javascript
+  Scenario: Business readiness finder with one facet selected
+    When I view the business readiness finder with a preselected facet
+    Then only the selected facet should be expanded
+    And all other facets should be closed by default
+
+  @javascript
   Scenario: Skip to results after inputing some keywords
     When I view the news and communications finder
     And I should not see a "Skip to results" link
--- a/features/finders.feature
+++ b/features/finders.feature
@@ -220,7 +220,7 @@ Feature: Filtering documents
   Scenario: Adding keyword filter to facet search in business finder
     When I view the business readiness finder
     Then I see topic order selected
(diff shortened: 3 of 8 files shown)


app/views/finders/_option_select_facet.html.erb
```diff
@@ -1,12 +1,11 @@
 <% cache_if option_select_facet.cacheable?, option_select_facet.cache_key do %>
-  <% close_facet = option_select_facet.unselected? && ((option_select_facet_counter > 0) || option_select_facet.close_facet?) %>
```
oscarwyatt: 👍

app/views/finders/_option_select_facet.html.erb
```diff
@@ -1,12 +1,11 @@
...
   <%= render partial: 'components/option-select', locals: {
     :key => option_select_facet.key,
     :title => option_select_facet.name,
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => close_facet,
+    :closed_on_load => option_select_facet.close_facet?,
```
oscarwyatt: I'd suggest renaming `close_facet?` to `closed_on_load?` as the former sounds like it might do something while the latter is clearer about what it does

app/models/facet_collection.rb
```diff
@@ -20,6 +20,17 @@ def filters
     facets.select(&:filterable?)
   end
 
+  def ensure_minimum_one_open_facet(facets_to_ensure)
+    at_least_one_facet_selected = (facets_to_ensure.map { |f|
+      # TODO figure out where to move this method so that it only applies to `OptionSelectFacet`. Then can remove the `defined?` check.
+      (defined? f.close_facet?) ? !f.close_facet? : false
+    }).any?
```
oscarwyatt: I agree with your comment, perhaps add `close_facet?` to `FilterableFacet` or `Facet` defaulting to false, or make it into a module and include it in the necessary facet classes - either way a default of false and overwriting the method where necessary seems sensible

app/models/facet_collection.rb
```diff
@@ -20,6 +20,16 @@ def filters
     facets.select(&:filterable?)
   end
 
+  def ensure_minimum_one_open_facet(facets_to_ensure)
+    at_least_one_facet_selected = (facets_to_ensure.map { |f| f.open_on_load? }).any?
```
oscarwyatt: Pedantic, but you can pass `any?` a block and thus refactor this to be `facets_to_ensure.any? { |f| f.open_on_load? }` which is slightly shorter (probably also a tiny, tiny bit more efficient, although I've probably wasted more CPU cycles and time then this will ever save by typing this comment.
https://apidock.com/ruby/Enumerable/any%3F

features/finders.feature
```diff
@@ -296,6 +296,21 @@ Feature: Filtering documents
...
+    Then only the first facet should be expanded
+    And all other facets should be closed by default
+
+  # TODO - remove the '@wip'. See `And(/^the second facet is selected$/) do` in `filtering_steps.rb`.
+  @javascript @wip
+  Scenario: Business readiness finder with one facet selected
+    When I view the business readiness finder
+    And the second facet is selected
+    Then only the second facet should be expanded
+    And all other facets should be closed by default
```
oscarwyatt: These are great 🥇

.gitignore
```diff
@@ -9,3 +9,4 @@
 /doc/
 /public/finder-frontend/
 /coverage/*
+app/assets/javascripts/templates.js
```
andysellick: I think this file needs to exist but be blank. I just tried deleting it and restarting my local server and I got an error. Can you check?

app/views/finders/_option_select_facet.html.erb
```diff
@@ -1,12 +1,11 @@
 <% cache_if option_select_facet.cacheable?, option_select_facet.cache_key do %>
-  <% close_facet = option_select_facet.unselected? && ((option_select_facet_counter > 0) || option_select_facet.close_facet?) %>
   <%= render partial: 'components/option-select', locals: {
     :key => option_select_facet.key,
     :title => option_select_facet.name,
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => close_facet,
+    :closed_on_load => !option_select_facet.open_on_load?,
```
andysellick: I'm not sure this is clearer. How about `:closed_on_load => !option_select_facet.closed_on_load?` ?

ChrisBAshton: Agreed this one line is less clear, but think the `open_on_load?` [method](https://github.com/alphagov/finder-frontend/pull/1063/files#diff-449972920813cbaacebc9efa5bfb12d4R50) itself is vastly clearer!

(For context, I've gone from `close_facet?` to `closed_on_load?` to `open_on_load?`)

.gitignore
ChrisBAshton: I've just checked - yes, if you delete the file, it causes an error. But `app/assets/javascripts/templates.js` exists in the repo already, this commit will just ignore any changes to it, which is what we want 😄 people who newly install the repo will continue to get a blank `templates.js` file by default.

andysellick: Okay that's fine, but I think this should be explained somewhere. Otherwise at some point someone is going to notice that this file is ignored but still in the repo and try to delete it.

ChrisBAshton: Have added a comment to the `.gitignore`.

ChrisBAshton: EDIT: git still tries to commit changes to this file 😓 so have removed from .gitignore.

andysellick:
@ChrisBAshton quick question - which finders is this intended for? Can you provide links? e.g. it doesn't seem to expand the first one on https://finder-frontend-pr-1063.herokuapp.com/search/news-and-communications, but I assume that's intentional?

ChrisBAshton:
Oh, this is just for the Business Finder: https://finder-frontend-pr-1063.herokuapp.com/find-eu-exit-guidance-business

app/models/facet_collection.rb
```diff
@@ -20,6 +20,16 @@ def filters
     facets.select(&:filterable?)
   end
 
+  def ensure_minimum_one_open_facet(facets_to_ensure)
+    at_least_one_facet_selected = facets_to_ensure.any?(&:open_on_load?)
```
brucebolt: It feels like `at_least_one_facet_selected` is acting like a method that returns a boolean.  I wonder whether a `at_least_one_facet_selected?` method would be a better choice here, rather than defining a variable?

app/views/finders/_option_select_facet.html.erb
brucebolt: Would it be possible to add a method that returns the value of `!option_select_facet.open_on_load?`?  To someone looking at the code for the first time, it might not be obviously clear what `!option_select_facet.open_on_load?` is doing, so having a named method would make that clearer.

spec/models/option_select_facet_spec.rb
```diff
@@ -59,31 +59,24 @@
...
+    context "when a facet is selected" do
+      subject { OptionSelectFacet.new(facet_data, "allowed-value-1") }
+      specify { expect(subject.open_on_load?).to be true }
     end
 
-    context "large number of options" do
-      let(:allowed_values) {
-        11.times.map { |i| { 'label' => "Label #{i}", 'value' => "allowed-value-#{i}" } }
-      }
+    context "when NO facet is selected" do
```
brucebolt: Just to be a total pedant, would it be possible to lowercase the `NO` in the context?

spec/models/option_select_facet_spec.rb
```diff
@@ -59,31 +59,24 @@
...
 
-      let(:large_facet_data) {
-        {
-          'type' => "multi-select",
-          'name' => "Test values",
-          'key' => "test_values",
-          'preposition' => "of value",
-          'allowed_values' => allowed_values,
-        }
+    context "when NO facet is selected but we have called `open_facet!`" do
```
brucebolt: Same as my previous comment regarding the `NO`

app/views/finders/_option_select_facet.html.erb
ChrisBAshton: I'm not sure what you mean - are you suggesting adding a `closed_on_load?` method to `option_select_facet.rb` like so:

```ruby
def closed_on_load?
  !open_on_load?
end
```

...and then updating this line to `:closed_on_load => option_select_facet.closed_on_load?` ...?

ChrisBAshton:
> A concern I have is that this might change behaviour on other finders. Have you tested this, and is there any way we could introduce a test to be confident behaviour elsewhere hasn't changed?

I have the same concern, so I'm grateful this is being reviewed. 😀 

The codebase doesn't seem to distinguish between particular finders, and the more the Business Finder behaviour is tweaked, the more of an issue this might become. A technical architecture problem to consider later...?

In the meantime, not knowing enough about how the other finders are supposed to work makes it tricky to test if I've broken them. The integration tests are passing, which I'm taking to mean that this hasn't broken the other finders. And I've added integration tests to cover our new expectations specific to the business finder.

app/views/finders/_option_select_facet.html.erb
brucebolt: Personally, I'd create a `closed_on_load?` method.  But I guess it's down to personal preference.  I'll leave it up to you to decide which you prefer.

ChrisBAshton: Hmm, Andrew said the same thing here: https://github.com/alphagov/finder-frontend/pull/1063#pullrequestreview-230542295

And I agree it would make this one line clearer. But it will also make other lines harder to read, e.g.

```ruby
def at_least_one_facet_selected?(facets_to_ensure)
  facets_to_ensure.any?(&:open_on_load?)
end
```

...becomes:

```ruby
def at_least_one_facet_selected?(facets_to_ensure)
  !facets_to_ensure.all?(&:closed_on_load?)
end
```

Unless we _add_ a `closed_on_load?` method in addition to `open_on_load?`...? Then we could pick and choose whatever reads best in the context. Or are you suggesting we pick `closed_on_load?` and use that exclusively?

(I don't feel strongly either way. I mainly renamed it to `open_on_load?` to make that method easier to mentally parse, as it was much easier when having to deal with the `>10 options` requirement too. But now that's gone there's not much in it!)

ChrisBAshton: Have chatted on Slack, agreed we should add additional methods.

app/models/facet_collection.rb
```diff
@@ -20,6 +20,19 @@ def filters
...
+  def ensure_minimum_one_open_facet(facets_to_ensure)
+    unless at_least_one_facet_selected? facets_to_ensure
+      if defined? facets_to_ensure.first.open_facet!
+        facets_to_ensure.first.open_facet!
+      end
+    end
+    facets_to_ensure
+  end
+
+  def at_least_one_facet_selected?(facets_to_ensure)
```
brucebolt: Could `at_least_one_facet_selected?` be renamed to `facets_selected?`?  It's fewer characters to type.

brucebolt: I'm not sure about the name `facets_to_ensure`.  Is there something simpler it could be called, e.g. just `facets` if this variable name hasn't already been used?

ChrisBAshton:
I misunderstood the requirements:

1) The open/closed state of a facet does _not_ depend on the states of other facets. So I overcomplicated this PR a lot!
2) I missed a major requirement - that it should only apply on desktop. (So I undercomplicated it too!)

Closing this in favour of #1073.

PR #1064, opened 2019-04-25
Use 'q' instead of 'keywords' parameter in text search [DO NOT MERGE]

koetsier:
Because our Google Analytics uses the 'q' parameter to process
events, we have to change the 'keywords' parameter to use 'q'
instead.

The legacy 'keywords' to 'q' translation has been reversed so that
legacy URLs using 'keywords' are now translated to 'q'

Trello: https://trello.com/c/bJQyIrmp/675-change-faceted-search-query-param-from-keywords-to-q?menu=filter&filter=label:Finders

### DIFF
--- a/app/models/keyword_facet.rb
+++ b/app/models/keyword_facet.rb
@@ -21,7 +21,7 @@ class KeywordFacet
   end
 
   def key
-    'keywords'
+    'q'
   end
 
   def value
@@ -51,7 +51,7 @@ private
         keyword_fragments << {
           'label' => keyword,
           'parameter_key' => key,
-          'name' => 'keywords',
+          'name' => 'q',
           'value' => keyword.gsub('"', "&quot;")
         }
       end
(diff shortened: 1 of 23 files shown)


app/models/keyword_facet.rb
```diff
@@ -51,7 +51,7 @@ def value_fragments
         keyword_fragments << {
           'label' => keyword,
           'parameter_key' => key,
-          'name' => 'keywords',
+          'name' => 'q',
```
barrucadu: Could `key` be used here instead so that there's no need to change both places?  Or does `key` and this `name` refer to different things?

sihugh:
There's a test for sending analytics that is no longer referenced - is it possible to reinstate? https://github.com/alphagov/finder-frontend/blob/55d601249b79a1a3af3ca42c19653d8edc75358f/features/step_definitions/search_steps.rb#L232

koetsier:
We have decided to keep 'keywords' because and to make GA differentiate between finders using the URL if necessary

PR #1065, opened 2019-04-26
Option select with sticky search

injms:
As per the [card](https://trello.com/c/wlc28Ccl/632-option-select-make-filter-stay-at-top-when-checkbox-list-is-scrolled), this makes the option-select search input remain visible whilst the checkboxes are scrolled.

Video:

option-select-sticky-input

### DIFF
--- a/app/views/components/_option-select.html.erb
+++ b/app/views/components/_option-select.html.erb
@@ -24,9 +24,7 @@
       }
     %>
   <% end %>
-  <% # NOTE - could this be achieved with a <template> instead of putting it into a data attribute? Or by putting the filter_id into a data attribute to us?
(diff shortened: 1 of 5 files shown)


app/views/components/_option-select.html.erb
```diff
@@ -24,7 +24,7 @@
       }
     %>
   <% end %>
-  <%
+  <% # NOTE - could this be achieved with a <template> instead of putting it into a data attribute? Or by putting the filter_id into a data attribute to us?
```
andysellick: Let's remove this comment and add it into the backlog - https://trello.com/c/N16toILy/572-fix-minor-option-select-issues

```suggestion
  <% 
```

PR #1066, opened 2019-04-26
Shave mustache from finder-frontend

andysellick:
Remove mustache templating from finder-frontend. 

Why?

- mustache is implemented using [shared_mustache](https://github.com/alphagov/shared_mustache), which relies upon hogan.js, which is no longer supported
- hogan.js works in a way that will violate our upcoming CSP

How?

- mustache templates have been replaced with regular erb templates
- the JSON file called by AJAX each time a finder page is dynamically updated (when a user changes a facet) has been updated to return pre-rendered blocks of HTML, generated from the new erb templates
- mustache JS updates have been replaced with taking the block of HTML from the JSON and inserting it into the relevant DOM element on the page (they are: result count, facet tags, search results, pagination).

Result?

- processing time in the browser is reduced as hogan.js no longer converts JSON data into markup
- AJAX calls have slightly increased in size as the full markup is being returned instead of just the data (from around 15KB to around 20KB, depending on the search query)
- code is easier to understand (hopefully)

Test links:

- http://finder-frontend-pr-1066.herokuapp.com/search/all
- http://finder-frontend-pr-1066.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1066.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1066.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1066.herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-1066.herokuapp.com/search/news-and-communications

Trello card: https://trello.com/c/mkyenxJy/680-remove-sharedmustache-dependency


app/views/finders/_date_facet.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render "components/date-filter", {
-                                      name: date_facet.name,
-                                      key: date_facet.key,
-                                      from_value: @results.user_supplied_date(date_facet.key, 'from'),
-                                      to_value: @results.user_supplied_date(date_facet.key, 'to'),
-                                      aria_controls_id: "js-search-results-info"
-                                    } %>
+  name: date_facet.name,
```
andysellick: 🎉 ❤️ 🎉 👍

app/presenters/result_set_presenter.rb
```diff
@@ -82,6 +110,22 @@ def signup_links
...
+    current_page = finder.pagination['current_page']
+    previous_page = current_page - 1 if current_page > 1
+    next_page = current_page + 1 if current_page < finder.pagination['total_pages']
+    pages = {}
+
+    pages[:previous_page] = build_page_link("Previous page", previous_page) if previous_page
+    pages[:next_page] = build_page_link("Next page", next_page) if next_page
+
+    markup = ""
+    markup = view_context.render(formats: %w[html], partial: 'govuk_publishing_components/components/previous_and_next_navigation', locals: pages) if pages
```
ChrisBAshton: `if pages` does nothing here, as we've already defined `pages = {}` on line 119.

app/presenters/result_set_presenter.rb
```diff
@@ -82,6 +110,22 @@ def signup_links
...
+    markup.squish unless markup.blank?
```
ChrisBAshton: `markup` is always going to be a string, isn't it? In which case, `markup.squish` is safe to call even if the string is `""`.

app/views/finders/_search_results.html.erb
```diff
@@ -0,0 +1,80 @@
...
+        <a href='<%= result[:link] %>'
+            class="document__link<% if result[:top_result] %> document__link--top<% end %>"
+            data-track-category='navFinderLinkClicked'
+            data-track-action='<%= finder_name %>.<%= document[:document_index] %>'
+            data-track-label='<%= result[:link] %>'
+            data-track-options='{"dimension28":"<%= page_count %>","dimension29":"<%= result[:title] %>"}'
+        ><%= result[:title] %></a>
+        <% if result[:summary] %>
+          <p><%= result[:summary] %></p>
+        <% end %>
```
ChrisBAshton: Can we extract lines 35-46 and lines 11-21 to a separate template as they're almost the same?

features/finders.feature
```diff
@@ -69,6 +69,16 @@ Feature: Filtering documents
     And I can see that Google won't index the page
     Then I browse to a huge page number and get an appropriate error
 
+  @javascript
+  Scenario: Updating search results
+    Given a finder with paginated results exists
+    And I visit the benefits-reform page
+    Then I should see results and pagination
+    And I fill in a keyword that should match no results
```
ChrisBAshton:
```suggestion
    When I fill in a keyword that should match no results
```

features/finders.feature
```diff
@@ -69,6 +69,16 @@ Feature: Filtering documents
...
+    Then I should not see results and pagination
+    And I click the xxxxxxxxxxxxxxYYYYYYYYYYYxxxxxxxxxxxxxxx remove control
```
ChrisBAshton: This requires internal knowledge of the step definition for `And I fill in a keyword that should match no results`.

```suggestion
    When I click the remove control for the unmatchable keyword
```

features/step_definitions/filtering_steps.rb
```diff
@@ -787,6 +792,22 @@
...
+end
+
+Then(/^I should see results and pagination$/) do
+  expect(page).to have_text('20 reports')
+  expect(page).to have_css('.finder-results', visible: true)
+  expect(page).to have_css('.gem-c-pagination')
+end
+
+Then(/^I should not see results and pagination$/) do
+  expect(page).not_to have_text('20 reports')
```
ChrisBAshton: Hmm, looks like this step should actually be `Then the results and pagination should be removed`, as that is what we're testing.

We're not concerned with the results page not having the text '20 reports', per sé - we're concerned that the page _should_ have the text '20 reports' and then we perform an action that means that text should no longer be there.

spec/javascripts/live_search_spec.js
```diff
@@ -386,59 +391,38 @@ describe('liveSearch', function () {
     })
   })
 
+  // FIXME - this needs to be replaced with a chunk of markup
```
ChrisBAshton: Fix ^

spec/javascripts/live_search_spec.js
```diff
@@ -491,7 +475,7 @@ describe('liveSearch', function () {
       liveSearch.state = { search: 'state' }
 
       expect($('#order option').length).toBe(2)
-      $('#order option').remove() // Delete all the options
+      $('#order').remove() // Delete all the options
       expect($('#order option').length).toBe(0)
```
ChrisBAshton: This `expect()` isn't necessary, surely? 🤔

spec/javascripts/live_search_spec.js
```diff
@@ -491,7 +475,7 @@ describe('liveSearch', function () {
       liveSearch.state = { search: 'state' }
 
       expect($('#order option').length).toBe(2)
-      $('#order option').remove() // Delete all the options
+      $('#order').remove() // Delete all the options
```
ChrisBAshton: This comment is probably no longer necessary.

spec/presenters/grouped_result_set_presenter_spec.rb
```diff
@@ -162,7 +162,7 @@
       allow(presenter).to receive(:any_filters_applied?).and_return(true)
       allow(view_context).to receive(:render)
       allow(presenter).to receive(:grouped_display?).and_return(true)
-      allow(presenter).to receive(:grouped_documents).and_return(key: 'value')
+      #allow(presenter).to receive(:grouped_documents).and_return(key: 'value')
```
ChrisBAshton: ^

app/presenters/result_set_presenter.rb
andysellick: Interesting! That if clause was already there. I'm happy to remove it if it's useless though.

app/presenters/result_set_presenter.rb
```diff
@@ -22,25 +23,52 @@ def initialize(finder, filter_params, view_context, sort_presenter, show_top_res
...
+
+  def sort_options_markup
+    ApplicationController.render(partial: "finders/sort_options", locals: sort_options_content).squish
+  end
+
   def any_filters_applied?
     selected_filters.length.positive? || keywords.present?
   end
 
+  # FIXME is this needed?
```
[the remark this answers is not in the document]
andysellick: Agreed. I've looked through the code and can't find any other references to it, just wanted someone knowledgeable to concur.

app/presenters/result_set_presenter.rb
andysellick: I thought so too, but I got errors trying that. I'll check again.

app/views/advanced_search_finder/show.html.erb
```diff
@@ -55,13 +55,23 @@
...
 
       <div class="govuk-grid-column-two-thirds">
-        <div class='js-live-search-results-block'>
+        <div class="js-live-search-results-block" data-module="remove-filter">
           <div class="filtered-results">
-            <div aria-live='assertive' id='js-search-results-info'>
-              <%= render_mustache('result_count', @results.to_hash) %>
+            <div class="result-info">
+              <h2 class="result-region-header__counter" id="js-result-count">
+                <%= number_with_delimiter(@results.total, :delimiter => ',') %> <%= @results.pluralised_document_noun %>
```
bilbof: You do the same thing in the ResultSetPresenter `to_hash` method for total. Could this be reused as `<%= @results.to_hash[:total] ...`?

spec/javascripts/live_search_spec.js
andysellick: Oops.

app/presenters/result_set_presenter.rb
```diff
@@ -22,25 +23,52 @@ def initialize(finder, filter_params, view_context, sort_presenter, show_top_res
 
   def to_hash
     {
```
bilbof: This gets called a fair bit. Shall we wrap this in a `@to_hash ||= begin ... end`?

app/views/finders/_facet_tags.html.erb
```diff
@@ -0,0 +1,26 @@
...
+                data-name="<%= filter[:data_name] %>">&#x2715;</button>
+            </div>
+          </div>
+        <% end %>
+      </div>
+    <% end %>
+  </div>
+<% end %>
+
+<%= local_assigns[:screen_reader_filter_description].html_safe %>
```
bilbof: There's a fair amount of `html_safe` being used. We should do an audit of these calls to ensure that the user input is escaped upstream of this.

app/views/finders/_search_results.html.erb
```diff
@@ -0,0 +1,80 @@
+<ul class="finder-results" data-module="track-click">
+  <% if local_assigns[:display_grouped_results] %>
```
bilbof: Might want to get someone from the team responsible for the grouping stuff to review this.

app/views/layouts/finder_layout.html.erb
```diff
@@ -26,6 +26,5 @@
       </main>
       <%= render 'govuk_publishing_components/components/feedback' %>
     </div>
-    <%= render_mustache_templates unless Rails.env.production? %>
```
DilwoarH: Why did we have this in here? Do you know? Didn't we render the page the same as dev?

app/views/finders/show.html.erb
```diff
@@ -99,26 +99,35 @@
         <div class="result-info">
           <div class="govuk-grid-row">
             <div class="govuk-grid-column-one-third">
-              <%= render_mustache('result_count', @results.to_hash) %>
+              <h2 class="result-region-header__counter" id="js-result-count">
+                <%= number_with_delimiter(@results.total, :delimiter => ',') %> <%= @results.pluralised_document_noun %>
```
bilbof: Same here, could `@results.to_hash[:total]` be used?

features/finders.feature
```diff
@@ -69,6 +69,16 @@ Feature: Filtering documents
     And I can see that Google won't index the page
     Then I browse to a huge page number and get an appropriate error
 
+  @javascript
```
bilbof: Shall we have the same test without JS enabled?

andysellick:
> I'm a bit concerned that the API for finders are no longer available like they used to be.

@DilwoarH I intended to keep the existing route and add a new one, but was told that would involve republishing everything. And expanding the current route to include the new content while leaving the existing content in place would result in an unnecessarily large ajax call.

app/assets/javascripts/live_search.js
```diff
@@ -311,14 +311,18 @@
     this.$loadingBlock.text('Error. Please try modifying your search and trying again.')
   }
 
+  LiveSearch.prototype.updateElement = function updateElement (element, content) {
+    element.html(content)
```
bilbof: This opens us up to potential XSS, as `.html` does not "remove or escape any user input before adding content to the document" ([source](http://api.jquery.com/html/)). So we need to have a look at what we're returning via the JSON request.

app/views/finders/_search_results.html.erb
andysellick: They do look similar, but have a number of key differences that I think would make that new template a little messy. For example, 

`data-track-action='<%= finder_name %><%= result[:document_index] %>'` 
`data-track-action='<%= finder_name %>.<%= document[:document_index] %>'`

and

`data-track-options='{"dimension28":"<%= result[:page_count] %>","dimension29":"<%= result[:title] %>"}'`
`data-track-options='{"dimension28":"<%= page_count %>","dimension29":"<%= result[:title] %>"}'`

features/finders.feature
andysellick: It does, yes, but it also reuses an existing method for clicking a given keyword facet tag. I think I'd rather reuse code here.

app/views/layouts/finder_layout.html.erb
andysellick: This is mentioned in the installation section of the [shared_mustache readme](https://github.com/alphagov/shared_mustache/blob/bd11dbd0b1ccf700627ff12de8ddc2029d77bceb/README.md) - it's part of installing shared_mustache. No longer needed.

app/assets/javascripts/live_search.js
DilwoarH: @bilbof - Isn't the escaping happening on the ERB template?

andysellick:
> As this is a fairly significant change to the way results are updated and rendered first time, do we know the performance implications of this change for end users?

@bilbof broadly speaking, it should mean less processing by the browser but slightly more data needing to be downloaded when facets are changed. In terms of specifics, we've got speedcurve set up to track the finders already, so we have a performance baseline ready to assess the impact of these changes.

app/assets/javascripts/live_search.js
bilbof: By default it does, but we're using `html_safe` some things which I think will disable this.

DilwoarH:
This there is a reason why grouped documents still need to come back in the API if the rendering is happening in the backend?

bilbof:
Thanks, @andysellick. Perhaps we could deploy this to integration and take a look at what the performance for the requests look like?

app/presenters/advanced_search_result_presenter.rb
```diff
@@ -33,4 +34,42 @@ def show_metadata?
 
     true
   end
+
+  def metadata_bundle
```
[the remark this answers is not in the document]
injms: It's used in the `to_hash` method in `AdvancedSearchResultPresenter`. It's to make sure that the metadata information is in the same shape for both the finder and the advanced finder. With hindsight, it means that 

```ruby
document_type: document_type,
organisations: organisations,
publication_date: publication_date,
```

can be removed from `to_hash`.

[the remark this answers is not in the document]
bilbof: Ok cool. In this case, could we refactor this method a bit, as there is some duplication and potentially unrequired type checking.

I think @hannako has been working on refactoring the metadata presentation logic, so might have some context on how to do this.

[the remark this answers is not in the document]
andysellick: @injms can you have a think about this and I'll let you know when I've finished going through the rest of this feedback so you can implement it then?

app/presenters/result_set_presenter.rb
andysellick: I'm not familiar with that syntax. Can you be more specific?

bilbof: This result of this change is that the first time `to_hash` is called, we cache the result as a variable, and then rather than performing expensive operations again, we return the cached value.

```
def to_hash
   @to_hash ||= begin
      { data: big_request_that_takes_a_long_time }
   end
end
```

app/views/finders/_facet_tags.html.erb
andysellick: I've removed this instance of `html_safe`.

app/assets/javascripts/live_search.js
andysellick: We're now only using `html_safe` on the next and prev links, which doesn't involve any user input so I think it's safe (plus a few other instances of things that weren't added by this PR but also look safe).

app/views/finders/show.html.erb
```diff
@@ -99,15 +99,18 @@
...
+              </h2>
             </div>
             <% if @results.has_email_signup_link? %>
               <div class="govuk-grid-column-two-thirds">
                 <%= render "govuk_publishing_components/components/subscription-links", @results.signup_links %>
               </div>
             <% end %>
           </div>
-          <%= render_mustache('result_facet_tags', @results.to_hash) %>
+          <%#= render_mustache('result_facet_tags', @results.to_hash) %>
```
leenagupte: Should this line be deleted?

app/views/finders/show.html.erb
```diff
@@ -116,7 +119,8 @@
       <%= render_mustache('sort_options', @results.to_hash) %>
 
       <div id="js-results">
-        <%= render_mustache('results', @results.to_hash) %>
+        <%#= render_mustache('results', @results.to_hash) %>
```
leenagupte: And this one? Should it be deleted too?

spec/presenters/result_set_presenter_spec.rb
```diff
@@ -237,8 +237,8 @@
 
   describe '#to_hash' do
     before(:each) do
-      allow(presenter).to receive(:selected_filter_descriptions).and_return("a sentence summarising the selected filters")
-      allow(presenter).to receive(:documents).and_return(key: 'value')
+      #allow(presenter).to receive(:selected_filter_descriptions).and_return("a sentence summarising the selected filters")
+      #allow(presenter).to receive(:documents).and_return(key: 'value')
```
leenagupte: Comment police: Should these lines be removed?

app/views/finders/_search_results.html.erb
leenagupte: cc @steventux

app/views/finders/_facet_tags.html.erb
```diff
@@ -1,28 +1,26 @@
...
-                <button type="button"
-                  class="facet-tag__remove"
-                  aria-label="Remove filter <%= filter[:text] %>"
-                  data-module="remove-filter-link"
-                  data-track-label="<%= filter[:data_track_label] %>"
-                  data-facet="<%= filter[:data_facet] %>"
-                  data-value="<%= filter[:data_value] %>"
-                  data-name="<%= filter[:data_name] %>">&#x2715;</button>
-              </div>
+<% if local_assigns[:applied_filters] %>
```
leenagupte: What was `js-facet-tag-wrapper`? and why was it removed?

app/views/finders/show.html.erb
```diff
@@ -109,20 +109,21 @@
               </div>
             <% end %>
           </div>
-          <%#= render_mustache('result_facet_tags', @results.to_hash) %>
-          <%= render "facet_tags", @results.to_hash %>
+          <div id="js-facet-tag-wrapper">
```
leenagupte: Why was `js-facet-tag-wrapper` removed from the `facet_tags` partial and moved here? Do we need to render the `facet_tags` partial without it being wrapped in `js-facet-tag-wrapper`?

app/views/finders/_facet_tags.html.erb
andysellick: It was a class used as a hook by the JS for updating content dynamically. I've replaced it with an ID to make it consistent with the other hooks.

app/views/finders/show.html.erb
andysellick: It was moved from the partial because we're now rendering the partial in the backend and passing the resulting markup to the frontend via AJAX (previously done with mustache). The JS replaces the content of `js-facet-tag-wrapper` with this chunk, so if it was still inside the partial it would include itself. So after you'd changed the facets a few times, the markup would look like this:

```
<div id="js-facet-tag-wrapper">
   <div id="js-facet-tag-wrapper">
      <div id="js-facet-tag-wrapper">
         (content)
      </div>
   </div>
</div>
```

app/presenters/advanced_search_result_presenter.rb
```diff
@@ -33,4 +34,42 @@ def show_metadata?
...
+    return unless show_metadata?
+
+    metadata_to_return = Array.new
+
+    if publication_date.is_a? Hash
```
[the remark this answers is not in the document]
andysellick: @injms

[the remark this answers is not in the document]
injms: I tried and it complained a lot - I thought that if `publication_date` was `nil`, the check for `publication_date.has_key? :label` would throw an error as that it relied on `publication_date` being a `Hash`. Is that a correct assumption?

I also tried both ways around, with `has_key?` being before and after the `&&` and errors were thrown both ways.

app/presenters/result_set_presenter.rb
```diff
@@ -22,25 +23,52 @@ def initialize(finder, filter_params, view_context, sort_presenter, show_top_res
...
+      total: "#{number_with_delimiter(total)} #{document_noun.pluralize(total)}",
+      any_filters_applied: any_filters_applied?,
+      next_and_prev_links: next_and_prev_links,
+      facet_tags: facet_tags_markup,
+      search_results: search_results_markup,
+      sort_options: sort_options_markup
+    }
+  end
+
+  def search_results_content
```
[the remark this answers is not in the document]
andysellick: I've moved some of them to be private now, specifically all the `_markup` ones, that only get called inside the presenter. The rest of the ones I've introduced are called in templates.

app/presenters/result_set_presenter.rb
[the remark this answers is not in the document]
andysellick: Removed.

app/presenters/result_set_presenter.rb
andysellick: Removed.

app/presenters/result_set_presenter.rb
andysellick: Refactored.

spec/javascripts/live_search_spec.js
andysellick: It's probably not but I'm going to leave it as it isn't part of this PR.

app/presenters/advanced_search_result_set_presenter.rb
```diff
@@ -3,7 +3,7 @@ class AdvancedSearchResultSetPresenter < ResultSetPresenter
 
   def to_hash
     super
-      .merge(applied_filters: applied_filters_or_all_subgroups)
+      .merge(applied_filters: selected_filter_descriptions)
```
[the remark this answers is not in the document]
andysellick: @injms

app/presenters/advanced_search_result_presenter.rb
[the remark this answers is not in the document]
andysellick: Updated.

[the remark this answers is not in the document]
koetsier: ```&&``` will evaluate the first argument and if that's false won't evaluate the second argument because the result will always be false so it should not throw an error.
It does need some extra parentheses to make it work so that might be the cause of the errors

DilwoarH:
> Is there is a reason why grouped documents still need to come back in the API if the rendering is happening in the backend?
> 
> https://finder-frontend-pr-1066.herokuapp.com/find-eu-exit-guidance-business.json

^^

app/presenters/advanced_search_result_set_presenter.rb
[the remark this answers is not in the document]
bilbof: Chatted with Si and Ian. Lets remove the `applied_filters_or_all_subgroups` method!

app/views/finders/show.html.erb
leenagupte: Could you add this to the commit message?

andysellick:
> Is there is a reason why grouped documents still need to come back in the API if the rendering is happening in the backend?

@DilwoarH this has been removed, good spot, thanks.

app/presenters/advanced_search_result_presenter.rb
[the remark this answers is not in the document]
hannako: @injms 

Maybe you could create a new subclass of the `MetadataPresenter` for handling metadata on the advanced search results page? eg  `return AdvancedMetadataPresenter if is_advanced_search?`to the content item [here](https://github.com/alphagov/finder-frontend/blob/d6aed62ef251a5606ff08c2676f461d989952b44/app/models/content_item.rb#L31)?

bilbof:
Lets hold off on merging this for now. There's a minor issue come up from testing this against live data on staging.

app/views/finders/show.html.erb
andysellick: Added.

app/presenters/result_set_presenter.rb
```diff
@@ -34,8 +34,16 @@ def to_hash_data
...
       facet_tags: facet_tags_markup,
       search_results: search_results_markup,
-      sort_options: sort_options_markup
-    }
+      sort_options_markup: sort_options_markup,
+    }.merge(legacy_attributes)
+  end
+
+  # Provided for backwards compatibility
+  def legacy_attributes
```
koetsier: nice 👍

PR #1067, opened 2019-04-26
Fix removing keyword filters which have a "

barrucadu:
The issue is that the input field value had `"` but the filter tag had `&quot;`.  I've solved *this specific case*, but I can't find a nice way to handle HTML entities in general (without rendering a div to a page and mucking about with its innerHTML... but that then opens the possibility of XSS attacks).

If there's a better way to do this, it would be good to know.

Tested on staging as I couldn't get the javascript tests to run locally even with a clean checkout of master:

```
=> finder-frontend component guide available at: /component-guide
Running `"phantomjs"  "/usr/lib/rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/jasmine-rails-0.14.8/lib/jasmine_rails/../assets/javascripts/jasmine-runner.js" "file:///var/govuk/finder-frontend/tmp/jasmine/runner.html?root=/var/govuk/finder-frontend&spec="`
Running: file:///var/govuk/finder-frontend/tmp/jasmine/runner.html?root=/var/govuk/finder-frontend&spec=
Starting...
GOVUK.OptionSelect initialising when the parent is hidden and data-closed-on-load is true sets the height of the container sensibly when the option select is opened
  Expected 512 to be less than 500.
    Error: Expected 512 to be less than 500.
(a big stack trace goes here)
```

---

[Trello card](https://trello.com/c/lUl9kTAQ/676-bug-search-input-text-with-surrounding-quotes-cannot-be-removed-using-a-facet-tag)

### DIFF
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -39,10 +39,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         $input.trigger(onChangeSuppressAnalytics);
       }
       else if (inputType == 'text' || inputType == 'search') {
+        var needle = decodeEntities(removeFilterValue.toString());
         var splitKeywords = currentVal.split(" ");
 
         for (var i = 0; i < splitKeywords.length; i++) {
-          if (splitKeywords[i].toString() === removeFilterValue.toString()) {
+          if (splitKeywords[i].toString() === needle) {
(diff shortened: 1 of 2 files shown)


bilbof:
Regarding the JS specs error: I saw the same thing locally. I asked @andysellick about it this morning, and I understand it's a transient error on master. We should fix it at some point.

PR #1069, opened 2019-04-26
Use a different sort presenter for statistics finder

bilbof:
Trello: https://trello.com/c/IjADqrbb/591

Test the change: https://finder-frontend-pr-1069.herokuapp.com/search/research-and-statistics.

## What does this do?

This PR enables something quite simple: it makes it possible exclude some sort options on the stats finder depending on some conditions.

Those conditions are:

- we remove 'Updated at' options from the sort select input when Statistics (upcoming) is selected
- we remove 'Release date' options from the sort select input when Statistics (published) is selected

Screen Shot 2019-04-26 at 16 32 48
Screen Shot 2019-04-26 at 16 32 57

This logic is needed for these reasons:

- we need feature parity with the Whitehall finder this one is replacing
- we need to show only one date sort field to the user, and that field should be relevant to the documents the user is viewing.

## Why so much code for a simple change?

A lot of this is just moving things around in the refactor, and there are a fair few tests for the new behaviour.

The refactor should help us make changes like this (special sort behaviour) more easily in the future. For example, if we want to refresh the sort options on the fly in the future (say, if we wanted a supergroup facet with different sort options for each supergroup), it'll be a bit easier.

## How should I review this?

Reviewing it by commit will be easiest.

In the first commit I move the sort presenter logic from `FinderPresenter` to separate classes `SortPresenter` and `SortOptionPresenter`.

In the second commit I update the sort options in the DOM with Ajax calls.

In the final commit I add a new statistics sort presenter, and implement the special logic for the stats finder.


app/presenters/finder_presenter.rb
```diff
@@ -145,57 +141,8 @@ def text_metadata_keys
...
-    disabled_option = keywords.blank? ? relevance_sort_option_value : ''
-
-    selected_option = if values['order'].present? && sort.any? { |option| option['name'].parameterize == values['order'] }
-                        values['order']
-                      else
-                        default_sort_option_value
-                      end
-
-    options_for_select(options, selected: selected_option, disabled: disabled_option)
+    @sort_options ||= SortPresenter.new(content_item, values)
```
koetsier: I'm wondering if it is maybe better to instantiate this class outside of this class and move ```atom_feed_enabled?``` etc into ```SortPresenter```

app/presenters/sort_option_presenter.rb
```diff
@@ -0,0 +1,36 @@
+class SortOptionPresenter
```
koetsier: Could we maybe find a different name for this? - Or maybe move it into ```SortPresenter``` somehow. The difference between the name ```SortOptionPresenter``` and ```SortPresenter``` is not immediately obvious from the name alone

app/controllers/finders_controller.rb
```diff
@@ -98,7 +98,7 @@ def parent
   end
 
   def grouped_display?
-    params["order"] == "topic" || finder.default_sort_option_key == "topic"
+    params["order"] == "topic" || finder.sort_options.default_option.try(:key) == "topic"
```
koetsier: could move this logic into ```sort_options_presenter```

app/presenters/finder_presenter.rb
koetsier: It might simplify the code a bit as well as there are quite a number of references to ```finder. sort_options.method``` throughout the code

app/assets/javascripts/live_search.js
```diff
@@ -219,30 +216,36 @@
     }
   };
 
+  LiveSearch.prototype.updateSortOptions = function updateSortOptions(results, action) {
+    if(action !== $.param(this.state)) { return }
```
koetsier: What is the reason 'action' is passed here?

bilbof: We want to ensure that the action is in sync with the state.

If you make a request and then make another request we only want the second request to update the DOM. This ensures that happens by tracking the state at the time of the request with the current state.

app/presenters/finder_presenter.rb
bilbof: This methods is not related to presenting sort so I’d argue that it could stay here / be moved elsewhere. Looking at sort presenter to check whether one can render the atom feed might look a bit odd.

It’s a shame things rely on sort for their logic. Perhaps we could rewrite this a bit so they don’t?

app/presenters/sort_option_presenter.rb
bilbof: I’d prefer to keep the classes in separate files. Maybe we could rename sort presenter to something like sort collection presenter?

app/controllers/finders_controller.rb
```diff
@@ -90,6 +90,10 @@ def result_set_presenter_class
     ResultSetPresenter
   end
 
+  def sort_presenter
+    content_item.sorter_class.new(content_item.as_hash, filter_params)
```
koetsier: might want to memoize this one because it is called a number of times. Maybe better to instantiate in the 'show' method?

app/presenters/statistics_sort_presenter.rb
```diff
@@ -0,0 +1,98 @@
+class StatisticsSortPresenter < BaseSortPresenter
```
koetsier: There is a lot of duplicated code in ```SortPresenter```. Can we remove some of this? It might make it clearer what the difference is between ```SortPresenter``` and ```StatisticsSortPresenter```

app/controllers/finders_controller.rb
```diff
@@ -49,14 +49,14 @@ def content_item
   end
 
   def results
-    @results ||= result_set_presenter_class.new(finder, filter_params, view_context, show_top_result?)
+    @results ||= result_set_presenter_class.new(finder, filter_params, view_context, sort_presenter, show_top_result?)
```
koetsier: Could we somehow avoid having result_set_presenter and the finder_presenter depend on the sort_presenter?

bilbof: Yes, though it might be worth waiting for Andy's change to the templates to complete.

The result set presenter is responsible for sending a hash to the frontend as a response to the Ajax call to the `.json` endpoint.

As we now update the sort options (using the sort presenter) via this same route we need to give the sort presenter (or at least it's data) to the result set presenter.

Once Andy's template changes have been made perhaps we could make some changes to the relationship between presenters and their templates?

app/presenters/sort_option_presenter.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def initialize(label: label, key: key, default: false)
+    @label = label
+    @value = label.parameterize
+    @key = key
+    @default = default
+  end
+
+  def default?
+    default.present? && default
```
karlbaker02: Will `default` not always be present (as it's defaulted in the `initialize` if a value isn't passed in)? 

Then we could have

```
def default?
  default
end
```

or even just add `default` to `attr_reader` ?

spec/presenters/sort_presenter_spec.rb
```diff
@@ -0,0 +1,188 @@
+require 'spec_helper'
+require "helpers/taxonomy_spec_helper"
+
+RSpec.describe SortPresenter do
+  include GovukContentSchemaExamples
+
+  subject(:presenter) { described_class.new(content_item(sort_options: no_sort_options), {}) }
```
karlbaker02: Is this necessary (same as the one below which is used) ?

spec/presenters/sort_presenter_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+        sort_option('Updated (newest)', 'updated-newest'),
+        sort_option('Relevance', 'relevance', disabled: false)
+      ].join("\n")
+
+      presenter = described_class.new(content_item(sort_options: sort_options_with_relevance), "keywords" => "something not blank")
+
+      expect(presenter.for_select).to eql(expected_options)
+    end
+
+    it "returns sort options with no option selected when order is specified but does not exist in options" do
```
karlbaker02: Perhaps we should return sort options with the default specified, as in the test below, if order is specified but does not exist?

spec/presenters/sort_presenter_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    it "returns true if there is a default option specified in the content item" do
+      expect(presenter_with_default.has_default_option?).to be true
+    end
+
+    it "returns false if there is not a default option specified in the content item" do
+      expect(presenter_with_sort.has_default_option?).to be false
+    end
+  end
+
+  describe "#has_default_option?" do
```
karlbaker02: Duplicate `describe` block (first defined from line 132).

features/step_definitions/filtering_steps.rb
```diff
@@ -558,14 +560,18 @@
   expect(page).to have_select('order', selected: "Relevance")
 end
 
-Then(/^I see updated newest order selected$/) do
-  expect(page).to have_select('order', selected: "Updated (newest)")
-end
+# Then(/^I see updated newest order selected$/) do
```
karlbaker02: Can we remove this?

karlbaker02:
This is great work, will make the finders much more versatile moving forwards 🥇

spec/presenters/sort_presenter_spec.rb
bilbof: No it's not needed, I ended up removing it in a later commit.

https://github.com/alphagov/finder-frontend/pull/1069/files#diff-7c61a92169a590942f2d8b5226c6937d

app/presenters/sort_option_presenter.rb
bilbof: `@default` can be `nil`, so I prefer to define a method that explicitly returns a boolean. I could also do `@default = default || false` in the initialiser and then use an `attr_reader`, I suppose 😃

spec/presenters/sort_presenter_spec.rb
bilbof: Perhaps, yes. This behaviour is consistent with the existing behaviour, though so I think this change could be done in a separate commit.

https://github.com/alphagov/finder-frontend/pull/1069/files#diff-7c61a92169a590942f2d8b5226c6937dR120.

spec/presenters/sort_presenter_spec.rb
bilbof: Thanks. This has been fixed in a later commit.

features/step_definitions/filtering_steps.rb
bilbof: Good spot. I amended the previous commit.

app/presenters/sort_option_presenter.rb
karlbaker02: I see your thinking. That's cool 👍

PR #1073, opened 2019-04-29
Open first facet box by default - Business Finder

ChrisBAshton:
Trello: https://trello.com/c/5ZwqcRWh 
Heroku URL: https://finder-frontend-pr-1073.herokuapp.com/find-eu-exit-guidance-business

## What
- the first facet box on the business finder now opens by default, regardless of what is selected
- removes old "force facet closed if > 10 options" logic
- removes redundant `close_facet?` method

## Why
- the "10 option limit" was added in fa89296d41554268cc412e1df7611d7a79c9625c to fix a usability issue on mobile. The designs have since changed to toggle _all_ facets on mobile, so this should no longer be an issue.
- `close_facet?` cannot contain all of the information required to decide whether or not a facet should be closed, as the OptionSelectFacet model doesn't know whether it is the first facet in the page or not. It's _more_ confusing trying to delegate this decision to the model when in fact it also depends on external factors. This is a front-end problem and is best handled with a tiny bit of front-end logic.

## Screenshots

Here's the screenshots demonstrating the issue that the "10 option limit" resolved:

| Landing view | Scrolling below the fold |
|--------------|-------------------------|
|Image-1|Image-2|

And here is the most recent design, demonstrating that the 10 option limit is no longer required (expanded facet has a max height, making it easy to scroll past):

| Landing view (options behind toggle) | Options expanded |
|--------------|-------------------------|
|Screen Shot 2019-05-01 at 10 30 45|Screen Shot 2019-05-01 at 10 30 55|

### DIFF
--- a/app/models/option_select_facet.rb
+++ b/app/models/option_select_facet.rb
@@ -42,10 +42,6 @@ class OptionSelectFacet < FilterableFacet
     selected_values.any?
   end
 
-  def close_facet?
-    unselected? && allowed_values.count > 10
-  end
-
   def unselected?
     selected_values.empty?
   end
--- a/app/views/finders/_option_select_facet.html.erb
+++ b/app/views/finders/_option_select_facet.html.erb
@@ -1,12 +1,11 @@
 <% cache_if option_select_facet.cacheable?, option_select_facet.cache_key do %>
-  <% close_facet = option_select_facet.unselected? && ((option_select_facet_counter > 0) || option_select_facet.close_facet?) %>
   <%= render partial: 'components/option-select', locals: {
     :key => option_select_facet.key,
     :title => option_select_facet.name,
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => close_facet,
+    :closed_on_load => option_select_facet_counter > 0 && option_select_facet.unselected?,
     :show_filter => option_select_facet.show_option_select_filter
   }
   %>
--- a/features/finders.feature
+++ b/features/finders.feature
@@ -220,7 +220,7 @@ Feature: Filtering documents
   Scenario: Adding keyword filter to facet search in business finder
     When I view the business readiness finder
     Then I see topic order selected
-    And I click button "Sector / Business Area" and select facet Aerospace
+    And I select facet Aerospace in the already expanded "Sector / Business Area" section
     Then I see results grouped by primary facet value
     And I see results with pinned items
     And I fill in some keywords
@@ -285,7 +285,7 @@ Feature: Filtering documents
   @javascript
   Scenario: Filter documents and group by facets
     When I view the business readiness finder
-    And I click button "Sector / Business Area" and select facet Aerospace
+    And I select facet Aerospace in the already expanded "Sector / Business Area" section
     Then I see results grouped by primary facet value
 
   Scenario: Dynamic facets continue to show all options on page reload
--- a/features/step_definitions/filtering_steps.rb
+++ b/features/step_definitions/filtering_steps.rb
@@ -634,6 +634,10 @@ And(/^I click button \"([^\"]*)\" and select facet (.*)$/) do |button, facet|
   find('label', text: facet).click
 end
 
+And(/^I select facet (.*) in the already expanded \"([^\"]*)\" section$/) do |facet, _button|
+  find('label', text: facet).click
+end
+
 When(/^I click the (.*) remove control$/) do |filter|
   expect(page).to have_css(".js-enabled")
 
--- a/spec/models/option_select_facet_spec.rb
+++ b/spec/models/option_select_facet_spec.rb
@@ -59,34 +59,6 @@ describe OptionSelectFacet do
     end
   end
 
-  describe "#close_facet?" do
-    subject { OptionSelectFacet.new(facet_data, []) }
-
-    context "small number of options" do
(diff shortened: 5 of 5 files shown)


app/views/components/_option-select.html.erb
```diff
@@ -30,6 +30,8 @@
 
 <div class="app-c-option-select <% if expander_style %>app-c-option-select--expander<% end %>"
   <% if local_assigns.include?(:closed_on_load) && closed_on_load %>data-closed-on-load="true"<% end %>
+  <% if local_assigns.include?(:closed_on_desktop) && closed_on_desktop %>data-closed-on-desktop="true"<% end %>
+  <% if local_assigns.include?(:closed_on_mobile) && closed_on_mobile %>data-closed-on-mobile="true"<% end %>
```
leenagupte: Do we need to have separate values for desktop and mobile? Can `closed_on_load` expanded include the different conditions for desktop and mobile?

app/views/finders/_option_select_facet.html.erb
```diff
@@ -1,12 +1,14 @@
...
     :key => option_select_facet.key,
     :title => option_select_facet.name,
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => close_facet,
+    # on desktop, if it's unselected and it's not the first facet, close it.
+    :closed_on_desktop => option_select_facet.unselected? && option_select_facet_counter > 0,
+    # on mobile, if it's unselected and it's not the first facet, or if it IS the first (and unselected) facet but contains more than 10 options, close it.
+    :closed_on_mobile => option_select_facet.unselected? && (option_select_facet_counter > 0 || (option_select_facet_counter == 0 && option_select_facet.allowed_values.count > 10)),
```
leenagupte: As with the previous comment, rather than having a lot of logic in the view, which is much harder to test, I would expand `closed_on_load` to handle the different conditions for desktop and mobile. That way, the view would remain virtually untouched.

app/models/option_select_facet.rb
```diff
@@ -42,10 +42,6 @@ def has_filters?
     selected_values.any?
   end
 
-  def close_facet?
```
leenagupte: I understand your point about not having this decision be made in the model. It's something that would be better suited in a presenter rather than a view. Also, presenter tests are just unit tests, and therefore much easier to write!

leenagupte:
@ChrisBAshton I've added a couple of suggestions. They're mainly about simplifying the logic closing the facet by keeping all of the checks in one place, and removing code from the view.  This is a much simpler approach than #1063 though!

brucebolt:
A few thoughts from me:

- I think the logic for desktop/mobile should live in the frontend code and should depend on the viewport size, rather than a binary decision between desktop/mobile (so someone with a narrow screen on a computer would also see the same), but also so someone changing the viewport width wouldn't need to refresh the page to see the alternative behaviour.
- Following on from that, I don't think we actually need to change the behaviour between mobile and desktop.  Right now, the facets are all hidden on mobile by default (under a heading that only appears when the viewport width is reduced).  So it doesn't really matter if the facet box is open, since it will be hidden until the user interacts.  See the screenshots below.
- Looking at other finders (e.g. gov.uk/cma-cases or gov.uk/aab-reports), they have the first facet box expanded by default.  Could we either re-use whatever opens these by default, or disable whatever is closing the first facet on the business finder?

Desktop:
Screen Shot 2019-04-30 at 09 16 51

Mobile:
Screen Shot 2019-04-30 at 09 17 00

ChrisBAshton:
Agreed - I think the 10 option override was added to accommodate an earlier design, and should now be removed. Thanks for the reviews.

app/views/finders/_option_select_facet.html.erb
```diff
@@ -1,12 +1,11 @@
 <% cache_if option_select_facet.cacheable?, option_select_facet.cache_key do %>
-  <% close_facet = option_select_facet.unselected? && ((option_select_facet_counter > 0) || option_select_facet.close_facet?) %>
   <%= render partial: 'components/option-select', locals: {
     :key => option_select_facet.key,
     :title => option_select_facet.name,
     :aria_controls_id => "js-search-results-info",
     :options_container_id => option_select_facet.key,
     :options => option_select_facet.options("js-search-results-info", option_select_facet.key),
-    :closed_on_load => close_facet,
+    :closed_on_load => option_select_facet_counter > 0 && option_select_facet.unselected?,
```
leenagupte: @ChrisBAshton How does this work in AAIB reports? Does changing this conditional affect other finders?

app/models/option_select_facet.rb
leenagupte: I think we might need to keep this, so that facets that contain more than 10 items (i.e. organisations) are closed by default.

app/views/finders/_option_select_facet.html.erb
ChrisBAshton: This conditional is the same as before (https://github.com/alphagov/finder-frontend/pull/1073/files#diff-4d0b8cfd231c9d5050ca9adf67e518d4L2), except there is no call to `close_facet?`, so no check for the number of options.

But this is ok because the 10 option limit is a piece of legacy code that can be removed - see the commit message in 77a0b9370102e44e755bbeed3510f0c410e7783e.

app/models/option_select_facet.rb
ChrisBAshton: Please see https://github.com/alphagov/finder-frontend/pull/1073/files#r279796574

PR #1074, opened 2019-04-29
Fix short name lookup for content id facets

steventux:
https://trello.com/c/krOLVujz/68-update-title-for-employing-eu-citizens-when-ordered-by-topic

Facet display names can be overridden by the optional `short_name` attribute value.
eg. The EU Exit business readiness finder facet _Who you employ_ should display the short name _Employing EU citizens_ when displaying grouped results.

Screenshot from 2019-04-29 15-58-07

[Heroku preview](https://finder-frontend-pr-1074.herokuapp.com/find-eu-exit-guidance-business?parent=&keywords=&employ_eu_citizens%5B%5D=yes&order=topic)

See https://github.com/alphagov/content-tagger/pull/906

This should be supported when the facet type is 'content_id', this
commit adds similar support as the `label_for_metadata_key` method but with reduced complexity.

### DIFF
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -93,7 +93,7 @@ class FinderPresenter
   def facet_details_lookup
     @facet_details_lookup ||= begin
       result_hashes = @facet_hashes.map do |facet|
-        facet_name = facet['name']
+        facet_name = facet.fetch('short_name', facet['name'])
         facet_key = facet['key']
         facet.fetch('allowed_values', []).to_h do |value|
           [value['content_id'], {
--- a/app/lib/facets_from_facet_group_extractor.rb
+++ b/app/lib/facets_from_facet_group_extractor.rb
@@ -15,6 +15,7 @@ private
     facet_details = facet['details']
     {
       'name' => facet_details['name'],
+      'short_name' => facet_details['short_name'],
       'key' => facet_details['key'],
       'display_as_result_metadata' => facet_details['display_as_result_metadata'],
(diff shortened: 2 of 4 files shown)


app/presenters/finder_presenter.rb
```diff
@@ -93,6 +93,8 @@ def email_alert_signup_url
   def facet_details_lookup
     @facet_details_lookup ||= begin
       result_hashes = @facet_hashes.map do |facet|
+        puts ">>>>>>>>>>>>>>>>>>>>>>"
+        puts facet
```
leenagupte: Extraneous `puts`

app/presenters/finder_presenter.rb
```diff
@@ -93,7 +93,7 @@ def email_alert_signup_url
   def facet_details_lookup
     @facet_details_lookup ||= begin
       result_hashes = @facet_hashes.map do |facet|
-        facet_name = facet['name']
+        facet_name = facet.fetch('short_name', facet['name'])
```
leenagupte: I think I've missed something, why can you use `short_name` and not have to use `facet['short_name']`?

app/presenters/finder_presenter.rb
steventux: oops willdo! :+1:

app/presenters/finder_presenter.rb
steventux: `Hash#fetch` allows a second arg as a fallback if the key specified isn't found.
In short this is like doing

```ruby
facet['name'] unless facet.key?('short_name')
```

PR #1075, opened 2019-04-29
Remove keywords by find/replace, not by exploding into words.

barrucadu:
Splitting the query on spaces falls down if we're trying to remove a quoted search phrase which contains spaces.  Rather than make the splitting logic quote-aware, I removed it.

`string.replace(old, new)` only replaces the first occurrence, which is what the previous loop did.

Padding everything with spaces is to avoid needing to check for the boundaries of the string.

---

[Trello card](https://trello.com/c/lUl9kTAQ/676-bug-search-input-text-with-surrounding-quotes-cannot-be-removed-using-a-facet-tag)

### DIFF
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -39,17 +39,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         $input.trigger(onChangeSuppressAnalytics);
       }
       else if (inputType == 'text' || inputType == 'search') {
-        var needle = decodeEntities(removeFilterValue.toString());
-        var splitKeywords = currentVal.split(" ");
-
-        for (var i = 0; i < splitKeywords.length; i++) {
-          if (splitKeywords[i].toString() === needle) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/remove-filter.js
```diff
@@ -39,17 +39,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-          if (splitKeywords[i].toString() === needle) {
-            splitKeywords.splice(i, 1);
-            break;
-          }
-        }
-
-        var newVal = splitKeywords.join(" ").trim();
+        var haystack = ' ' + currentVal.trim() + ' ';
+        var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' ';
+        var newVal = haystack.replace(needle, ' ').replace(/  /g, ' ').trim();
```
bilbof: Nice work here. It may not be clear in the future why this is necessary, so perhaps putting this into a named method  and then calling it here would help? Or perhaps adding a comment / longer commit message giving some context?

PR #1079, opened 2019-04-30
Refactor option-select-spec.js

ChrisBAshton:
## What

- tidies up formatting across the file
- removes repetition through caching variables & hoisting, and through creation of a `optionSelectWithAttrs` function
- removes unused variables and `spyOn` declarations

## Why

Because I worked on this file in #1073 (before taking a different route) and spotted a number of ways we can tidy up the file so that it's easier to work with next time unit tests need to be added or changed.

### DIFF
--- a/spec/javascripts/components/option-select-spec.js
+++ b/spec/javascripts/components/option-select-spec.js
@@ -10,7 +10,7 @@ describe('GOVUK.OptionSelect', function() {
   }
 
   beforeEach(function(){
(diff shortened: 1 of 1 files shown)


andysellick:
@ChrisBAshton looks like we've overlapped a bit... can we catch up?

https://github.com/alphagov/finder-frontend/pull/1080

ChrisBAshton:
Closing in favour of #1080.

PR #1083, opened 2019-05-01
Option select code improvements

andysellick:
This PR:

- removes the IE6/7 specific code on the option select
- fixes a minor styling issue (see below)
- BONUS: removes an unused function from the expander component

Before:

Screen Shot 2019-05-01 at 12 00 01

After:

Screen Shot 2019-05-01 at 12 02 14

Trello card: https://trello.com/c/N16toILy/572-fix-minor-option-select-issues


andysellick:
@injms to let you know that I added another commit, it's a little unrelated but it's such a small change I didn't think it worthy of an entire PR to itself.

PR #1084, opened 2019-05-01
Restore hiding of appear_in_find_eu_exit_guidance_buisness_finder checkbox

brucebolt:
The `appear_in_find_eu_exit_guidance_buisness_finder` field was previously a hidden field in the email alert signup form.  The hiding was inadvertently removed in 9ee0aa6d5.  This restores the hiding of the field and restores the related test.

Behaviour now (not hiding the field):
Screen Shot 2019-05-01 at 13 42 30

Expected behaviour (with hiding of field):
Screen Shot 2019-05-01 at 13 44 01

Trello card: https://trello.com/c/hKKTZdoj/48-update-the-email-signup-link-in-the-finder

### DIFF
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -113,6 +113,6 @@ private
   end
 
   def ignore_facet?(facet_id)
-    %W(facet_groups).include?(facet_id)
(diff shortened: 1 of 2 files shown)


brucebolt:
Turns out that `search-api` needed to be updated.  Closing this as the `appear_in_find_eu_exit_guidance_business_finder` field no longer needs to be hidden.

PR #1086, opened 2019-05-01
update JavaScript to conform to standardjs

ChrisBAshton:
This PR updates all of our JS to conform to [standardjs](https://standardjs.com/) standards - as per our [styleguides/js](https://github.com/alphagov/styleguides/blob/master/js.md#linting) - and adds it as a CI step so that we don't have to remember to run the linter in the future.

Where possible, I've used `standard --fix` to fix errors automatically. I've then manually fixed the other linting errors, ensuring that `bundle exec rake spec:javascript` continues to pass.

Changes to application code:

- Updated references to window functions to explicitly use `window`, e.g. `alert` => `window.alert`.
- Updated some comparisons from `==` to `===`, etc.

Changes to test code:

- Added `/* eslint-disable */` to the top of `spec/javascripts/helpers/jquery-1.12.4.js`. (Used by Jasmine)
- Added `/* eslint-disable */` around multiline strings, which aren't supported in <ES2015 browsers (but we're only using them in our tests, so that's ok).
- ~~There were a number of variables that were initialised but never used. In most cases, I've deleted these, but in [live_site_search_spec.js](https://github.com/alphagov/finder-frontend/pull/1086/files#diff-31069cfc3f323052c2d8115e1c4e324eR215) one of the unused variables looked crucial to the test, so rather than delete, I changed the test slightly. See commit f50199bd03d16e34f5e6587921117268e040ad8c.~~ As of #1099, this no longer applies and I've removed the files.
- ~~There were also functions which were called in some of the tests (e.g. search_spec.js: `var extractedURLs = GOVUK.search.extractSearchURLs($resultsList);`) whose outputs weren't being used, and their removal doesn't break the tests.~~ Ditto.

Changes to both:

- Adds `/* global var1 var2 */` or `/* eslint-env jquery jasmine */` to the top of files where the associated globals exist, so that StandardJS doesn't complain that referenced variables don't exist.


Jenkinsfile
```diff
@@ -4,5 +4,16 @@ library("govuk")
 
 node {
   govuk.setEnvar("PUBLISHING_E2E_TESTS_COMMAND", "test-finder-frontend")
-  govuk.buildProject(sassLint: false, publishingE2ETests: true, brakeman: true, rubyLintDiff: false)
+  govuk.buildProject(
+    beforeTest: {
+      stage("Lint Javascript") {
+        sh("npm i standard")
+        sh("./node_modules/.bin/standard --global it --global expect --global beforeEach --global afterEach --global describe --global \$ --global GOVUK --global jasmine --global spyOn --global jQuery")
```
kevindew: Can you use eslint-env? we already have precedence for this in Content Publisher: https://github.com/alphagov/content-publisher/blob/f1587ab27002c01a45a17b7ded1b627b002b8954/spec/javascripts/components/contextual-guidance-spec.js#L1

It feels a bit fishy passing all those as globals across the entire codebase when most of those are jasmine too.

Jenkinsfile
```diff
@@ -4,5 +4,16 @@ library("govuk")
 
 node {
   govuk.setEnvar("PUBLISHING_E2E_TESTS_COMMAND", "test-finder-frontend")
-  govuk.buildProject(sassLint: false, publishingE2ETests: true, brakeman: true, rubyLintDiff: false)
+  govuk.buildProject(
+    beforeTest: {
+      stage("Lint Javascript") {
+        sh("npm i standard")
```
kevindew: Have you thought about using a package.json file? that's how it's set up in Content Publisher: https://github.com/alphagov/content-publisher/blob/f1587ab27002c01a45a17b7ded1b627b002b8954/package.json and allows version to be kept consistent. The usage of ./node_modules/.bin also seems like a bit of a smell since it's calling into a hidden directory

Yarn is used too since rails has in built support for that over NPM: https://github.com/alphagov/content-publisher/blob/master/Jenkinsfile#L11-L14

Jenkinsfile
ChrisBAshton: Now done in 4478f5355f2c2e48d2c27035dd78b428c65bcc8d. Find it odd that your linked example uses both `eslint-env` and `global` in the same file - was there a reason for that?

I've used `global` as it reads better and seems to be less "vendor locked in" than the alternative, but happy to change if there's a good reason!

app/assets/javascripts/live-site-search.js
```diff
@@ -13,157 +13,159 @@
...
     },
-    popState: function(event){
-      if(event.originalEvent.state){
-        liveSearch.saveState(state);
-        liveSearch.updateResults();
-        liveSearch.restoreCheckboxes();
-        liveSearch.pageTrack();
+    popState: function (event) {
+      if (event.originalEvent.state) {
+        liveSearch.saveState(event.originalEvent.state)
```
andysellick: Can you explain this change please?

app/assets/javascripts/live-site-search.js
```diff
@@ -13,157 +13,159 @@
...
-          filterCount = 0,
-          i, _i;
+    checkFilterLimit: function (e) {
+      var newState = liveSearch.$form.serializeArray()
 
-      for(i=0,_i=newState.length; i<_i; i++){
-        if(newState[i].name.lastIndexOf('filter_', 0) === 0){
-          filterCount = filterCount + 1;
+      var filterCount = 0
+
```
andysellick: Nitpick: probably don't need a newline between every variable declaration.

Jenkinsfile
ChrisBAshton: I did think about that, but as I only want to use standardjs (no other libs, all vanilla application code) it seems an unnecessary overhead. I really just want to `npx standard` across the repo on CI, but `npx` isn't a thing on our Jenkins setup.

Agreed about the code smell, I cried a little writing that.

On your recommendation I'm now using `yarn`, which looks cleaner 😄 I think long term I want to add a PR to https://github.com/alphagov/govuk-jenkinslib so that applications can hook into standardjs linting with a simple `javascriptLint: true`.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -55,38 +54,37 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
          * Just removing " beta " from the haystack would result in
          * "alphagamma", which is wrong.
          */
-        var haystack = ' ' + currentVal.trim() + ' ';
-        var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' ';
-        var newVal = haystack.replace(needle, ' ').replace(/  /g, ' ').trim();
+        var haystack = ' ' + currentVal.trim() + ' '
+        var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
+        var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
```
andysellick: Can you explain this change please?

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -55,38 +54,37 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        $input.val(newVal).trigger(onChangeSuppressAnalytics)
+      } else if (elementType === 'OPTION') {
+        $('#' + removeFilterFacet).val('').trigger(onChangeSuppressAnalytics)
       }
     }
 
-    function getInput(removeFilterName, removeFilterValue, removeFilterFacet) {
-      var selector = (!!removeFilterName) ? " input[name='" + removeFilterName + "']" : " [value='" + removeFilterValue + "']";
+    function getInput (removeFilterName, removeFilterValue, removeFilterFacet) {
+      var selector = (removeFilterName) ? " input[name='" + removeFilterName + "']" : " [value='" + removeFilterValue + "']"
```
andysellick: Can you explain the removal of the `!!` please?

app/assets/javascripts/search.js
```diff
@@ -1,114 +1,121 @@
...
+      search.trackSearchClicks($searchResults)
+      search.trackSearchResultsAndSuggestions($searchResults)
+      search.trackSearchResultsAndSuggestionsOnPageTrack($searchResults)
     },
     buildSearchResultsData: function ($searchResults) {
-      var searchResultData = {'urls': []},
-          searchURLs = search.extractSearchURLs($searchResults),
-          searchSuggestion = search.extractSearchSuggestion();
+      var searchResultData = { 'urls': [] }
+
```
andysellick: Nitpick: we probably don't need a newline between each variable declaration.

app/assets/javascripts/search.js
```diff
@@ -1,114 +1,121 @@
...
-      $searchResults.on('click', 'a', function(e){
-        var $link = $(e.target),
-            sublink = '',
-            position, href, startAt;
+      $searchResults.on('click', 'a', function (e) {
+        var $link = $(e.target)
+
+        var sublink = ''
+
+        var position; var href; var startAt
```
andysellick: This is a bit inconsistent - two var declarations separated by newlines, then three all on one line. Can these be separated into individual lines, then remove the separating newlines?

app/assets/javascripts/search.js
```diff
@@ -1,114 +1,121 @@
...
-          var queryString = window.location.search,
-              startAtRegex = /(&|\?)start=([0-9]+)(&|$)/,
-              startAt = 0,
-              matches;
+        var getStartAtValue = function () {
+          var queryString = window.location.search
 
-          matches = queryString.match(startAtRegex);
+          var startAtRegex = /(&|\?)start=([0-9]+)(&|$)/
+
```
andysellick: Nitpick: we probably don't need a newline between each variable declaration.

app/assets/javascripts/taxonomy-select.js
```diff
@@ -3,68 +3,70 @@
...
+    return options
   }
 
-  TaxonomySelect.prototype.resetSubTaxonValue = function resetSubTaxonValue() {
-    var selected = this.$subTaxon().find(':selected'),
-        parentTaxon = this.$topLevelTaxon().val(),
-        isOrphanedSubTaxon = selected && selected.attr('data-topic-parent') !== parentTaxon;
+  TaxonomySelect.prototype.resetSubTaxonValue = function resetSubTaxonValue () {
+    var selected = this.$subTaxon().find(':selected')
+
```
andysellick: Nitpick: we probably don't need a newline between each variable declaration.

spec/javascripts/components/option-select-spec.js
```diff
@@ -73,382 +74,383 @@ describe('An option select component', function() {
...
-      lastCheckbox = optionSelect.$allCheckboxes[optionSelect.$allCheckboxes.length -1];
-    });
+      optionSelect.setContainerHeight(100)
+      optionSelect.$optionsContainer.width(100)
+      firstCheckbox = optionSelect.$allCheckboxes[0]
+      lastCheckbox = optionSelect.$allCheckboxes[optionSelect.$allCheckboxes.length - 1]
+    })
 
     it('returns true if a label is visible', function () {
-      expect(optionSelect.isCheckboxVisible.call(optionSelect, 0, firstCheckbox)).toBe(true);
```
andysellick: Can you explain this change please?

spec/javascripts/components/option-select-spec.js
```diff
@@ -73,382 +74,383 @@ describe('An option select component', function() {
...
-    });
+      expect(optionSelect.isCheckboxVisible(0, firstCheckbox)).toBe(true)
+    })
 
     it('returns true if a label is outside its container', function () {
-      expect(optionSelect.isCheckboxVisible.call(optionSelect, 0, lastCheckbox)).toBe(false);
```
andysellick: Can you explain this change please?

spec/javascripts/search_spec.js
```diff
@@ -1,97 +1,94 @@
...
+        })
 
         it('assigns the suggestions key when suggestions are present', function () {
           expect(GOVUK.search.buildSearchResultsData($resultsList))
             .toEqual({
               'urls': [],
               'suggestion': 'testing'
-          });
-        });
+            })
```
andysellick: Nitpick: is this indentation correct?

spec/javascripts/search_spec.js
```diff
@@ -1,97 +1,94 @@
...
+        })
         expect(extractedURLs[1]).toEqual({
           href: 'guidance/content-design/research-and-evidence'
-        });
-      });
+        })
+      })
 
       it('fires an event when trackSearchResultsAndSuggestions is called', function () {
-        var extractedURLs = GOVUK.search.extractSearchURLs($resultsList);
```
andysellick: Looks like `extracedURLs` has been removed here, doesn't look like it was used so that's fine, just wanted to mention it as you'd specifically noted the removal of another unused var, but not this one.

app/assets/javascripts/components/option-select.js
```diff
@@ -1,4 +1,5 @@
-window.GOVUK = window.GOVUK || {};
+/* global ieVersion */
```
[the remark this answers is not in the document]
ChrisBAshton: Done 👍

app/assets/javascripts/live-site-search.js
ChrisBAshton: `state` was being flagged as a global, and I figured if we're explicitly checking that `event.originalEvent.state` exists, that's the value we'd want to be using...?

Passing `state` alone probably worked by coincidence, but better to be explicit. A bit like writing `if (window.alert) { alert() }`.

app/assets/javascripts/live-site-search.js
ChrisBAshton: StandardJS does that for us 🤷‍♂ think we'd better leave it. Removing the newlines would be us imposing our own nitpicks on the standards.

app/assets/javascripts/application.js
```diff
@@ -22,24 +22,29 @@
...
 
-jQuery(function($) {
-  var $form = $('.js-live-search-form'),
-      $results = $('.js-live-search-results-block'),
-      $elementsRequiringJavascript = $('.js-required'),
-      $atomAutodiscoveryLink = $("link[type='application/atom+xml']").eq('0');
+jQuery(function ($) {
+  var $form = $('.js-live-search-form')
 
-  $elementsRequiringJavascript.show();
```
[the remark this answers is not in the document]
ChrisBAshton: Doesn't look like it's reordered to me...? It's just split out `var`s onto their own lines.

Jenkinsfile
kevindew: eslint-env only knows about common environment globals: https://eslint.org/docs/user-guide/configuring#specifying-environments whereas GOVUK is one we've invented.

So the file in question is in both a jQuery and jasmine environment and then also uses a GOVUK global variable.

Best reason for eslint-env in the case of jasmine is that you don't need to keep adding to globals each time you use a different one of their functions.

app/assets/javascripts/search.js
ChrisBAshton: See https://github.com/alphagov/finder-frontend/pull/1086#discussion_r280335766

app/assets/javascripts/search.js
ChrisBAshton: See https://github.com/alphagov/finder-frontend/pull/1086#discussion_r280335766

app/assets/javascripts/taxonomy-select.js
ChrisBAshton: See https://github.com/alphagov/finder-frontend/pull/1086#discussion_r280335766

Jenkinsfile
kevindew: >  I think long term I want to add a PR to https://github.com/alphagov/govuk-jenkinslib so that applications can hook into standardjs linting with a simple `javascriptLint: true`.

That is part of a long term plan too. It's raised in a recent RFC: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-100-linting.md (also note the bit about all apps have a package.json).

One thing I think that is missing here is ability to lint locally - I'm not sure if finder frontend has a rake task for that but this is the content publisher one: hhttps://github.com/alphagov/content-publisher/blob/f1587ab27002c01a45a17b7ded1b627b002b8954/lib/tasks/lint.rake#L8

I can imagine npx would have been nicer :) we could always install it onto our boxes but I think that does move us away from a consistent destination.

app/assets/javascripts/application.js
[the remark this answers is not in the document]
andysellick: Oh okay, cool.

app/assets/javascripts/live-site-search.js
andysellick: Oh really? Oh well.

app/assets/javascripts/modules/remove-filter.js
ChrisBAshton: Again, this was a `standard --fix` change. I guess they think it's better to use a regex that counts 2 spaces, rather than to use two spaces itself.

Seems odd that we're doing a 2-space-to-1-space replacement in the first place 🤔 but out of scope of this PR!

app/assets/javascripts/modules/remove-filter.js
ChrisBAshton: `standard --fix` did this. I guess it knows that `removeFilterName` is truthy or falsy, even if it's not explicitly a `boolean`, and so casting it via `!!` is unnecessary...?

app/assets/javascripts/search.js
ChrisBAshton: Absolutely! Frustrating standardjs doesn't handle this for us. I'll make the change 👍

spec/javascripts/components/option-select-spec.js
ChrisBAshton: StandardJS complained about unnecessary usage of `.call` (if the first parameter of `.call` is the same as the object that is calling `.call`, then it's effectively the same as if the object calls the method directly.

i.e.

```js
foo.call(foo, bar, 1)
```

is the same as

```js
foo.bar(1)
```

spec/javascripts/components/option-select-spec.js
ChrisBAshton: See https://github.com/alphagov/finder-frontend/pull/1086#discussion_r280355076

app/assets/javascripts/modules/remove-filter.js
andysellick: Yeah, I don't know. Although I guess if all the tests pass it should be fine, I just wanted to check.

spec/javascripts/search_spec.js
ChrisBAshton: Yep - matches up to the `.toEqual` on line 47. In any case, this is what `standard --fix` did for me.

spec/javascripts/search_spec.js
ChrisBAshton: Yep, I removed this as it wasn't being used, and it didn't fail the tests. I'll edit the PR description to mention this too.

Jenkinsfile
ChrisBAshton: Thanks - makes sense. Have moved over to `eslint-env`.

Jenkinsfile
ChrisBAshton: Thanks, great to know the background! So according to the [implementation steps](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-100-linting.md#implementation), looks like we want `package.json` to handle both `standard` JS and `sass-lint`. I sense the risk that this PR could grow out of scope, so here are the options AFAIK:

1. Go ahead with PR as-is, then open another PR at a later date to implement RFC-100 package solution that handles JS and Sass. Make devs run a globally installed `standard` or rely on CI in the meantime.
2. Get a head start on RFC-100 by adding a package and using this to handle the install of `standard` and expose a local linting task to rake, then open another PR at a later date to extend the package.json to handle Sass linting too. This gives devs a command to run local linting without relying on any globals.
3. Implement RFC-100 in this PR, by adding package.json and tasks for JS and Sass linting, and configure the CI scripts such that Sass linting runs the new way and not via [govuk-lint](https://github.com/alphagov/govuk-lint).

What would you suggest?

kevindew: I'd recommend 2 at this point. 

1 varies away from a defined course, whereas 3 requires going a bit above and beyond and I believe there are still prerequisite steps needed for sass linting.

Jenkinsfile
kevindew: Awesome :+1:

Jenkinsfile
ChrisBAshton: Done 👍

package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "finder-frontend",
+  "description": "Finder Frontend renders search (and other pages) on GOV.UK",
+  "private": true,
+  "author": "Government Digital Service",
+  "license": "MIT",
+  "scripts": {
+    "lint": "standard app/javascripts/{*.js,**/*.js} spec/javascripts/{*.js,**/*.js}",
```
kevindew: should this be app/assets ?

ChrisBAshton: Yes it should! Thanks

app/assets/javascripts/live-site-search.js
injms: Removing the newlines and then running standard with or without `--fix` doesn't flag or fix the change. (Double-checked it was actually running by adding a semicolon.) It seems that this doesn't seem to matter either way.

PR #1095, opened 2019-05-03
Remove unused SearchResultPresenter classes

bilbof:
These classes appear to no longer be used, so we can remove them!


koetsier:
See: https://github.com/alphagov/finder-frontend/pull/1054

PR #1104, opened 2019-05-07
Pass `default_frequency` to email-alert-frontend

ChrisBAshton:
Trello card: https://trello.com/c/f7DWeg4q

Whilst there's nothing stopping us merging this, we may want to wait until we have a working E2E test: https://github.com/alphagov/publishing-e2e-tests/pull/319

## What & why

- Passes `default_frequency` parameter to email-alert-frontend (complementing https://github.com/alphagov/email-alert-frontend/pull/461) so that we can tell the email-alert-frontend to select a different frequency by default.
- Only does this for the Business Finder, through a [check in the signup_presenter.rb](https://github.com/alphagov/finder-frontend/pull/1104/commits/1ef9bbbf5e558d75e58422aa4f88ef8977682b74#diff-913d6f25b5b0d15e67e374c07c896ff4R21). All other finders should pass `nil`, which does not add a `default_frequency` parameter to the signup URL in the first place. Whilst we could explicitly set a default of `immediately` for all other finders, this would change the signup URL and could have unintended side-effects.
- I did have reservations about adding an additional parameter to `EmailAlertSignupAPI` when it's not actually used by the upstream API to determine which subscriber list is returned. However, this particular class is more of a 'wrapper' around the upstream API and is Finder-Frontend specific. Therefore it _does_ seem a fair place to encapsulate the decision around the passing of the `default_frequency` parameter. And it just so happens to be easier to test than putting it in the controller!

### DIFF
--- a/app/helpers/eu_exit_finder_helper.rb
+++ b/app/helpers/eu_exit_finder_helper.rb
@@ -1,0 +1,9 @@
+module EuExitFinderHelper
+  def self.eu_exit_finder?(content_id)
+    content_id == "42ce66de-04f3-4192-bf31-8394538e0734"
+  end
+
+  def self.eu_exit_finder_email_signup?(content_id)
+    content_id == "2818d67a-029a-4899-a438-a543d5c6a20d"
+  end
+end
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -218,7 +218,7 @@ class FinderPresenter
   # FIXME: This should be removed once we have a way to determine
   # whether to display metadata in the finder definition
   def eu_exit_finder?
-    self.content_item['content_id'] == "42ce66de-04f3-4192-bf31-8394538e0734"
+    EuExitFinderHelper.eu_exit_finder? self.content_item['content_id']
   end
 
 private
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -16,6 +16,12 @@ class SignupPresenter
     content_item['title']
   end
 
+  def default_frequency
+    # return 'nil' if NOT the business finder email signup page to avoid `default_frequency` appearing in other URLs
+    # as this may not be expected and could have some side-effects
+    EuExitFinderHelper.eu_exit_finder_email_signup?(@content_item['content_id']) ? 'daily' : nil
+  end
+
   def body
     content_item['description']
   end
--- a/lib/email_alert_signup_api.rb
+++ b/lib/email_alert_signup_api.rb
@@ -1,19 +1,31 @@
 class EmailAlertSignupAPI
-  def initialize(applied_filters:, default_filters:, facets:, subscriber_list_title:, finder_format:)
+  def initialize(applied_filters:, default_filters:, facets:, subscriber_list_title:, finder_format:, default_frequency: nil)
     @applied_filters = applied_filters.deep_symbolize_keys
     @default_filters = default_filters.deep_symbolize_keys
     @facets = facets
     @subscriber_list_title = subscriber_list_title
     @finder_format = finder_format
+    @default_frequency = default_frequency
   end
 
   def signup_url
-    subscriber_list['subscription_url']
+    if @default_frequency
+      add_url_param(subscriber_list['subscription_url'], default_frequency: @default_frequency)
+    else
+      subscriber_list['subscription_url']
+    end
   end
 
 private
 
   attr_reader :applied_filters, :default_filters, :facets, :subscriber_list_title, :finder_format
+
+  def add_url_param(url, param)
+    # this method safely adds a URL parameter using the correct one of '?' or '&'
+    parsed_uri = Addressable::URI.parse(url)
+    parsed_uri.query_values = (parsed_uri.query_values || {}).merge(param)
+    parsed_uri.to_s
+  end
 
   def subscriber_list
     Services.email_alert_api.find_or_create_subscriber_list(subscriber_list_options).dig('subscriber_list')
--- a/spec/lib/email_alert_signup_api_spec.rb
+++ b/spec/lib/email_alert_signup_api_spec.rb
@@ -23,21 +23,25 @@ describe EmailAlertSignupAPI do
   let(:finder_format) {}
   let(:default_frequency) { nil }
 
+  def init_simple_email_alert_api(subscription_url)
+    email_alert_api_has_subscriber_list(
+      "tags" => {},
+      "subscription_url" => subscription_url
+    )
+
+    expect(Services.email_alert_api).to receive(:find_or_create_subscriber_list).with(
+      "tags" => {},
+      "title" => subscriber_list_title,
+    ).and_call_original
+  end
+
   describe "default_attributes" do
     context "no default_attributes or attributes" do
       describe "#signup_url" do
         let(:subscription_url) { "http://gov.uk/email" }
 
         it "returns the url email-alert-api gives back" do
-          email_alert_api_has_subscriber_list(
-            "tags" => {},
-            "subscription_url" => subscription_url
-          )
-
-          expect(Services.email_alert_api).to receive(:find_or_create_subscriber_list).with(
-            "tags" => {},
-            "title" => subscriber_list_title,
-          ).and_call_original
+          init_simple_email_alert_api(subscription_url)
 
           expect(subject.signup_url).to eql subscription_url
         end
@@ -73,33 +77,13 @@ describe EmailAlertSignupAPI do
 
     it "returns the url email-alert-api gives back, and appends the default_frequency param" do
       subscription_url = "http://gov.uk/email/some-subscription"
-
-      email_alert_api_has_subscriber_list(
-        "tags" => {},
-        "subscription_url" => subscription_url
-      )
-
-      expect(Services.email_alert_api).to receive(:find_or_create_subscriber_list).with(
-        "tags" => {},
-        "title" => subscriber_list_title,
-      ).and_call_original
-
+      init_simple_email_alert_api(subscription_url)
       expect(subject.signup_url).to eql "http://gov.uk/email/some-subscription?default_frequency=daily"
     end
 
     it "appends the default_frequency param with an ampersand if other URL parameters exist" do
       subscription_url = "http://gov.uk/email/some-subscription?foo=bar"
-
-      email_alert_api_has_subscriber_list(
-        "tags" => {},
-        "subscription_url" => subscription_url
-      )
-
-      expect(Services.email_alert_api).to receive(:find_or_create_subscriber_list).with(
-        "tags" => {},
-        "title" => subscriber_list_title,
-      ).and_call_original
-
+      init_simple_email_alert_api(subscription_url)
       url_params = Rack::Utils.parse_query(URI.parse(subject.signup_url).query)
       expect(url_params).to eq("foo" => "bar", "default_frequency" => "daily")
     end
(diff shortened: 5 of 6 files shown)


spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -66,6 +71,22 @@
...
+    it "returns the url email-alert-api gives back, and appends the default_frequency param" do
+      subscription_url = "http://gov.uk/email/some-subscription"
+      init_simple_email_alert_api subscription_url
+      expect(subject.signup_url).to eql "http://gov.uk/email/some-subscription?default_frequency=daily"
+    end
+
+    it "appends the default_frequency param with an ampersand if other URL parameters exist" do
+      subscription_url = "http://gov.uk/email/some-subscription?foo=bar"
+      init_simple_email_alert_api subscription_url
+      expect(subject.signup_url).to eql "http://gov.uk/email/some-subscription?foo=bar&default_frequency=daily"
```
brucebolt: There's an assumption here that `default_frequency` will always be the last parameter in the query string.  Whilst that's true at the moment, later refactors may mean that's not always the case.

Maybe `Rack::Utils.parse_query` would be better here, so the parameters can be turned into an array and parsed, e.g. https://stackoverflow.com/a/23923728.

app/presenters/signup_presenter.rb
```diff
@@ -16,6 +16,11 @@ def name
     content_item['title']
   end
 
+  def default_frequency
+    # deliberately return 'nil' if NOT the business finder to avoid `default_frequency` appearing in other signup URLs
+    name == 'Find EU Exit guidance for your business' ? 'daily' : nil
```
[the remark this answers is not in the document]
ChrisBAshton: Thanks, this was a definite code smell! Are those methods documented anywhere?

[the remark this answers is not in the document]
brucebolt: I couldn't see any documentation on this method.  The intention is/was that the method would get removed when the business finder was normalised, or when the business finder is no longer needed.

spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -12,6 +12,7 @@
       facets: facets,
       subscriber_list_title: subscriber_list_title,
       finder_format: finder_format,
+      default_frequency: defined?(default_frequency) ? default_frequency : nil,
```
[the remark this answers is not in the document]
ChrisBAshton: Nope - just tried. We only define `default_frequency` in a couple of the tests, but we don't in the rest of them. Ends up throwing this error:

```
NameError:
        undefined local variable or method `default_frequency' for #<RSpec::ExampleGroups::EmailAlertSignupAPI::WithAMultiFacetFinder::SignupUrl::WithOneChoiceSelectedAndATitlePrefix:0x00007fe35dcf0c18>
```

lib/email_alert_signup_api.rb
```diff
@@ -1,20 +1,35 @@
...
+    else
+      subscriber_list['subscription_url']
+    end
   end
 
 private
 
   attr_reader :applied_filters, :default_filters, :facets, :subscriber_list_title, :finder_format
 
+  def add_url_param(url, param)
```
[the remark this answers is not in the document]
ChrisBAshton: I've now used the `Addressable` library as requested. Tried putting it inline in the `signup_url` method but frankly it looked quite messy, think it's nicer to extract it out to a separate method.

lib/email_alert_signup_api.rb
```diff
@@ -1,20 +1,34 @@
...
+    # this method safely adds a URL parameter using the correct one of '?' or '&'
+    utm_uri = Addressable::URI.parse(url)
```
brucebolt: `utm_uri` is presumably copied from elsewhere?  Could this be renamed to something else, since the `utm` part relates to Google Analytics's campaign tracking functionality.

app/presenters/finder_presenter.rb
```diff
@@ -218,7 +220,7 @@ def all_content_finder?
   # FIXME: This should be removed once we have a way to determine
   # whether to display metadata in the finder definition
   def eu_exit_finder?
-    self.content_item['content_id'] == "42ce66de-04f3-4192-bf31-8394538e0734"
+    EUExitFinderHelper.eu_exit_finder? self.content_item['content_id']
```
brucebolt: Could you make the refactoring of this method into a separate commit?

app/helpers/eu_exit_finder.rb
```diff
@@ -0,0 +1,5 @@
+module EUExitFinderHelper
```
brucebolt: I'd suggest a few changes around this helper:

- Rename `eu_exit_finder.rb` to match the module name (i.e. `eu_exit_finder_helper.rb`, then you won't need the `require eu_exit_finder` lines in the presenters, since Rails autoloads them
- Lowercase `EUExitFinderHelper` to `EuExitFinderHelper`, otherwise the file would have to be named `e_u_exit_finder_helper.rb` as Rails splits on the changing case in the module name
- Replace the method name from `EUExitFinderHelper.eu_exit_finder?` to `eu_exit_finder?` and update the calls in the presenters, then the `eu_exit_finder?` method in the finder presenter could also be removed

spec/lib/email_alert_signup_api_spec.rb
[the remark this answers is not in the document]
brucebolt: Cool, that's my Perl background coming back to haunt me 😬

lib/email_alert_signup_api.rb
ChrisBAshton: Whoops

app/helpers/eu_exit_finder.rb
ChrisBAshton: That last bulletpoint is tricky, as all existing calls to `eu_exit_finder?` rely on the presenter holding the state of the content ID, but my new `eu_exit_finder?` helper takes a content ID as an argument.

Quite happy to keep the existing `eu_exit_finder?` method as a façade for my `eu_exit_finder?` method and change less code.

app/helpers/eu_exit_finder_helper.rb
```diff
@@ -0,0 +1,9 @@
+module EuExitFinderHelper
+  def self.eu_exit_finder?(content_id)
+    content_id == "42ce66de-04f3-4192-bf31-8394538e0734"
```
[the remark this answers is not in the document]
ChrisBAshton: Thanks Steve - do you mean [like this](http://www.chrisrolle.com/en/blog/private-constants-in-ruby):

```ruby
module EuExitFinderHelper
  CONTENT_ID_EU_EXIT_FINDER = "42ce66de-04f3-4192-bf31-8394538e0734"
  CONTENT_ID_EU_EXIT_FINDER_EMAIL_SIGNUP = "2818d67a-029a-4899-a438-a543d5c6a20d"
  private_constant :CONTENT_ID_EU_EXIT_FINDER
  private_constant :CONTENT_ID_EU_EXIT_FINDER_EMAIL_SIGNUP

  def self.eu_exit_finder?(content_id)
    content_id == CONTENT_ID_EU_EXIT_FINDER
  end

  def self.eu_exit_finder_email_signup?(content_id)
    content_id == CONTENT_ID_EU_EXIT_FINDER_EMAIL_SIGNUP
  end
end
```

...or [like this](https://www.honeybadger.io/blog/when-to-use-freeze-and-frozen-in-ruby/):

```ruby
module EuExitFinderHelper
  def self.eu_exit_finder?(content_id)
    content_id == "42ce66de-04f3-4192-bf31-8394538e0734".freeze
  end

  def self.eu_exit_finder_email_signup?(content_id)
    content_id == "2818d67a-029a-4899-a438-a543d5c6a20d".freeze
  end
```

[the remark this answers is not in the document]
steventux: The first way means a string is only created when once when the module is loaded whereas defining it in a method means it's recreated for each call, it's an almost insignificant performance consideration. It's not like the method is called when looping through millions of records.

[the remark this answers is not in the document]
ChrisBAshton: Interesting. I suppose if this _were_ in a big loop then it would be a change worth making. As it is, feels to me like option 1 sacrifices code readability for little benefit! And option 2, while readable, is probably pointless (don't see how anyone could alter the value of the hardcoded string from outside the method)

What do you think? Shall I leave it?

app/presenters/finder_presenter.rb
```diff
@@ -218,7 +218,7 @@ def all_content_finder?
   # FIXME: This should be removed once we have a way to determine
   # whether to display metadata in the finder definition
   def eu_exit_finder?
-    self.content_item['content_id'] == "42ce66de-04f3-4192-bf31-8394538e0734"
+    EuExitFinderHelper.eu_exit_finder? self.content_item['content_id']
```
leenagupte: Do we still need this method? Could we just replace all instances of [`eu_exit_finder?`](https://github.com/alphagov/finder-frontend/search?q=eu_exit_finder%3F&unscoped_q=eu_exit_finder%3F) with a call to the new module?

spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -20,22 +21,27 @@
...
-          email_alert_api_has_subscriber_list(
-            "tags" => {},
-            "subscription_url" => subscription_url
-          )
-
-          expect(Services.email_alert_api).to receive(:find_or_create_subscriber_list).with(
-            "tags" => {},
-            "title" => subscriber_list_title,
-          ).and_call_original
+          init_simple_email_alert_api subscription_url
```
brucebolt: I feel this should be written with parentheses, to make it consistent with the rest of the code in this repo
```suggestion
          init_simple_email_alert_api(subscription_url)
```

spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -66,6 +72,23 @@
     end
   end
 
+  context "when a default_frequency is provided" do
+    let(:default_frequency) { "daily" }
+
+    it "returns the url email-alert-api gives back, and appends the default_frequency param" do
+      subscription_url = "http://gov.uk/email/some-subscription"
+      init_simple_email_alert_api subscription_url
```
brucebolt: As commented earlier, this would be better with parentheses
```suggestion
      init_simple_email_alert_api(subscription_url)
```

spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -66,6 +72,23 @@
...
+      expect(subject.signup_url).to eql "http://gov.uk/email/some-subscription?default_frequency=daily"
+    end
+
+    it "appends the default_frequency param with an ampersand if other URL parameters exist" do
+      subscription_url = "http://gov.uk/email/some-subscription?foo=bar"
+      init_simple_email_alert_api subscription_url
```
brucebolt: As commented earlier, this would be better with parentheses
```suggestion
      init_simple_email_alert_api(subscription_url)
```

spec/lib/email_alert_signup_api_spec.rb
ChrisBAshton: I thought it was Ruby convention to omit parentheses - but I've just done some Googling and perhaps that isn't the case! I agree adding parentheses is clearer 👍

This is the sort of thing that really should be picked up by [govuk-lint](https://github.com/alphagov/govuk-lint) - I'd like to raise a PR there to avoid these conversations reoccurring!

I'll go back and capitalise the commit messages too.

ChrisBAshton: Hmmm adding it to govuk-lint is easier said than done as it breaks things that are in the DSL (e.g. `expect(foo) to(bar)`) and there's [no way of telling what's in the DSL](https://github.com/rubocop-hq/rubocop/issues/714).

Have done it in the repo now though!

PR #1121, opened 2019-05-13
Adds StandardJS linter to Rake task

injms:
StandardJS lints the JavaScript in Jenkins, but not when running `bundle exec rake` locally. 

This pull request adds `npm run lint` (which is an alias listed `package.json`) to the lint rake task so that it's easier to run all checks prior to pushing.

Since StandardJS is installed using npm, the `npm install` command has been added to the `startup.sh` script to make sure that it is installed before anything is run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,13 @@ If you are using the GDS development virtual machine then the application will b
 
 ## Running the test suite
 
-Before you can run the test suite you'll need the [govuk-content-schemas]
-repository locally. See
-[`lib/govuk_content_schema_examples.rb`][content_schema_examples] for more
-details.
+Before you can run the test suite you'll need the [govuk-content-schemas] repository locally. See [`lib/govuk_content_schema_examples.rb`][content_schema_examples] for more details.
+
+The StandardJS JavaScript linter needs to be installed separately before running the default `rake` task. To make sure that it's been installed run:
+
+```sh
+$ yarn install
+```
 
 The default `rake` task runs all the tests:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,10 +54,13 @@ If you are using the GDS development virtual machine then the application will b
 
 ## Running the test suite
 
-Before you can run the test suite you'll need the [govuk-content-schemas]
-repository locally. See
-[`lib/govuk_content_schema_examples.rb`][content_schema_examples] for more
-details.
+Before you can run the test suite you'll need the [govuk-content-schemas] repository locally. See [`lib/govuk_content_schema_examples.rb`][content_schema_examples] for more details.
```
bilbof: Would be cool if we could use the gem https://github.com/alphagov/govuk_schemas instead of requiring another repo locally.

README.md
```diff
@@ -54,10 +54,13 @@ If you are using the GDS development virtual machine then the application will b
...
+
+The StandardJS JavaScript linter needs to be installed separately before running the default `rake` task. To make sure that it's been installed run:
+
+```sh
+$ npm install
```
bilbof: What about using `yarn` here? :D

startup.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+npm install
```
bilbof: `yarn` would be better I suppose. I don't know if that's installed in the VM though..

README.md
injms: That'd be good indeed. I think it needs to be in another pull request as I only removed the line breaks to make the readme formatting a bit more consistent.

README.md
injms: Good point! This version of Rails _should_ support Yarn out of the box. I'll test it and change it.

bilbof: That's awesome, I didn't know that :)

injms: Screenshot showing yarn installing dependencies with four emojis

Well, `yarn` apparently has emoji - so that's also a win too 👍

startup.sh
injms: Done 👍

PR #1123, opened 2019-05-13
Add names to anonymous functions.

injms:
Fixes part of the [frontend technical debt card](https://trello.com/c/l0DjvG2M/124-frontend-technical-debt) to investigate why `live_search.js` creates some anonymous and some named functions.

Both named and anonymous functions are being used throughout the JavaScript files. There's little difference when running, but using named functions should give a better debugging experience and choosing one way will give better consistency.

### DIFF
--- a/app/assets/javascripts/support.js
+++ b/app/assets/javascripts/support.js
@@ -1,6 +1,6 @@
 if (typeof window.GOVUK === 'undefined') { window.GOVUK = {} }
 if (typeof window.GOVUK.support === 'undefined') { window.GOVUK.support = {} }
 
-window.GOVUK.support.history = function () {
+window.GOVUK.support.history = function history () {
   return window.history && window.history.pushState && window.history.replaceState
 }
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/support.js
```diff
@@ -1,6 +1,6 @@
 if (typeof window.GOVUK === 'undefined') { window.GOVUK = {} }
 if (typeof window.GOVUK.support === 'undefined') { window.GOVUK.support = {} }
 
-window.GOVUK.support.history = function () {
+window.GOVUK.support.history = function history () {
```
bilbof: Does this overwrite the `window.history` method?

bilbof: _Is such a thing even possible?_

injms: I couldn't get `window.history` or `history` to be overwritten - even when I tried to ([Codepen sketch](https://codepen.io/injms/pen/BeQEeP?editors=1111)). MDN says it's a read-only property, so I think that this is okay.

bilbof: Good stuff, thanks for checking!

PR #1132, opened 2019-05-16
[DO NOT MERGE] Replace Sass Rails with SassC Rails

### LINKED ISSUE #1124: SassC-Rails upgrade - support CSS `min()` and `max()`
Replacing 
```ruby
gem 'sass-rails', '~> 5.0'
```
with
```ruby
gem 'sassc-rails', '>= 2.0.1'
```
results in errors that (I think) are to do with the conflict between CSS's `max()` and Sass's `max()`. 

[SassC Rails](https://github.com/sass/sassc-rails) uses [SassC Ruby](https://github.com/sass/sassc-ruby), which uses [Libsass](https://github.com/sass/libsass). Unfortunately, Libsass doesn't currently support CSS `min()` and `max()` - [issue 2701](https://github.com/sass/libsass/issues/2701).

Slightly strange is when SassC Rails is used the app runs fine with no errors, but the tests fail. This might be to do with some gems still relying and installing on Ruby Sass, such as GOV.UK Publishing Components ([just updated](https://github.com/alphagov/govuk_publishing_components/pull/853) in [v16.7.0](https://github.com/alphagov/govuk_publishing_components/pull/854)) and the deprecated [GOV.UK Frontend Toolkit](https://github.com/alphagov/govuk_frontend_toolkit).

The error thrown is:
```
Error: "calc(0px)" is not a number for `max'
```

The Sass function also doesn't like having `calc()` added to it, since `calc()` will only work in the browser. Sass can't compute percentages, since they rely on knowing the window width - for example: `calc(100% - 1em)` can only be worked out when you know how big 100% is.

This is triggered by the [width container](https://github.com/alphagov/govuk-frontend/blob/master/src/objects/_width-container.scss#L13) element in GOV.UK Frontend:
```scss
@supports (margin: unquote("max(calc(0px))"))
```
When the library supports CSS min/max, the unquote function ensures that this is output as CSS instead of being processed as a Sass max function.

## Solutions:
 * ~Wait for Libsass support~ (No longer an option as [Libsass is deprecated](https://sass-lang.com/blog/libsass-is-deprecated))
 * Use Dart Sass which [already supports CSS `min` and `max`](https://github.com/sass/dart-sass/pull/428) - but I haven't been able to find a Rails-friendly way of doing that
 * Use [Webpacker](https://www.github.com/rails/webpacker) to use Webpack to compile the CSS. This would mean making sure that all JavaScript and CSS dependencies are JavaScript modules rather than gems. See [Ticket 505 on GOV.UK Publishing Components](https://github.com/alphagov/govuk_publishing_components/issues/505).

injms:
Thhis pull request will swap Sass Rails for SassC Rails. This is because Ruby Sass is [deprecated](http://sass.logdown.com/posts/7828841) and  Sass Rails relied on Ruby Sass.

This replaces Sass Rails with [SassC Rails](https://github.com/sass/sassc-rails/), which uses [SassC Ruby](https://github.com/sass/sassc-ruby), which uses [Libsass](https://github.com/sass/libsass). 

As well as removing a deprecated dependency, it promises a speed increase for the Sass compilation.

Note: there will still be a warning that Ruby Sass is deprecated. This is because GOV.UK Frontend Toolkit is a dependency and it uses Ruby Sass.

Fixes #1124.

- http://finder-frontend-pr-1132.herokuapp.com/search/all
- http://finder-frontend-pr-1132.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1132.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1132.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1132.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem 'rails', '~> 5.2.3'
 gem 'slimmer', '~> 13.1.0'
 
 gem 'govuk_frontend_toolkit', '~> 8.2'
-gem 'sass-rails', '~> 5.0'
+gem 'sassc-rails', '>= 2.1.1'
 gem 'uglifier', '~> 4.1'
 
 group :doc do
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'slimmer', '~> 13.1.0'
 gem 'govuk_frontend_toolkit', '~> 8.2'
 gem 'sassc-rails', '>= 2.1.1'
 gem 'uglifier', '~> 4.1'
+gem 'yui-compressor', '~> 0.12.0'
 
 group :doc do
   # bundle exec rake doc:rails generates the API under doc/api.
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -396,7 +396,7 @@ DEPENDENCIES
   rails (~> 5.2.3)
   rails-controller-testing
   rspec-rails (~> 3.8.2)
-  sass-rails (~> 5.0)
+  sassc-rails (>= 2.1.1)
   sdoc
   simplecov (~> 0.16.1)
   slimmer (~> 13.1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -295,12 +295,6 @@ GEM
     sass-listen (4.0.0)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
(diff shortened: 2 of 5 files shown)


injms:
It looks as this app is not compatible with SassC Rails, so I'm closing this pull request. See #1124 for further information and background.

PR #1134, opened 2019-05-17
Refactor Sass and CSS

andysellick:
This PR restructures finder-frontend's Sass to generally improve readability and structuring but also specifically to solve the problem of most of the CSS being nested underneath the `#finder-frontend` id, which can override any styles set by our components.

As this is clearly a large task, I'm going to approach it in chunks in order to make reviewing the changes easier. This PR is only the beginning of this work.

---

- http://finder-frontend-pr-1134.herokuapp.com/search/all
- http://finder-frontend-pr-1134.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1134.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1134.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1134.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -66,15 +66,15 @@
   }
 }
 
+.metadata-summary {
+  @include govuk-font(19);
+  margin-bottom: govuk-spacing(6);
+}
+
 #finder-frontend {
   margin-bottom: govuk-spacing(8);
 
   header {
-    .summary {
-      margin-bottom: govuk-spacing(6);
-      @include govuk-font(19);
-    }
-
     .logo {
       margin: govuk-spacing(6) 0;
     }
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -56,26 +56,18 @@
...
-      margin: govuk-spacing(6) 0;
-      max-width: $two-thirds;
-    }
+  @include media(tablet) {
+    margin: govuk-spacing(6) 0;
+    max-width: $two-thirds;
   }
+}
 
-  legend {
```
vanitabarrett: There are legends on the Q&A pages, which I think is where the overrides are being used. The difference isn't huge though, so may be fine to remove.

## On live
Screen Shot 2019-05-29 at 09 55 42

## Heroku
Screen Shot 2019-05-29 at 09 55 28

andysellick: Ah, great spot. Looks like finder-frontend is increasing the bottom margin from 15px to 20px. As you say, not a huge difference, and also this legend is part of a component and already styled. The override looks like a more general thing, so I wonder if it used to apply to something but doesn't anymore? I think I'm going to take the decision that this is okay to remove. Thanks!

PR #1139, opened 2019-05-20
Fix flaky JS test

bilbof:
The container height is sometimes 512. For good measure I've made
the expected height 550, so this should ensure the height doesnt
grow way over what is expected.

The error:

```
An option select component initialising when the parent is hidden and data-closed-on-load is true sets the height of the container sensibly when the option select is opened
  Expected 512 to be less than 500.
```

https://trello.com/c/SfmHnIcY/750-fix-flaky-optionselect-test


bilbof:
Possibly the tests are run in random order, but aren't teared down properly?

It might be worth asking @andysellick this when he's back in the office.

PR #1154, opened 2019-05-28
Official Documents Finder

hannako:
This PR allows finder front end to render an Official Documents Finder. 

It displays a pre-filtered list of documents that return `true` for `has_official_document` eg https://www-origin.staging.publishing.service.gov.uk/api/search.json?filter_has_official_document=true
and provides users with the option of filtering on command papers, or act paper only.

[ref: Trello card](https://trello.com/c/6xFJqfAw/724-spin-up-finder-for-content-that-has-official-doc-status-s)

Screen Shot 2019-05-29 at 11 46 41

---

- http://finder-frontend-pr-1154.herokuapp.com/search/all
- http://finder-frontend-pr-1154.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1154.herokuapp.com/search/official-documents
- http://finder-frontend-pr-1154.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1154.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1154.herokuapp.com/find-eu-exit-guidance-business


hannako:
@bilbof thanks bill. My thought was that the duplication was fine until a third custom facet and filter is required and then I should refactor. But am happy to do so now.

bilbof:
I'd remove the duplicated code now, while we're very familiar with it. Future us might forget how this stuff works.

If we can extract the duplicated stuff into additional classes that will make future changes easier.

For example, since the only difference between `OfficialDocumentsFacet` and `ResearchAndStatisticsFacet` is the `filter_hashes`, you could have a method called `filter_hashes` in both of these classes, and have them inherit from `RadioFacetWithRequiredValue` (or something like that), a new class that uses those. If you can think of a way to do this with composition rather than inheritance that would be even better.

I would also see - this might be out of scope for this PR - if `filter_hashes` could give an array of classes rather than hashes. This would make the code less dependent on mutable filter_hashes, and instead on a collection of immutable (e.g.) `Filter` classes.

sihugh:
@hannako there are some linting failures on the build :-(

app/lib/filters/official_documents_filter.rb
```diff
@@ -0,0 +1,7 @@
+module Filters
+  class OfficialDocumentsFilter < RadioFilterForMultipleFields
+    def filter_hashes
+      Filters::OfficialDocumentsHashes.new.call
```
bilbof: You have used inheritance to provide `filter_hashes` here, and composition for the facet classes. It might be worth using composition for both, but can always do this later!

PR #1163, opened 2019-05-30
De-nest finder-frontend styles

andysellick:
Continuing my mission to de-nest all the styles in finder-frontend from underneath a single id.

---

- http://finder-frontend-pr-1163.herokuapp.com/search/research-and-statistics/email-signup
- http://finder-frontend-pr-1163.herokuapp.com/search/all
- http://finder-frontend-pr-1163.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1163.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1163.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1163.herokuapp.com/find-eu-exit-guidance-business

Trello card: https://trello.com/c/geNECHkE/788-restructure-finder-frontend-css-to-not-all-sit-under-a-single-id

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -132,11 +132,6 @@
 #finder-frontend {
   margin-bottom: govuk-spacing(8);
 
-  .search-results-title {
-    border-bottom: 1px solid govuk-colour('black');
-    padding-bottom: govuk-spacing(4);
-  }
-
   .live-search-loading-message {
     margin-bottom: govuk-spacing(4);
     display: none;
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -119,15 +119,6 @@
 #finder-frontend {
   margin-bottom: govuk-spacing(8);
 
-  .help-text {
-    @include govuk-font(14);
-    display:block;
-    padding: govuk-spacing(2)/2 0;
-    @include media(tablet) {
-      padding: govuk-spacing(2) 0;
-    }
-  }
-
   .form {
     @include govuk-font(19);
     @include media(tablet) {
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -56,16 +56,6 @@
   margin-top: govuk-spacing(2);
 }
 
-.signup-content p {
-  margin: govuk-spacing(3) 0;
-  @include govuk-font(19);
-
-  @include media(tablet) {
-    margin: govuk-spacing(6) 0;
-    max-width: $two-thirds;
-  }
-}
-
 .metadata-summary {
   @include govuk-font(19);
   margin-bottom: govuk-spacing(6);
@@ -116,33 +106,31 @@
   }
 }
 
+.signup-choices {
+  @include govuk-font(19);
+  margin-top: 50px;
+
+  @include media(tablet) {
+    max-width: $two-thirds;
+  }
+
+  fieldset.invalid {
+    border-left: solid 4px $govuk-error-colour;
+    padding-left: govuk-spacing(3);
+    margin-left: -govuk-spacing(3) - 4;
+  }
+}
+
+.signup-choices__message {
+  margin-bottom: govuk-spacing(3);
+  padding: govuk-spacing(3);
+  color: $govuk-error-colour;
+  font-weight: bold;
+  border-left: solid 4px $govuk-error-colour;
+}
+
 #finder-frontend {
   margin-bottom: govuk-spacing(8);
-
-  .form {
-    @include govuk-font(19);
-    @include media(tablet) {
-      max-width: $two-thirds;
-    }
-
-    p {
-      max-width: $full-width;
-    }
-
-    fieldset.invalid {
-      border-left: solid 4px $govuk-error-colour;
-      padding-left: govuk-spacing(3);
-      margin-left: -govuk-spacing(3) - 4;
-    }
-
-    .validation-message {
-      color: $govuk-error-colour;
-      font-weight: bold;
-      padding: govuk-spacing(3);
-      margin-left: -govuk-spacing(3) - 4;
-      border-left: solid 4px $govuk-error-colour;
-    }
-  }
 
   .search-results-title {
     border-bottom: 1px solid govuk-colour('black');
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -56,119 +56,86 @@
...
-    }
+  .js-required {
+    display: none;
+  }
 
-    p {
-      max-width: $full-width;
-    }
+  .govuk-select {
+    width: 100%;
```
injms: Would it be better to use the `govuk-!-width-full` class on the element to get the same outcome?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -56,119 +56,86 @@
...
-      border-left: solid 4px $govuk-error-colour;
-    }
+  @include media(tablet) {
+    max-width: $two-thirds;
   }
 
-  .search-results-title {
-    border-bottom: 1px solid govuk-colour('black');
-    padding-bottom: govuk-spacing(4);
+  fieldset.invalid {
```
injms: Could using `.govuk-form-group--error` replace `fieldset.invalid`?

vanitabarrett:
@andysellick Could you get this deployed to the Heroku review app so we can check the changes visually too, please?

app/assets/stylesheets/finder_frontend.scss
andysellick: I don't think we can use this here as it's applying this style to a component. Select components are the width they are, there's no GOVUK Frontend alternative. We could update the component to include this option, that might be worth doing in the future.

andysellick:
@vanitabarrett hmm odd that it hasn't done it automatically. Afraid I don't have heroku access to deploy it (yeah, I know) any chance you could give it a kick?

app/assets/stylesheets/finder_frontend.scss
andysellick: Possibly. TBH I hadn't finished investigating this properly - decided it was better to stop here and get it reviewed while I'm not able to work on it anymore this week. Will make a note to investigate further for the next round.

injms:
Probably another thing for another day - on the [email signup for drugs](https://finder-frontend-pr-1163.herokuapp.com/drug-device-alerts/email-signup) this is the default:

Screen Shot 2019-06-05 at 11 51 37

Should the parent checkbox also be selected?

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -7,46 +7,41 @@
...
-
-      <%= render "govuk_publishing_components/components/button", {
-        text: "Create subscription",
-        margin_bottom: true
-      } %>
+<%= form_tag email_alert_subscriptions_path(subscriber_list_params: subscriber_list_params), class: 'signup-choices' do %>
+  <% if @signup_presenter.hidden_choices.any? %>
+    <%= render partial: 'govuk_publishing_components/components/title', locals: {
+      title: @signup_presenter.name,
+      context: "Email alert subscription"
```
vanitabarrett: I know this isn't your change, but is it worth putting this in a locale file?

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -7,46 +7,41 @@
...
+    } %>
+    <% @signup_presenter.hidden_choices.each do |hidden_choice| %>
+      <%= hidden_field_tag hidden_choice[:name], hidden_choice[:value] %>
+    <% end %>
+  <% elsif @signup_presenter.can_modify_choices? %>
+    <%= render "govuk_publishing_components/components/checkboxes", {
+      name: nil,
+      heading: @signup_presenter.name,
+      hint_text: "Select the email alerts you need.",
```
vanitabarrett: Same comment here about the locale file

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -7,46 +7,41 @@
...
+      items: @signup_presenter.choices_formatted,
+      is_page_heading: true,
+    } %>
+    <% if @error_message.present? %>
+      <p class="signup-choices__message"><%= @error_message %></p>
     <% end %>
+  <% else %>
+    <%= render partial: 'govuk_publishing_components/components/title', locals: {
+      title: @signup_presenter.name,
+      context: "Email alert subscription"
```
vanitabarrett: Same comment here about the locale file

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -7,46 +7,41 @@
...
-  </div>
-</div>
+  <% if @signup_presenter.body %>
+    <div class="govspeak-wrapper">
+      <%= render 'govuk_publishing_components/components/govspeak', content: @signup_presenter.body.html_safe %>
+    </div>
+  <% end %>
+
+  <%= render "govuk_publishing_components/components/button", {
+    text: "Create subscription",
```
vanitabarrett: Same comment here about the locale file

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -71,18 +71,18 @@
   margin-bottom: govuk-spacing(6);
 }
 
+.logo {
```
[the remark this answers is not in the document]
andysellick: Great spot, thanks. Have BEM-ified the class a bit to `finder-logo`.

app/views/finders/show.html.erb
```diff
@@ -77,9 +77,11 @@
 
     <% if finder.related.any? %>
       <div class="related-links govuk-grid-column-one-third">
-        <ul class="finder-results">
```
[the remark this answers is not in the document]
andysellick: Another great spot, thanks. I've added the class back in but changed it to be more obviously js related (so no one else falls into the same trap I did). Appearance should be unaffected.

app/views/email_alert_subscriptions/new.html.erb
andysellick: I agree, but I think that's a change for a separate PR. I think I've spotted more of these, could be a bit of a rabbit hole.

app/views/finders/show.html.erb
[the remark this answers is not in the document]
andysellick: Oop, actually a bit more complicated. Just working on a fix now.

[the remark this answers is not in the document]
andysellick: I've finalised my changes. The `finder-results` class remains as it's being used by the main results lists. I've also added a `js-finder-results` class that is only used by the JS. It's slightly weird that this means this is nested inside itself (see `finders/search_results.html.erb`) but this was technically how it was before. At some point we should investigate exactly how this works and whether the js class needs to be nested like this, but this is outside the scope here.

PR #1164, opened 2019-05-30
Add EmailAlertListTitleBuilder for better displaying of multi-facet subscriptions

oscarwyatt:
Trello: https://trello.com/c/iC2YECj8

## What

Adds an email alert title builder that presents choices in a comma-separated quoted list. This will be applied to email signups where `combine_mode` is set to `or`.

This builder contains overrides that are specific to the Business Finder, as it is the only application we expect to use this for the foreseeable future.

Example output:

> EU Exit guidance for your business or organisation in the following categories: ‘Financial services’, ‘Electronics, parts and machinery’, ‘Sell products or goods in the UK’, ‘Buy products or goods from abroad’, ‘Employing EU citizens’, ‘Trade marks’

## Why

We will be allowing users to sign up to mailing lists picked from **multiple facets** (differing from, say, [/cma-cases](https://www.gov.uk/cma-cases/email-signup), which lets you choose lists from one facet).

By default, such mailing lists would have a dynamically rendered title that is difficult to parse:

> Find EU Exit guidance for your business with Sector / Business Area of Banking, markets and infrastructure, Electronics, Imports, and Retail, Business activity of I sell products or goods in the UK, I buy products or goods from abroad, and I sell products or goods abroad, Employ EU citizens of No, and Intellectual property of Trade marks

In the spirit of [optimising code for deletion](https://work.stevegrossi.com/2016/11/04/optimize-for-deletion/), we've encapsulated this in a separate `EmailAlertListTitleBuilder` class, to avoid adding complex logic to the already difficult to read `EmailAlertTitleBuilder` class. In the same spirit, we've kept the Business Finder specific overrides in the class, rather than attempt to abstract it into say the email signup JSON schema.

---

- http://finder-frontend-pr-1164.herokuapp.com/search/all
- http://finder-frontend-pr-1164.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1164.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1164.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1164.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -73,7 +73,8 @@ private
   end
 
   def subscriber_list_title
-    EmailAlertTitleBuilder.call(
+    title_builder = signup_presenter.combine_mode == "or" ? EmailAlertListTitleBuilder : EmailAlertTitleBuilder
+    title_builder.call(
       filter: applied_filters,
       subscription_list_title_prefix: content.dig('details', 'subscription_list_title_prefix'),
       facets: signup_presenter.choices
--- a/app/lib/email_alert_list_title_builder.rb
+++ b/app/lib/email_alert_list_title_builder.rb
@@ -1,0 +1,63 @@
+class EmailAlertListTitleBuilder
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def initialize(filter:, subscription_list_title_prefix:, facets:)
+    @filter = filter
+    @subscription_list_title_prefix = subscription_list_title_prefix
+    @facets = facets
+  end
+
+  def call
+    join_fragment = "in the following #{'category'.pluralize(selected_facet_choice_names.count)}"
+    "#{subscription_list_title_prefix.strip} #{join_fragment}: #{comma_separated_facet_choices}"
+  end
+
+private
+
+  attr_accessor :filter, :subscription_list_title_prefix, :facets
+
+  def comma_separated_facet_choices
+    selected_facet_choice_names.map { |choice_name| "'#{choice_name}'" }.join(", ")
+  end
+
+  def selected_facet_choice_names
+    @selected_facet_choice_names ||= begin
+      overwritten_filters = filter.map do |selected_facet_id, selected_facet_choices|
+        selected_facet_choices.map do |selected_facet_choice_id|
+          fetch_facet_choice_name(selected_facet_id, selected_facet_choice_id)
+        end
+      end
+      overwritten_filters.flatten
+    end
+  end
+
+  def fetch_facet_choice_name(facet_id, facet_choice_id)
+    fetch_facet_choice_name_from_overrides(facet_id, facet_choice_id) || fetch_facet_choice_name_from_data(facet_id, facet_choice_id)
+  end
+
+  def fetch_facet_choice_name_from_overrides(facet_id, facet_choice_id)
+    facet_name_overrides.fetch(facet_id, {})[facet_choice_id]
+  end
+
+  def fetch_facet_choice_name_from_data(selected_facet_id, selected_facet_choice_key)
+    facets.detect { |facet| facet["facet_id"] == selected_facet_id }
+      .fetch("facet_choices")
+      .detect { |choice| choice["key"] == selected_facet_choice_key }
+      .fetch("topic_name")
+  end
+
+  def facet_name_overrides
+    {
+      "employ_eu_citizens" => {
+        "yes" => "Employing EU citizens",
+        "no" => "Employing non-EU citizens"
+      },
+      "public_sector_procurement" => {
+        "civil-government-contracts" => "Public sector procurement - civil government contracts",
+        "defence-contracts" => "Public sector procurement - defence contracts"
+      }
+    }
+  end
+end
--- a/features/support/document_helper.rb
+++ b/features/support/document_helper.rb
@@ -295,7 +295,9 @@ module DocumentHelper
 
   def content_store_has_business_readiness_email_signup
     finder_fixture = File.read(Rails.root.join('features', 'fixtures', 'business_readiness_email_signup.json'))
-
+    finder_fixture_without_facets = JSON.parse(finder_fixture)
+    finder_fixture_without_facets['details']['email_filter_facets'] = []
+    finder_fixture = finder_fixture_without_facets.to_json
     content_store_has_item('/find-eu-exit-guidance-business/email-signup', finder_fixture)
   end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -107,7 +107,7 @@ GEM
     faraday (0.15.4)
       multipart-post (>= 1.2, < 3)
     ffi (1.11.1)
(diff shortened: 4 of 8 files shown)


app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -72,7 +72,9 @@ def email_alert_signup_api
   end
 
   def subscriber_list_title
-    EmailAlertTitleBuilder.call(
+    title_builder = EmailAlertTitleBuilder
+    title_builder = EmailAlertListTitleBuilder if signup_presenter.combine_mode == "or"
```
brucebolt: Could this be written using a ternary operator?

```suggestion
    title_builder = signup_presenter.combine_mode == "or" ? EmailAlertListTitleBuilder : EmailAlertTitleBuilder
```

features/support/document_helper.rb
```diff
@@ -299,7 +299,9 @@ def content_store_has_business_readiness_finder
 
   def content_store_has_business_readiness_email_signup
     finder_fixture = File.read(Rails.root.join('features', 'fixtures', 'business_readiness_email_signup.json'))
-
+    finder_fixture_without_facets = JSON.parse(finder_fixture)
```
brucebolt: I'm getting slightly confused here as to why the feature test is failing.  If the frontend changes haven't been made, then there should be no facets to select.  In which case, we can leave the tests as they are, then modify to allow selection of facets when the frontend change is made.  Or have I completely misunderstood this commit?

app/lib/email_alert_list_title_builder.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def initialize(filter:, subscription_list_title_prefix:, facets:)
+    @filter = filter
+    @subscription_list_title_prefix = subscription_list_title_prefix
+    @facets = facets
+  end
+
+  def call
+    join_fragment = "in the following #{'category'.pluralize(selected_facet_choice_names.count)}"
```
brucebolt: I like the use of `pluralize` here 👍

app/lib/email_alert_list_title_builder.rb
```diff
@@ -0,0 +1,55 @@
...
+  def comma_separated_facet_choices
+    selected_facet_choice_names.map { |choice_name| "'#{choice_name}'" }.join(", ")
+  end
+
+  def selected_facet_choice_names
+    @selected_facet_choice_names ||= begin
+      overwritten_filters = filter.map do |selected_facet_id, selected_facet_choices|
+        selected_facet_choices.map do |selected_facet_choice_id|
+          overwritten_name = facet_name_overrides.fetch(selected_facet_id, {})[selected_facet_choice_id]
+          overwritten_name || fetch_facet_choice_name(selected_facet_id, selected_facet_choice_id)
```
brucebolt: Could these two lines be simplified into one without any assignment?

```suggestion
          facet_name_overrides.fetch(selected_facet_id, {})[selected_facet_choice_id] || fetch_facet_choice_name(selected_facet_id, selected_facet_choice_id)
```

features/support/document_helper.rb
oscarwyatt: We added filters in `features/fixtures/business_readiness_email_signup.json` so it'll try and display those. They weren't finalised filters (and isn't the point of this PR) so I'm working on another PR to properly include those. Arguably, yes, we could choose to not include the facets here, inject the facets into the relevant tests and in my PR to include the facets, remove those injections.

app/lib/email_alert_list_title_builder.rb
ChrisBAshton: That line looked long and hard to read. Instead, I've got it calling a `fetch_facet_choice_name` method, which in turn calls `fetch_facet_choice_name_from_overrides` or `fetch_facet_choice_name_from_data`. Looks good to me now, please double-check 👍

app/lib/email_alert_list_title_builder.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+      overwritten_filters.flatten
+    end
+  end
+
+  def facet_name_overrides
+    {
+      "employ_eu_citizens" => {
+        "yes" => "Employing EU citizens",
+        "no" => "Employing non-EU citizens"
```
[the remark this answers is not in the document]
ChrisBAshton: Checking with Sharon.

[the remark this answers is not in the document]
ChrisBAshton: We've decided to rename the questions value to "Non-EU citizens". Trello: https://trello.com/c/8iw52MRb

features/support/document_helper.rb
ChrisBAshton: We can either:

a) leave as-is. Once we've added facets to the frontend, remove this hack & the "no facets" assertion. The fixture data is likely to change once the facets are finalised, but not by much.
b) undo our changes to the fixture data and instead hack the facets and `combine_mode` into the fixture data for our tests only (this is likely to have a negative impact on the readability of the tests we've added). Once we've added facets to the frontend, remove the "no facets" assertion, update the fixture data with the finalised facets and finally remove our hacking of extra facets.

Option b feels marginally more correct but is a fair bit of extra work and for no tangible benefit IMO. Let me know what you think 👍

app/lib/email_alert_list_title_builder.rb
```diff
@@ -0,0 +1,62 @@
...
+    facet_name_overrides.fetch(facet_id, {})[facet_choice_id]
+  end
+
+  def fetch_facet_choice_name_from_data(selected_facet_id, selected_facet_choice_key)
+    facet_info = facets.detect { |facet| facet["facet_id"] == selected_facet_id }
+    facet_choice = facet_info["facet_choices"].detect { |choice| choice["key"] == selected_facet_choice_key }
+    facet_choice["topic_name"]
+  end
+
+  def facet_name_overrides
```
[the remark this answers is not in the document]
ChrisBAshton: We did consider that, but felt it would make it harder to delete later on. On the assumption that these overrides won't be necessary forever, and even that this class won't be necessary forever, we can easily delete the snowflakey code here as it's in one place.

If we integrate it properly into the fixture data / content item metadata, it's more coupled into the application as a whole and therefore harder to remove.

It would make _this_ class much simpler, but add potential technical debt to a few other places. On balance we chose not to. Do let me know if you disagree!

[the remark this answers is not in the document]
brucebolt: I see.

Originally, I thought these values would be best placed in https://github.com/alphagov/search-api/blob/master/config/find-eu-exit-guidance-business-email-signup.yml, however this would require `govuk_content_schemas` to be updated.

As you say, this would involve a more substantial process when it comes to removing this finder.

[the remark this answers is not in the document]
ChrisBAshton: 👍 in that case, are you happy to approve this?

brucebolt:
I've fix the typo in the commit message

PR #1166, opened 2019-05-31
Fixes deprecation notice by calling GdsApi::Search

ChrisBAshton:
Otherwise when running tests we see:

```
.............................................................GdsApi::Rummager is deprecated.  Use GdsApi::Search instead.
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -16,7 +16,7 @@ module Services
   end
 
   def self.rummager
-    @rummager ||= GdsApi::Rummager.new(Plek.find("search"))
+    @rummager ||= GdsApi::Search.new(Plek.find("search"))
   end
 
   def self.email_alert_api


app/lib/services.rb
```diff
@@ -16,7 +16,7 @@ def self.cached_content_item(base_path)
   end
 
   def self.rummager
```
DilwoarH: should we rename these too?

ChrisBAshton: Good question! Maybe in another PR? This one is about fixing a warning to tidy up the test output. A future one could be to make the methods accurate.

I notice the search API have decided _not_ to do a find-and-replace because they want to avoid messing up `git blame`: https://github.com/alphagov/search-api#search-api

DilwoarH: Yeah this might be a bigger piece of work

PR #1170, opened 2019-06-04
Fix screenreader text on Research and Statistics finder

injms:
When filtering results using the facets, a screenreader should read out 
- the filters that are being used
- how many results

When using the Research and Statistics finder, when the results being filtered by 'Statistics (upcoming)', 'Statistics (published)', or 'Research' is now announced.

Relates to: 
https://trello.com/c/iyd3UY3z/734-fix-hidden-text-for-screen-readers-on-finders)
https://trello.com/c/okfLseFa/895-hidden-text-for-screen-readers-is-in-the-wrong-place

Testing required in:
- [x] VoiceOver on Mac OS
- [x] VoiceOver on iOS
- [x] NVDA with Mozilla Firefox
- [x] JAWS with Internet Explorer 11

Source: [testing with assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#what-to-test) 

---

- http://finder-frontend-pr-1170.herokuapp.com/search/all
- http://finder-frontend-pr-1170.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1170.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1170.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1170.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/app/models/filters.rb
+++ b/app/models/filters.rb
@@ -4,6 +4,7 @@ module Filters
       [
         {
           'key' => 'upcoming_statistics',
+          'value' => 'upcoming_statistics',
           'label' => 'Statistics (upcoming)',
           'filter' => {
             'release_timestamp' => "from:#{Time.zone.today}",
@@ -12,6 +13,7 @@ module Filters
         },
         {
           'key' => 'published_statistics',
+          'value' => 'published_statistics',
           'label' => 'Statistics (published)',
           'filter' => {
             'content_store_document_type' => %w(statistics national_statistics statistical_data_set official_statistics)
@@ -20,6 +22,7 @@ module Filters
         },
         {
           'key' => 'research',
+          'value' => 'research',
           'label' => 'Research',
           'filter' => {
             'content_store_document_type' => %w(dfid_research_output independent_report research)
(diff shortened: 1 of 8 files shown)


andysellick:
@hannako seems to be a bug with this... changes to the radio buttons for the type of stats/research don't seem to be updating the hidden text - it consistently reads something like this ` 46,922 results that are Statistics (published), sorted by Most viewed `. Can you check?

hannako:
@andysellick in what environment are you seeing that bug? Did you deploy the branch to integration?

app/models/filters.rb
```diff
@@ -4,6 +4,7 @@ def call
       [
         {
           'key' => 'upcoming_statistics',
+          'value' => 'upcoming_statistics',
```
andysellick: Why do we need a `value` attribute that contains the same value as the `key` attribute? Can't we reuse `key`?

injms: There is a reason for this, but I can't remember what it is at the moment. I'll take a look this arvo and see what it is.

hannako: We use `key` in one place and `value` in another. I'll do a little refactor to remove duplication. Thanks 👍

injms: Thank you @hannako !

hannako:
@andysellick I made that change. Look OK?

PR #1174, opened 2019-06-06
Don't facet on things that will be ignored

sihugh:
Piggy backs off https://github.com/alphagov/finder-frontend/pull/1172

Faceting on things that are populated from local registries just adds load to the system for no benefit.

On a search from https://www.gov.uk/search/all we go from a query like:

https://www.gov.uk/api/search.json?count=20&start=0&fields=title%2Clink%2Cdescription%2Cpublic_timestamp%2Cpopularity%2Ccontent_purpose_supergroup%2Cformat%2Cpart_of_taxonomy_tree%2Cmanual%2Corganisations%2Cpeople%2Cworld_locations&q=government&reject_link[]=%2Fsearch%2Fall&facet_manual=1500%2Corder%3Avalue.title&facet_organisations=1500%2Corder%3Avalue.title&facet_people=1500%2Corder%3Avalue.title&facet_world_locations=1500%2Corder%3Avalue.title&filter_all_part_of_taxonomy_tree[]&filter_all_part_of_taxonomy_tree[]

to

https://www.gov.uk/api/search.json?count=20&start=0&fields=title%2Clink%2Cdescription%2Cpublic_timestamp%2Cpopularity%2Ccontent_purpose_supergroup%2Cformat%2Cpart_of_taxonomy_tree%2Cmanual%2Corganisations%2Cpeople%2Cworld_locations&q=government&reject_link[]=%2Fsearch%2Fall&filter_all_part_of_taxonomy_tree[]&filter_all_part_of_taxonomy_tree[]

which is a reduction in response payload from 559kb to 29.5kb, and in response time
from ~650ms to ~250ms (cold) and from ~200ms to ~25ms (warm - ie submitting
the same query to search-api again).

I deployed this to staging for a test - the initial low period is because traffic replay was not hitting finder frontend.  
The high spike is current master, and the low part afterwards is the improvement due to this PR being deployed.
Screenshot 2019-06-07 09 38 31

I'm all for dynamic faceting, but we should only use it when it's useful.

---

- http://finder-frontend-pr-1174.herokuapp.com/search/all
- http://finder-frontend-pr-1174.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1174.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1174.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1174.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/app/lib/search_query_builder.rb
+++ b/app/lib/search_query_builder.rb
@@ -191,8 +191,14 @@ private
   end
 
   def facet_query
-    count_dynamic_facets(facet_params.keys)
-    facet_params.reduce({}) { |query, (k, v)| query.merge("facet_#{k}" => v) }
+    dynamic_facet_params = facets_not_overridden_by_registries(facet_params)
+    count_dynamic_facets(dynamic_facet_params.keys)
+
+    dynamic_facet_params.reduce({}) { |query, (k, v)| query.merge("facet_#{k}" => v) }
+  end
+
+  def facets_not_overridden_by_registries(facet_params)
+    facet_params.reject { |k, _v| Services.registries.all.has_key?(k) }
   end
 
   def count_dynamic_facets(facet_names)
--- a/app/lib/finder_api.rb
+++ b/app/lib/finder_api.rb
@@ -90,15 +90,21 @@ private
   end
 
   def augment_facets_with_dynamic_values(content_item_hash)
-    search_results.fetch("facets", {}).each do |facet_key, facet_details|
-      facet = content_item_hash['details']['facets'].find { |f| f['key'] == facet_key }
+    facets = content_item_hash['details']['facets']
+    return unless facets.present?
 
-      if registries.all.has_key?(facet_key) && facet
-        facet['allowed_values'] = allowed_values_from_registry(facet_key)
-      elsif facet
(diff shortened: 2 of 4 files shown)


app/lib/search_query_builder.rb
```diff
@@ -191,8 +191,14 @@ def base_reject
   end
 
   def facet_query
-    count_dynamic_facets(facet_params.keys)
-    facet_params.reduce({}) { |query, (k, v)| query.merge("facet_#{k}" => v) }
+    dynamic_facet_params = facets_not_overridden_by_registries(facet_params)
+    count_dynamic_facets(dynamic_facet_params.keys)
```
bilbof: Is it possible that all facet key will always be present? We serialize the search form, but I'm not sure if we strip out any unused params (so if you've not selected organisations you'll still see `&organisations[]=&` in the query params).

sihugh: The facets should be independent of the query string I think?  It should just provide options of additional things to filter by rather than filtering the query itself.

bilbof: Ah yeah you are right, I thought `facet_params` came from the query but it looks like it comes from the content item.

sihugh:
Thanks Bill!

PR #1175, opened 2019-06-07
Refactor finder-frontend Sass

andysellick:
The final PR to finish de-nesting styles from underneath a single ID.

Trello card: https://trello.com/c/geNECHkE/788-restructure-finder-frontend-css-to-not-all-sit-under-a-single-id

---

- http://finder-frontend-pr-1175.herokuapp.com/search/all
- http://finder-frontend-pr-1175.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1175.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1175.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1175.herokuapp.com/find-eu-exit-guidance-business


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -324,7 +316,8 @@
 }
 
 .result-info {
-  margin-bottom: govuk-spacing(2);
+  margin: 0 0 govuk-spacing(4) 0;
```
sihugh: Did this override the previous version because it was defined in 
```
#finder-frontend { 
  .filtered-results {	
    .result-info {	
      margin: 0 0 govuk-spacing(4) 0;
}}} ```?

andysellick: Yeah - turns out there were two declarations for `.result-info`. The one with the `margin` declaration was overriding the one with the `margin-bottom` declaration because it was nested under the ID (the other one wasn't). So I replaced that one with this one.

andysellick:
The question over `@import "govuk_publishing_components/all_components";` is a good one, and I can't currently remember the reason we do it, instead of only pulling in the components we use.

I think we do it like this because it's simpler. You don't have to go find the right bits of CSS and JS if you want to use a new component. Also, in theory, if every app pulls in the same CSS then it should be cached between pages (except that's not true is it because it's imported into another file - it would only be true if we were pulling in the components gem CSS as a separate file. Hey we should maybe try that...)

PR #1176, opened 2019-06-07
[DO NOT MERGE] Add A/B test for search-api elasticsearch cluster

barrucadu:
We're going to run elasticsearch 5 and 6 in parallel and gradually switch between them, A/B style.  This will let us monitor machine performance, and any effects in changes to the result ordering.

This PR implements the finder-frontend bit of the test.  We *don't* want to make this live until both clusters are up and running in production and we're confident they're in sync.

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -12,6 +12,16 @@ class FindersController < ApplicationController
 
   # rubocop:disable Metrics/BlockLength
   def show
+    ab_test = GovukAbTesting::AbTest.new(
+      "SearchClusterABTest",
+      dimension: 24601, # todo: talk to a PA
+      allowed_variants: %w(A B),
+      control_variant: 'A'
+    )
+    @requested_variant = ab_test.requested_variant(request.headers)
+    @requested_variant.configure_response(response)
+    @ab_params = { cluster: @requested_variant.variant_name }
+
     respond_to do |format|
       format.html do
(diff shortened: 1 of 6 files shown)


app/controllers/finders_controller.rb
```diff
@@ -11,6 +11,16 @@ class FindersController < ApplicationController
   ATOM_FEED_MAX_AGE = 300
 
   def show
+    ab_test = GovukAbTesting::AbTest.new(
+      "SearchClusterABTest",
+      dimension: 24601, # todo: talk to a PA
```
sihugh: Jean gets everywhere

app/controllers/finders_controller.rb
```diff
@@ -11,6 +11,16 @@ class FindersController < ApplicationController
...
+      allowed_variants: %w(A B),
```
sihugh: We've got some evidence that our method of assigning people to variants is not fool proof (see [A/B/C method](https://docs.google.com/document/d/1ysC8NYkcXIHUDAK0E7Nz-KQ7-TYlIcABt8-ca99QTHo/edit#heading=h.iwqog86sbciv).  We got around the discrepancy by assigning A as the default, B as the control (behaving the same as A) and C as the variant.  I need to work out better ways of explaining this

sihugh: Suganya was the discoverer of this

PR #1184, opened 2019-06-11
Allow atom feeds for all finders

barrucadu:
Currently finder-frontend refuses to render an atom feed if the default sort order isn't suitable.  This means that /search/all.atom returns a 404, even though the route is published.  A feed *is* useless if not sorted by date, but disallowing non-date-ordered finders seems like the wrong solution to me.

We only publish a route for a feed if we want the feed to be served.  Therefore finder-frontend should make the feed work.  This PR achieves that by making finder-frontend use either release (preferred if the finder allows it) or public (fallback) timestamp order.

---

- http://finder-frontend-pr-1184.herokuapp.com/search/all.atom
- http://finder-frontend-pr-1184.herokuapp.com/search/research-and-statistics.atom
- http://finder-frontend-pr-1184.herokuapp.com/drug-device-alerts.atom
- http://finder-frontend-pr-1184.herokuapp.com/find-eu-exit-guidance-business.atom

---

[Trello card](https://trello.com/c/kd6iAV1I/794-allow-atom-feeds-on-the-all-content-finder-s)

### DIFF
--- a/app/controllers/concerns/finder_top_result_ab_testable.rb
+++ b/app/controllers/concerns/finder_top_result_ab_testable.rb
@@ -27,7 +27,7 @@ module FinderTopResultAbTestable
   end
 
   def test_in_scope?
-    content_item.is_finder? && finder.eu_exit_finder?
+    content_item.is_finder? && finder_api && finder.eu_exit_finder?
   end
 
   def show_top_result?
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -10,9 +10,11 @@ class FindersController < ApplicationController
 
   ATOM_FEED_MAX_AGE = 300
 
+  # rubocop:disable Metrics/BlockLength
   def show
     respond_to do |format|
       format.html do
+        finder_api
         @results = results
         @raw_content_item = content_item.as_hash
         @breadcrumbs = fetch_breadcrumbs
@@ -21,6 +23,7 @@ class FindersController < ApplicationController
         @pagination = pagination_presenter
       end
       format.json do
+        finder_api
         if content_item.is_search? || content_item.is_finder?
           render json: json_response
         else
@@ -28,6 +31,7 @@ class FindersController < ApplicationController
         end
       end
       format.atom do
+        finder_api(is_for_feed: true)
         if content_item.is_redirect?
           redirect_to_destination
         else
@@ -39,6 +43,7 @@ class FindersController < ApplicationController
   rescue ActionController::UnknownFormat
     render plain: 'Not acceptable', status: :not_acceptable
   end
+  # rubocop:enable Metrics/BlockLength
 
 private
 
@@ -87,8 +92,12 @@ private
     )
   end
 
-  def finder_api
-    @finder_api ||= finder_api_class.new(content_item.as_hash, filter_params)
+  def finder_api(is_for_feed: false)
+    @finder_api ||= finder_api_class.new(
+      content_item.as_hash,
+      filter_params,
+      override_sort_for_feed: is_for_feed,
+    )
   end
 
   def raw_finder
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -30,11 +30,9 @@ class FindersController < ApplicationController
       format.atom do
         if content_item.is_redirect?
           redirect_to_destination
-        elsif finder.atom_feed_enabled?
+        else
           expires_in(ATOM_FEED_MAX_AGE, public: true)
           @feed = AtomPresenter.new(finder, results, facet_tags)
-        else
-          render plain: 'Not found', status: :not_found
         end
       end
     end
@@ -85,7 +83,6 @@ private
     @finder ||= finder_presenter_class.new(
       raw_finder,
       search_results,
-      sort_presenter,
       filter_params,
     )
   end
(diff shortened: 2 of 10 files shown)


app/controllers/finders_controller.rb
```diff
@@ -30,11 +30,15 @@ def show
...
         else
-          render plain: 'Not found', status: :not_found
+          expires_in(ATOM_FEED_MAX_AGE, public: true)
+          @feed = AtomPresenter.new(finder_for_feed, results_for_feed, facet_tags_for_feed)
+
+          # clobber @finder so that the FinderAnswerABTest doesn't
+          # trigger another call to search-api when it accesses this
+          # variable - all it wants to know is if this is the eu exit
+          # finder.
+          @finder = @finder_for_feed
```
bilbof: As `finder_for_feed` is memoized, could you pass the method rather than the variable here?

app/controllers/finders_controller.rb
```diff
@@ -98,6 +101,35 @@ def raw_finder
...
+        override_sort_for_feed: true
+      )
+
+      raw_finder_for_feed = finder_api_for_feed.content_item_with_search_results
+
+      finder_presenter_class.new(
+        raw_finder_for_feed,
+        finder_api_for_feed.search_results,
+        filter_params,
+      )
```
bilbof: Could you instead do something like this, rather than having separate methods for the feed in the controller?

```
  def finder_api(is_atom_feed: false)
    @finder_api ||= finder_api_class.new(content_item.as_hash, filter_params, override_sort_for_feed: is_atom_feed)
  end
```

bilbof: I think that the AtomPresenter shouldn't need a FinderPresenter, and instead should be able to render search results from the FinderApi alone. But that's a bit out of scope for this PR!

app/lib/finder_api.rb
```diff
@@ -70,6 +76,7 @@ def fetch_search_response(content_item)
     queries = query_builder_class.new(
       finder_content_item: content_item,
       params: filter_params,
+      override_sort_for_feed: override_sort_for_feed,
```
bilbof: If you've already over-written the order argument above with 'most-recent', does `OrderQueryBuilder` also need to know that it is fetching data for an atom feed?

barrucadu: Yes, as the `order` argument is only used when merging batch search results

barrucadu:
This is now blocked by the failing test, the fix for which is blocked by the broken e2e tests...

barrucadu:
Rebased on `master` to hopefully fix the failing `or` test

app/controllers/concerns/finder_top_result_ab_testable.rb
```diff
@@ -27,7 +27,7 @@ def set_test_response_header
   end
 
   def test_in_scope?
-    content_item.is_finder? && finder.eu_exit_finder?
+    content_item.is_finder? && finder_api && finder.eu_exit_finder?
```
bilbof: Why is it necessary to check for `finder_api`?

spec/controllers/finders_controller_spec.rb
```diff
@@ -108,22 +108,21 @@
...
               filter_document_type: "mosw_report",
-              order: "-closing_date",
+              order: "-public_timestamp",
               start: 0
             }
           )
           .to_return(status: 200, body: rummager_response, headers: {})
-      end
 
-      it "returns a 404 when requesting an atom feed, rather than a 500" do
```
bilbof: It might be nice to retain a test for what happens when one requests a finder and it doesn't exist.

app/controllers/concerns/finder_top_result_ab_testable.rb
barrucadu: Here's a comment I wrote and didn't commit, should I add this?

```
    # The finder is requested lazily, if the format is invalid then
    # 'finder_api' will be nil.  Rather than unconditionally request
    # the finder regardless of format (only to then serve an error
    # page), check here if the finder has been initialised before
    # trying to use it.
```

spec/controllers/finders_controller_spec.rb
barrucadu: Do you mean if the content item doesn't exist?

app/controllers/concerns/finder_top_result_ab_testable.rb
bilbof: Does `finder` instantiate `finder_api`?

spec/controllers/finders_controller_spec.rb
bilbof: Ah my bad. The 404 wouldn't reach this application if an atom feed route wasn't published.

app/controllers/concerns/finder_top_result_ab_testable.rb
barrucadu: `finder_api` is instantiated in `show` to the normal finder (in the html and json branches) or the special feedy finder (in the atom branch), but it's not instantiated at all in the `ActionController::UnknownFormat` branch.

There's a test which checks an error is served if you request the finder at an invalid format, which is failing because this AB test tries to use `finder_api`, which is no longer being set.  It seems silly to request the content item from content-store to then immediately discard it and serve an error.

PR #1186, opened 2019-06-11
Disable end-to-end tests as a required status check

brucebolt:
Temporarily removing end-to-end tests for finder-frontend until the tests are fixed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ node {
       }
     },
     sassLint: false,
-    publishingE2ETests: true,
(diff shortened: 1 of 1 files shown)


cbaines:
Given this is more of a GitHub configuration thing (at least the branch protections), it might be easier to change that here [1].

1: https://github.com/alphagov/govuk-saas-config/blob/master/github/lib/configure_repo.rb#L97

brucebolt:
Good point.  I'll raise a PR on `govuk-saas-config`.

PR #1193, opened 2019-06-14
Refresh registries upon deploy

bilbof:
We store data for facets in the local cache (memcached), with 'registry' classes that manage them.

Currently, when a user visits a search page, and the 1 hour cache has expired on a registry, fresh data will be fetched for the registry and added to the cache as part of the same request.

This means that some requests by users are unfairly slow. The first request to a newly started application is especially slow because all of the registry caches will be empty.

This change does the following:

- Adds a rake task to refresh the cache, which will be called during the deploy, so the caches aren't empty when the first request comes in. Call it with `LIVE=true` set if you want to point it at live things.
- Sets `expires_in` to the default, which is 'never': no expiry on the registry caches. Once you add registry data, you must use the rake task to refresh it.
- Adds a healthcheck endpoint, which reports a warning if any caches are empty.

**Note:** Will be merged in after https://github.com/alphagov/govuk-app-deployment/pull/315 and https://github.com/alphagov/govuk-puppet/pull/9282 have been merged/run.

Trello: https://trello.com/c/uwm4BnOT/787

---

- http://finder-frontend-pr-1193.herokuapp.com/search/all
- http://finder-frontend-pr-1193.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1193.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1193.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1193.herokuapp.com/find-eu-exit-guidance-business

### DIFF
--- a/app/lib/registries/full_topic_taxonomy_registry.rb
+++ b/app/lib/registries/full_topic_taxonomy_registry.rb
@@ -1,21 +1,32 @@
 module Registries
   class FullTopicTaxonomyRegistry
-    CACHE_KEY = "#{NAMESPACE}/full_topic_taxonomy".freeze
+    CACHE_KEY = "registries/full_topic_taxonomy".freeze
 
     def [](base_path)
       taxonomy[base_path]
     end
 
+    def refresh_cache
+      Rails.cache.write(CACHE_KEY, taxonomy_hash)
+    rescue GdsApi::HTTPServerError
+      report_error
+      raise RefreshOperationFailed, "Could not update #{CACHE_KEY} cache due to HTTPServerError"
+    end
+
     def taxonomy
-      @taxonomy ||= Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
+      @taxonomy ||= Rails.cache.fetch(CACHE_KEY) do
         taxonomy_hash
       end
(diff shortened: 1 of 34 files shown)


app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -1,22 +1,34 @@
...
     def [](base_path)
       taxonomy[base_path]
     end
 
     def taxonomy
-      @taxonomy ||= Rails.cache.fetch(CACHE_KEY, expires_in: 1.hour) do
-        taxonomy_hash
+      @taxonomy ||= Rails.cache.fetch(cache_key) do
+        cacheable_data
       end
```
barrucadu: As this logic is the same for all the cacheable registries, I wonder if there should be a `fetch_from_cache` method in `CacheableRegistry` which does this.

Then this method would be

```ruby
def taxonomy
  @taxonomy = fetch_fron_cache
```

bilbof: Thanks, that's a good idea. I've added a commit to do this.

PR #1196, opened 2019-06-14
Use component for search results

andysellick:
This PR updates `finder-frontend` to use the [document list component](https://govuk-publishing-components.herokuapp.com/component-guide/document_list) for search results.

Before: 

Screen Shot 2019-06-24 at 13 41 33

After:

Screen Shot 2019-07-16 at 09 56 05

This involves some changes to the UI, specifically:

- [changes to the document list component](https://github.com/alphagov/govuk_publishing_components/pull/917) (done)
- metadata about search results is now output in a `ul` element, not a `dl` (more on this below)
- minor reduction of spacing between search results

## Changes to the code

The `documents` method in `result_set_presenter.rb` has been modified to present the data in the format the document list component expects. This has involved taking most of the logic out of the `_search_results.html.erb` template and putting it into `search_result_presenter.rb`.

## Metadata output

Metadata for search results includes a label and a value. There is a flag for whether the label should be shown (or visually hidden). This was previously done in the template by optionally adding the `govuk-visually-hidden` class to the `dt`, which contained the label.

The document list component outputs metadata in a `ul`, so the change is that both the label and value are inside a `li` but the label can be optionally wrapped in a span with the visually hidden class. I'm happy to discuss whether this is semantically appropriate but the benefit is that when the label appears with the value it guarantees they appear together on the same line, i.e.

Before:

Screen Shot 2019-06-17 at 07 47 44

After:

Screen Shot 2019-06-17 at 07 47 55

Additionally, search results can also output "First published during the <a previous government>" beneath the metadata. This is now also being output by the document list component using the 'subtext' option, but the styling for it is controlled by finder-frontend.

The outputting of this text [was recently fixed in this PR](https://github.com/alphagov/finder-frontend/pull/1198). 

## Highlighted or 'top' search results

Some searches have a top search result that is highlighted. In order to see this you may need to set the parameter `GOVUK-ABTest-FinderAnswerABTest` to `B` in your browser (I use [this browser extension](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj)) and do a keyword search for `eori`.

They look like this:

Screen Shot 2019-06-17 at 08 08 28

This is now being handled with an optional parameter to the document list component to alter the appearance of a search result. The component is also able to output metadata with this result but we currently only show the link and description for these results.

## Grouped results

The code for grouped results has largely remained the same.

##  Other information

If `&debug_score=1` is appended to the parameters of a finder, further information about search results is output beneath them, as shown below. This is used by PAs.

Screen Shot 2019-06-17 at 07 55 52

This is also now being output by the document list component. It is a chunk of markup generated in the presenter and passed to the component as part of the 'subtext' option, but styled by `finder-frontend`.

Supersedes https://github.com/alphagov/finder-frontend/pull/1194

Trello card: https://trello.com/c/DAblviIN/793-investigate-using-more-components-in-finder-frontend

---

## Specific examples to sanity check:

- [A finder with debug](http://finder-frontend-pr-1196.herokuapp.com/search/research-and-statistics?debug_score=1)
- a page with a [highlighted result](http://finder-frontend-pr-1196.herokuapp.com/find-eu-exit-guidance-business?parent=&keywords=eori&order=relevance)
- [a page with `First published by` on it](http://finder-frontend-pr-1196.herokuapp.com/search/all?parent=&keywords=%22every+child+matters%22&level_one_taxon=&manual=&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&order=relevance)

## Other search page examples to sanity check:

- http://finder-frontend-pr-1196.herokuapp.com/search/all
- http://finder-frontend-pr-1196.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1196.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1196.herokuapp.com/find-eu-exit-guidance-business


app/presenters/result_set_presenter.rb
```diff
@@ -77,6 +69,11 @@ def highlight_top_result?
       best_bet?
   end
 
+  def highlight(index)
+    index === 0 && highlight_top_result?
+  end
```
bilbof: If this method is needed only by `SearchResultPresenter`, could we move it in there?

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +8,80 @@ class SearchResultPresenter
...
+      metadata.each do |meta|
+        label = meta[:hide_label] ? "<span class='govuk-visually-hidden'>#{meta[:label]}:</span>" : "#{meta[:label]}:"
+
+        if meta[:is_date]
+          value = "<time datetime='#{meta[:machine_date]}'>#{meta[:human_date]}</time>"
+        else
+          value = meta[:value]
+        end
+
+        component_metadata[meta[:label]] = "#{label} #{value}".html_safe
```
bilbof: Could we use [`sanitize`](https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize) here, rather than `html_safe`?

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +8,80 @@ class SearchResultPresenter
...
+          track_options: {
+            dimension28: @count,
+            dimension29: title
+          }
+        }
+      },
+      metadata: structure_metadata,
+      metadata_raw: metadata,
+      subtext: subtext,
+      highlight: @highlight,
```
bilbof: Using a private `attr_reader :highlight` is preferred over using a variable `@highlight`. See `attr_reader :search_result` below.

koetsier: I never really get why that is.  ```@highlight``` makes me realise this is an instance variable which attr_reader does not.

app/presenters/search_result_presenter.rb
bilbof: It's a shame we had to move this out of the ERB templates. Could use ERB in this case? https://ruby-doc.org/stdlib-2.6.3/libdoc/erb/rdoc/ERB.html#class-ERB-label-Introduction

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +8,80 @@ class SearchResultPresenter
            :es_score,
            to: :search_result
 
-  def initialize(search_result, metadata)
+  def initialize(search_result, metadata, doc_index, doc_count, finder_name, debug_score, highlight)
```
bilbof: This is a few arguments, is it possible we could introduce a new subclass for the metadata? If we already have one, could the additional logic related to `structure_metadata` be handled by the metadata presenter?

If we need to give this many methods, could we add a single hash argument so we don't need to pass primitive types? E.g. `def initialize(result:, metadata, doc_index:, doc_count:, finder_name:, debug_score:, highlight: false)`

spec/presenters/search_result_presenter_spec.rb
```diff
@@ -36,11 +66,14 @@
...
-      expect(hash[:format]).to eql('cake')
-      expect(hash[:es_score]).to eql(0.005)
+
+      expect(hash).to eql(expected_document)
+
+      # expect(hash[:title]).to eql(title)
+      # expect(hash[:link]).to eql(link)
+      # expect(hash[:metadata]).to eql(metadata)
+      # expect(hash[:format]).to eql('cake')
+      # expect(hash[:es_score]).to eql(0.005)
```
bilbof: Should these ones be deleted?

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +8,80 @@ class SearchResultPresenter
...
+
+    component_metadata
+  end
+
+  def subtext
+    published_text = "<span class='published-by'>#{I18n.t('finders.first_published_during')} #{government_name}</span>" if is_historic
+    debug_text = "<span class='debug-results debug-results--link'>#{link}</span>"\
+                 "<span class='debug-results debug-results--meta'>Score: #{es_score || "no score (sort by relevance)"}</span>"\
+                 "<span class='debug-results debug-results--meta'>Format: #{format}</span>" if @debug_score
+    "#{published_text}#{debug_text}".html_safe if published_text || debug_text
```
bilbof: Same here with sanitize.

app/presenters/result_set_presenter.rb
```diff
@@ -28,23 +28,15 @@ def search_results_content
...
-        doc = SearchResultPresenter.new(result, metadata).to_hash
-        if  index === 0 && highlight_top_result?
-          doc[:top_result] = true
-          doc[:summary] = result.truncated_description
-        end
-        {
-          document: doc,
-          document_index: index + 1
-        }
+        SearchResultPresenter.new(result, metadata, index, results.count, finder.name, debug_score, highlight(index)).to_hash
```
koetsier: can we move index and finder.name into document? - document already has finder as a member so should be straightforward to delegate

app/presenters/search_result_presenter.rb
bilbof: I do this because Sandy Metz tells me to 😄. The main reason here is it's consistent in this codebase.

app/presenters/result_set_presenter.rb
andysellick: It's called on line 39 of the same file and depends on another method right next to it, which itself depends on a number of other variables. It seemed tidier to leave it here than trying to pass everything through to `SearchResultPresenter`, but if there's a way of doing that nicely, I'm up for changing it, sure.

app/presenters/search_result_presenter.rb
andysellick: If I read the doc for sanitize right, it strips out HTML tags - but I want to preserve them. Is that right?

andysellick: Yeah, moving the logic out of the ERB seemed like a wrench but actually I think the code is easier to read in a presenter and we shouldn't really have so much complex logic in the view.

app/presenters/search_result_presenter.rb
andysellick: Hmm. Let me know who wins this discussion and I'll go with the victor's preference.

koetsier: Consistency makes sense here yes. Sandi is almost always right, but in this case wrong 😉

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +8,84 @@ class SearchResultPresenter
...
+      highlight_text: highlight_text
     }
   end
 
   def link
     search_result.path
   end
 
+  def structure_metadata
+    component_metadata = {}
```
koetsier: Could use a guard here, i.e. ```return {} unless show_metadata```

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +8,84 @@ class SearchResultPresenter
...
+
+    if show_metadata
+      metadata.each do |meta|
```
koetsier: then this can become an each_with_object,

```
metadata.each_with_object({}) do |meta, component_metadata|
```

app/presenters/result_set_presenter.rb
andysellick: I agree that would be a good idea but it's used in a number of other places and I don't want to start introducing more changes in an already complex PR. Can we look at it in a separate PR?

app/presenters/search_result_presenter.rb
andysellick: The little I understand of this sounds good! Let's catch up offline so you can give me some pointers.

spec/presenters/search_result_presenter_spec.rb
andysellick: Oops, yes.

app/presenters/search_result_presenter.rb
bilbof: Sanitize will preserve your HTML while helping to prevent against XSS code injection by stripping things that aren't whitelisted (e.g. <script>). In contrast, html_safe won't change the string, it just marks it as safe to render as HTML.

andysellick:
> We should also add new unit tests to the search_results_presenter_spec for the new logic.

@bilbof have added.

app/presenters/search_result_presenter.rb
andysellick: I didn't know about this! This is brilliant. Changed, done, thank you.

app/presenters/search_result_presenter.rb
andysellick: I've learned more than one thing today! Changed, thanks!

app/presenters/search_result_presenter.rb
andysellick: I've changed `search_result_presenter` to only have a single hash argument as suggested.

Regarding the metadata, I had a crack at it (it does seem unnecessary to call `metadata = metadata_presenter_class.new(result.metadata).present` right before passing it somewhere else) but I don't know where `metadata_presenter_class` comes from - any idea how I can define that in `search_result_presenter`?

andysellick: Having looked into this a bit deeper, I agree that this needs to be addressed but I'd like to suggest we do so in a separate PR - this change feels potentially big and this is a big PR already.

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +10,76 @@ class SearchResultPresenter
...
   end
 
   def link
     search_result.path
   end
 
+  def structure_metadata
+    return {} unless show_metadata
+
+    if show_metadata
```
koetsier: this conditional seems superfluous because you have the guard at line 52

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +10,76 @@ class SearchResultPresenter
...
+
+  def summary_text
+    @highlight ? search_result.truncated_description : summary
+  end
+
+  def highlight_text
+    I18n.t('finders.most_relevant') if @highlight
+  end
+
 private
```
koetsier: Can we make more of these methods private?

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +10,76 @@ class SearchResultPresenter
...
+          track_action: "#{@finder_name}.#{@index}",
+          track_label: link,
+          track_options: {
+            dimension28: @count,
+            dimension29: title
+          }
+        }
+      },
+      metadata: structure_metadata,
+      metadata_raw: metadata,
```
koetsier: I assume metadata_raw isn't being used by the component? Maybe it is better to expose the raw metadata in a separate method instead of being part of this hash?

app/presenters/search_result_presenter.rb
```diff
@@ -8,29 +10,76 @@ class SearchResultPresenter
...
+    @search_result = data[:search_result]
+    @metadata = data[:metadata]
+    @index = data[:doc_index] + 1
+    @count = data[:doc_count]
+    @finder_name = data[:finder_name]
+    @debug_score = data[:debug_score]
+    @highlight = data[:highlight]
   end
 
   def to_hash
```
koetsier: Maybe we could rename this to govuk_component_hash or something to indicate that this hash is to be used directly in the component

app/presenters/result_set_presenter.rb
```diff
@@ -28,23 +28,15 @@ def search_results_content
       zero_results: total.zero?,
       page_count: documents.count,
       finder_name: finder.name,
-      debug_score: debug_score,
+      debug_score: debug_score
     }
   end
 
   def documents
     @documents ||= begin
```
koetsier: this isn't really a document anymore. Could we rename it to something more suitable?

app/presenters/search_result_presenter.rb
andysellick: Have removed.

app/presenters/search_result_presenter.rb
andysellick: Unfortunately, no. They're all called from somewhere else.

app/presenters/result_set_presenter.rb
andysellick: What is it a list of?

app/presenters/search_result_presenter.rb
andysellick: Correct. It's used by the grouped result presenter. I would like to do this in a different way, but it seems entangled to quite a degree, if you look at the code. Any suggestions welcome, but I wonder if this is something bigger we should tackle in a separate PR?

app/presenters/search_result_presenter.rb
andysellick: Is that important? Is `to_hash` generally reserved for something else?

app/presenters/result_set_presenter.rb
andysellick: We'll look at this in a separate refactor.

app/presenters/search_result_presenter.rb
andysellick: Renamed.

PR #1204, opened 2019-06-18
Add blue header bar for users from topics

injms:
The topic / advanced finder is being retired, but the aesthetic needs to remain.

The blue bar is turned on by the `topic` query string, which has to have the correct `content_id` in it.

For example:

- http://finder-frontend-pr-1204.herokuapp.com/search/research-and-statistics?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0

The all finder does not have a blue bar - even if the `topic` query string is present, it should display as normal.

Card: https://trello.com/c/A6eidvxn/799-implement-blue-topic-page-finder-headers-on-supergroup-finders-when-user-has-come-from-topic-page-m

---

- http://finder-frontend-pr-1204.herokuapp.com/search/all
- http://finder-frontend-pr-1204.herokuapp.com/search/all?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-1204.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1204.herokuapp.com/search/research-and-statistics?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-1204.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1204.herokuapp.com/drug-device-alerts?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-1204.herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-1204.herokuapp.com/find-eu-exit-guidance-business?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0

### DIFF
--- a/app/assets/stylesheets/views/_advanced-search.scss
+++ b/app/assets/stylesheets/views/_advanced-search.scss
@@ -26,3 +26,7 @@
   display: none;
 }
 // scss-lint:enable IdSelector
+
+.topic-finder__taxon-link {
+    @include govuk-font(24, $weight: bold);
+}
--- a/app/assets/stylesheets/views/_advanced-search.scss
+++ b/app/assets/stylesheets/views/_advanced-search.scss
@@ -21,12 +21,3 @@
   max-width: 100%;
   width: auto;
 }
-
-.full-width #global-header-bar {
-  display: none;
-}
-// scss-lint:enable IdSelector
-
-.topic-finder__taxon-link {
-    @include govuk-font(24, $weight: bold);
-}
--- a/app/lib/registries/full_topic_taxonomy_registry.rb
+++ b/app/lib/registries/full_topic_taxonomy_registry.rb
@@ -6,6 +6,19 @@ module Registries
       taxonomy[base_path]
     end
 
+    def find_parent_by_id id
+      topic = taxonomy.select { |_, value|
+        value['content_id'] == id
+      }
+
+      path = topic.keys[0]
+
+      {
+        path: path,
+        title: topic[path]['title'],
+      }
+    end
+
     def taxonomy
       @taxonomy ||= fetch_from_cache
     end
--- a/app/lib/registries/full_topic_taxonomy_registry.rb
+++ b/app/lib/registries/full_topic_taxonomy_registry.rb
@@ -6,17 +6,6 @@ module Registries
       taxonomy[base_path]
     end
 
-    def find_parent_by_id id
-      topic = taxonomy.find { |_, value|
-        value['content_id'] == id
-      }
-
-      title = topic.nil? ? nil : topic[1]['title']
-      path = topic.nil? ? nil : topic[0]
-
-      title.nil? ? nil : { path: path, title: title }
-    end
-
     def taxonomy
       @taxonomy ||= fetch_from_cache
     end
--- a/app/lib/registries/full_topic_taxonomy_registry.rb
+++ b/app/lib/registries/full_topic_taxonomy_registry.rb
@@ -11,10 +11,10 @@ module Registries
         value['content_id'] == id
       }
 
-      title = topic.second['title']
-      path = topic.first
+      title = topic.nil? ? nil : topic[1]['title']
+      path = topic.nil? ? nil : topic[0]
 
-      title.nil? ? nil : {path: path, title: title}
+      title.nil? ? nil : { path: path, title: title }
     end
 
     def taxonomy
--- a/app/lib/registries/full_topic_taxonomy_registry.rb
+++ b/app/lib/registries/full_topic_taxonomy_registry.rb
@@ -7,16 +7,14 @@ module Registries
     end
 
     def find_parent_by_id id
-      topic = taxonomy.select { |_, value|
+      topic = taxonomy.find { |_, value|
         value['content_id'] == id
       }
 
-      path = topic.keys[0]
+      title = topic.second['title']
+      path = topic.first
 
-      {
-        path: path,
-        title: topic[path]['title'],
-      }
+      title.nil? ? nil : {path: path, title: title}
     end
 
     def taxonomy
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -186,6 +186,10 @@ class FinderPresenter
     self.content_item['content_id'] == 'dd395436-9b40-41f3-8157-740a453ac972'
   end
 
+  def topic_finder?
+    values.include?('topic-finder')
+  end
+
   # FIXME: This should be removed once we have a way to determine
   # whether to display metadata in the finder definition
   def eu_exit_finder?
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -187,13 +187,13 @@ class FinderPresenter
   end
 
   def topic_finder?
-    values.include?('parent') && topic_finder_parent_name.present?
+    values.include?('parent') && topic_finder_parent.present?
   end
 
-  def topic_finder_parent_name
-    registry = Registries::FullTopicTaxonomyRegistry.new
-    parent_address = values['parent']
-    registry[parent_address].nil? ? '' : registry[parent_address]['title']
+  def topic_finder_parent
+    registry = Services.registries.all['full_topic_taxonomy']
+    parent = registry.find_parent_by_id values['parent']
+    parent.nil? ? '' : parent
   end
 
   # FIXME: This should be removed once we have a way to determine
--- a/app/presenters/finder_presenter.rb
+++ b/app/presenters/finder_presenter.rb
@@ -188,13 +188,12 @@ class FinderPresenter
   end
 
   def topic_finder?
-    values.include?('parent') && topic_finder_parent.present?
+    values.include?('topic') && topic_finder_parent.present?
   end
 
   def topic_finder_parent
     registry = Services.registries.all['full_topic_taxonomy']
-    parent = registry.find_parent_by_id values['parent']
-    parent
+    registry.find_parent_by_id values['topic']
   end
 
   # FIXME: This should be removed once we have a way to determine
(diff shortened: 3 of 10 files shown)


spec/lib/registries/full_topic_taxonomy_registry_spec.rb
```diff
@@ -48,6 +49,12 @@
       fetched_document = registry[first_level_base_path]
       expect(fetched_document['content_id']).to eq(first_level_content_id)
     end
+
+    it "can look up a taxon from the id" do
+      fetched_document = registry.find_parent_by_id(child_content_id)
+      expect(fetched_document[:path]).to eq(child_base_path)
+      expect(fetched_document[:title]).to eq(taxon_title)
+    end
```
bilbof: Would it be possible to add a test to the "when topic taxonomy API is unavailable" context block as well?

app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -6,6 +6,21 @@ def [](base_path)
...
+
+      path = topic.keys[0]
+
+      title = topic.dig(path, 'title')
+
+      {
+        path: path,
+        title: title,
+      }
+    end
```
bilbof: This returns `{:path=>nil, :title=>nil}` if the id isn't found, is that correct?

bilbof: This looks good. I'd recommend `find` since that's guaranteed to return only one item, whereas [`select`](https://apidock.com/ruby/Hash/select) could return multiple.

Maybe something like this?

```
path, topic = taxonomy.find { |_, value| value['content_id'] == id } || nil, {}
title = topic.dig(path, 'title')
{ path: path, title: title }
```

app/presenters/finder_presenter.rb
```diff
@@ -186,6 +186,16 @@ def all_content_finder?
...
   end
 
+  def topic_finder?
+    values.include?('parent') && topic_finder_parent.present?
+  end
+
+  def topic_finder_parent
+    registry = Services.registries.all['full_topic_taxonomy']
+    parent = registry.find_parent_by_id values['parent']
+    parent.nil? ? '' : parent
```
bilbof: I'd return `parent` here. If it's nil then calling `present?` on it will return false anyway.

bilbof:
Great work! There are some things that need looking at:

- CSS for the phase banner and the from field
- displaying the path of the finder at http://finder-frontend-pr-1204.herokuapp.com/search/research-and-statistics?parent=/hm-treasury

We should probably validate the parent param is a topic page before displaying the banner, as `?parent=/hm-treasury` displays a blue banner.

I wonder if it'd be possible to add some more tests for the expected behaviour. A feature test might be nice to ensure that we'll render the banner as expected.

app/assets/stylesheets/views/_advanced-search.scss
```diff
@@ -26,3 +26,7 @@
   display: none;
 }
 // scss-lint:enable IdSelector
+
+.topic-finder__taxon-link {
```
andysellick: I think this needs removing? It's already in `finder_frontend.scss`.

andysellick:
@injms the image above is from this URL - http://finder-frontend-pr-1204.herokuapp.com/search/services?parent=&keywords=&parent=/education&level_one_taxon=c58fdadd-7743-46d6-9629-90bb3ccc4ef0&level_two_taxon=&order=most-viewed

app/lib/registries/full_topic_taxonomy_registry.rb
injms: Good shout on both points. I've used the find suggested and fixed the method to return `nil` if there's nothing, otherwise it returns the hash - this also makes the check for the topic more robust.

app/presenters/finder_presenter.rb
injms: 👍

app/assets/stylesheets/views/_advanced-search.scss
injms: Good spot - it's been removed.

app/presenters/finder_presenter.rb
```diff
@@ -186,6 +187,16 @@ def all_content_finder?
...
   end
 
+  def topic_finder?
+    values.include?('parent') && topic_finder_parent.present?
+  end
+
+  def topic_finder_parent
+    registry = Services.registries.all['full_topic_taxonomy']
+    parent = registry.find_parent_by_id values['parent']
+    parent
```
koetsier: you can omit the 'parent' variable here

injms: ✅

app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -6,6 +6,17 @@ def [](base_path)
...
 
+    def find_parent_by_id id
+      topic = taxonomy.find { |_, value|
+        value['content_id'] == id
+      }
+
+      title = topic.second['title']
+      path = topic.first
+
+      title.nil? ? nil : {path: path, title: title}
```
koetsier: ```Find``` is a linear search and not very efficient - would it make sense to construct a second hash indexing on content_id?

bilbof: Would we cache this second hash?

spec/lib/registries/full_topic_taxonomy_registry_spec.rb
injms: 👍 now added

app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -6,6 +6,17 @@ def [](base_path)
       taxonomy[base_path]
     end
 
+    def find_parent_by_id id
```
bilbof: We can remove this once https://github.com/alphagov/finder-frontend/pull/1224 is merged.

app/presenters/finder_presenter.rb
```diff
@@ -186,6 +187,15 @@ def all_content_finder?
     self.content_item['content_id'] == 'dd395436-9b40-41f3-8157-740a453ac972'
   end
 
+  def topic_finder?
```
bilbof: I would use or move the `i_am_a_topic_page_finder` method here.

```
def i_am_a_topic_page_finder
   values[:topic].present? && Services.registries.all['full_topic_taxonomy'].taxonomy_registry.taxonomy.key? values[:topic]
end
```

app/views/finders/show.html.erb
```diff
@@ -16,86 +16,32 @@
...
-  <%= render 'govuk_publishing_components/components/breadcrumbs', breadcrumbs: @breadcrumbs.breadcrumbs %>
-<% else %>
-  <%= render 'govuk_publishing_components/components/contextual_breadcrumbs', content_item: @raw_content_item, prioritise_taxon_breadcrumbs: finder.eu_exit_finder?  %>
-<% end %>
-
-<% if finder.government? %>
-  <%= render 'govuk_publishing_components/components/government_navigation', active: finder.government_content_section %>
+<% if finder.topic_finder? && !finder.all_content_finder? %>
+  <% content_for :body_classes, "full-width" %>
+  <% inverse = true %>
```
bilbof: If possible can we move this variable into the finder_presenter? It's a bit unclear where this was defined. Alternatively you could reuse a method with `inverse: finder.i_am_a_topic_page_finder` where it's used :)

app/views/finders/show.html.erb
```diff
@@ -16,86 +16,32 @@
...
-        <%= image_tag finder.logo_path, class: "finder-logo__image" %>
-      </div>
+  <% if finder.topic_finder? && !finder.all_content_finder? %>
+    <%= render "govuk_publishing_components/components/inverse_header", {
+      full_width: true
+    } do %>
+      <%= render partial: 'show_header', locals: { inverse: inverse } %>
     <% end %>
+  <% else %>
+    <%= render partial: 'show_header', locals: { inverse: inverse } %>
```
bilbof: It seems like this line is duplicated above. Is there a way we can take this out of the if/else statement?

bilbof: Not easy to do, lets skip it!

app/presenters/finder_presenter.rb
injms: The `i_am_a_topic_finder` seems to use `?parent=` as opposed to `?topic=` as a parameter in the URL - so I'm hesitant to either copy or alter it, as it seems to have subtly different usage.

app/lib/registries/full_topic_taxonomy_registry.rb
injms: ✅

app/lib/registries/full_topic_taxonomy_registry.rb
injms: This is fixed by #1224.

app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -6,6 +6,17 @@ def [](base_path)
...
+      topic = taxonomy.find { |_, value|
```
koetsier: you can use
```key, value = hash.find ... ```

File: app/lib/registries/full_topic_taxonomy_registry.rb [hunk 12]
```diff
@@ -6,6 +6,17 @@ def [](base_path)
...
+        value['content_id'] == id
+      }
+
+      title = topic.nil? ? nil : topic[1]['title']
```
koetsier: maybe use a guard here to avoid the two comparisons, i.e.
```return nil if topic.nil?```

bilbof: We'll be removing this method, it's no longer needed.

PR #1211, opened 2019-06-21
Update sector list and business activity question for readiness finder

brucebolt:
Flattens and updates the wording of the business activity question, public sector contracts question and sectors for the EU Exit Business Readiness Finder.

Details of changes requested: https://docs.google.com/document/d/1xdEkrkN4OZFLsN6xWTBLWPgmXaDTMxp4PxSLqqX0wQ4/edit#
https://docs.google.com/document/d/1NWADDxTfn3aaTGPYD8hPJ-vZfPg0fpPDcDdo2gur3FQ/edit

Also updates the fixtures, so the tests are not tied to the configuration used by the actual finder.  This meant we had to update the tests every time the the configuration was updated.

Before:
Screen Shot 2019-06-21 at 10 21 17

After (also with changes in https://github.com/alphagov/content-tagger/pull/941):
Screen Shot 2019-06-21 at 13 30 01

Merge and deploy at same time as https://github.com/alphagov/content-tagger/pull/941

Trello card: https://trello.com/c/VTKL8TXk

### DIFF
--- a/app/controllers/qa_controller.rb
+++ b/app/controllers/qa_controller.rb
@@ -79,6 +79,11 @@ private
   end
   helper_method :single_wrapped_question?
 
+  def multiple_grouped_question?
+    question_type == "multiple_grouped"
+  end
+  helper_method :multiple_grouped_question?
+
   def multiple_question?
     question_type == "multiple"
   end
--- a/app/views/qa/show.html.erb
+++ b/app/views/qa/show.html.erb
@@ -19,7 +19,7 @@
         } %>
       <% end %>
 
-      <% if multiple_question? %>
+      <% if multiple_grouped_question? %>
         <% filter_groups.each do |filter_group| %>
           <%= render "govuk_publishing_components/components/checkboxes", {
             name: "#{current_facet['facet']['key']}[]",
@@ -28,6 +28,13 @@
             items: nested_options(filter_group)
           } %>
         <% end %>
+      <% elsif multiple_question? %>
+        <%= render "govuk_publishing_components/components/checkboxes", {
+          name: "#{current_facet['facet']['key']}[]",
+          heading: "Select all that apply",
+          visually_hide_heading: true,
+          items: options
+        } %>
       <% elsif single_wrapped_question? %>
         <% checkboxes = render "govuk_publishing_components/components/checkboxes", {
           name: "#{current_facet['facet']['key']}[]",
--- a/features/fixtures/prepare_business_uk_leaving_eu.yaml
+++ b/features/fixtures/prepare_business_uk_leaving_eu.yaml
@@ -1,0 +1,149 @@
+---
+base_path: "/prepare-business-uk-leaving-eu"
+finder_base_path: "/find-eu-exit-guidance-business"
+title: "Prepare your business or organisation for the UK leaving the EU"
+pages:
+  sector_business_area:
+    show_in_qa: true
+    question_type: "multiple_grouped"
+    question: "What does your business or organisation do?"
+    description: |
+      <p>Choose all the sectors and areas that your business or organisation operates in.</p>
+    filter_groups:
+      - name: "Advanced manufacturing and services"
+        filters:
+          - "aerospace"
+          - "automotive"
+          - "electronics-parts-machinery"
+          - "installation-servicing-repair"
+          - "marine"
+          - "rail"
+          - "space"
+          - "other-advanced-manufacturing"
+      - name: "Agriculture, animals, fisheries and forestry"
+        filters:
+          - "agriculture"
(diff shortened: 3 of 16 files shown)


app/views/qa/show.html.erb
```diff
@@ -28,6 +28,13 @@
             items: nested_options(filter_group)
           } %>
         <% end %>
+      <% elsif multiple_question? %>
+        <%= render "govuk_publishing_components/components/checkboxes", {
+          name: "#{current_facet['facet']['key']}[]",
+          heading: "Select all that apply",
+          visually_hide_heading: true,
```
ChrisBAshton: This looks as though it's going to render something like:

```html
<legend class="visually-hidden">Select all that apply</legend>
<span>Select all that apply</span>
```

...which would mean reading out "Select all that apply, Select all that apply" on screen readers.

I think what we actually want is for the `legend` to be "Does your business do any of the following activities?".

If you set `heading: "Does your business do any of the following activities?"` for the checkboxes, we'd end up repeating the page title. But it looks as though you can use that as the `h1` by setting `is_page_heading: true`.

https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/with_legend_as_page_heading

features/qa.feature
```diff
@@ -25,6 +25,13 @@ Feature: QA
       When I submit my answer
       Then my options are persisted as url params
 
+    Scenario: Answering a multiple_grouped option question
+      Given I am answering a multiple_grouped answer question
+      Then I should see a collection of checkboxes
+      When I select multiple checkboxes
+      When I submit my answer
```
ChrisBAshton:
```suggestion
      And I submit my answer
```

app/controllers/qa_controller.rb
```diff
@@ -13,7 +13,8 @@ def show
 private
 
   def qa_config
-    @qa_config ||= YAML.load_file("lib/#{request.path.tr('-', '_')}.yaml")
+    qa_config_path = Rails.application.config.qa_config_path || "lib"
```
leenagupte: What about making this method public instead, and then stubbing it in the tests? Or adding `config.qa_config_path` to all of the environments?

features/fixtures/prepare_business_uk_leaving_eu.yaml
```diff
@@ -94,7 +94,7 @@ pages:
           - "clothing-consumer-goods"
           - "food-drink-tobacco"
           - "motor-trades"
-          - retail
+          - "retail"
```
leenagupte: Can you fix this up to the previous commit?

app/views/qa/show.html.erb
brucebolt: This is the case on the live site too, e.g. https://www.gov.uk/prepare-business-uk-leaving-eu?page=2.  It should probably be fixed as a separate PR.

Screen Shot 2019-06-25 at 09 04 19

app/controllers/qa_controller.rb
brucebolt: Turns out a private method can also be stubbed, so I've stubbed it and removed the config.

app/views/qa/show.html.erb
ChrisBAshton: Have raised https://trello.com/c/V5zTnqDY/181-fix-qa-screen-reader-accessibility.

features/support/document_helper.rb
```diff
@@ -13,6 +13,14 @@ module DocumentHelper
 
   SEARCH_ENDPOINT = "#{Plek.current.find('search')}/search.json".freeze
 
+  def stub_business_finder_qa_config
+    allow_any_instance_of(QaController).to receive(:qa_config).and_return(business_finder_qa_config)
+  end
+
+  def business_finder_qa_config
+    @business_finder_qa_config ||= YAML.load_file('./features/fixtures/prepare_business_uk_leaving_eu.yaml')
+  end
```
ChrisBAshton: Is this additional code necessary or can we call the equivalent code in qa_helper.rb?

leenagupte: Perhaps they could be consolidated in [FixturesHelper](https://github.com/alphagov/finder-frontend/blob/66270f78f02810a084705bae886a1f3bcf28644c/spec/support/fixtures_helper.rb)?

features/qa.feature
```diff
@@ -25,6 +25,13 @@ Feature: QA
       When I submit my answer
       Then my options are persisted as url params
 
+    Scenario: Answering a multiple_grouped option question
+      Given I am answering a multiple_grouped answer question
```
leenagupte: Where is this step added?

PR #1218, opened 2019-06-25
Change heading to reflect question

DilwoarH:
### Background
The Q&A questions have 'Select all that apply' twice in the code - although one is visually hidden - meaning that for screen readers, 'Select all that apply' is read out twice.

Example: https://www.gov.uk/prepare-business-uk-leaving-eu?page=2

```
<h1>Do you sell goods in the UK, import or do business abroad?</h1>
<legend class="visually-hidden">Select all that apply</legend>
<span>Select all that apply</span>
```

The [checkboxes component supports using a heading as a label](https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/with_legend_as_page_heading), as does [the radio component](https://govuk-publishing-components.herokuapp.com/component-guide/radio/with_page_heading), which is what we should do in this case.

This means the markup would become something like:

```
<legend class="visually-hidden"><h1>Do you sell goods in the UK, import or do business abroad?</h1></legend>
<span>Select all that apply</span>
```

Care needs to be taken to apply the fix to radio-based questions AND checkbox-based questions.

Note that for the checkbox-based questions, we're waiting for https://github.com/alphagov/finder-frontend/pull/1211 to be merged.

Trello
https://trello.com/c/V5zTnqDY/181-fix-qa-screen-reader-accessibility

---

- http://finder-frontend-pr-1218.herokuapp.com/prepare-business-uk-leaving-eu?sector_business_area%5B%5D=electronics-parts-machinery&page=2

### DIFF
--- a/app/views/qa/show.html.erb
+++ b/app/views/qa/show.html.erb
@@ -5,23 +5,26 @@
   <% end %>
 <% end %>
 
+<%
+  formatted_description = nil
+  if current_facet["description"]
+    formatted_description = render("govuk_publishing_components/components/govspeak", {
+      content: current_facet["description"].html_safe
+    })
+  end
+%>
+
 <div class="govuk-width-container">
   <%= render 'govuk_publishing_components/components/breadcrumbs', breadcrumbs: breadcrumbs %>
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <form action="<%= next_page_url %>" method="get" id="finder-qa-facet-filter-selection" class="gem-c-title" data-module="track-qa-choices">
-        <%= render "govuk_publishing_components/components/title", {
-          title: current_facet["question"],
-        } %>
-
-        <% if current_facet["description"] %>
-          <%= render "govuk_publishing_components/components/govspeak", {
-            content: current_facet["description"].html_safe
-          } %>
-        <% end %>
-
+      <form action="<%= next_page_url %>" method="get" id="finder-qa-facet-filter-selection" data-module="track-qa-choices">
         <% if multiple_grouped_question? %>
+          <%= render "govuk_publishing_components/components/title", {
+            title: current_facet["question"],
+          } %>
+          <%= formatted_description %>
           <% filter_groups.each do |filter_group| %>
             <%= render "govuk_publishing_components/components/checkboxes", {
               name: "#{current_facet['facet']['key']}[]",
@@ -31,38 +34,54 @@
             } %>
           <% end %>
         <% elsif multiple_question? %>
-          <%= render "govuk_publishing_components/components/checkboxes", {
-            name: "#{current_facet['facet']['key']}[]",
-            heading: "Select all that apply",
-            visually_hide_heading: true,
-            items: options
-          } %>
+          <div class="govuk-!-margin-top-8">
+            <%= render "govuk_publishing_components/components/checkboxes", {
+              name: "#{current_facet['facet']['key']}[]",
+              heading: current_facet["question"],
+              description: formatted_description,
+              hint_text: current_facet["hint_title"],
+              is_page_heading: true,
+              items: options
+            } %>
+          </div>
         <% elsif single_wrapped_question? %>
           <% checkboxes = render "govuk_publishing_components/components/checkboxes", {
             name: "#{current_facet['facet']['key']}[]",
-            heading: "Select all that apply",
+            heading: current_facet["question"],
             visually_hide_heading: true,
             items: options
           } %>
-          <%= render "govuk_publishing_components/components/radio", {
-            name: "#{current_facet['facet']['key']}-yesno",
-            items: [
-              {
-                value: "yes",
-                text: "Yes",
-                conditional: checkboxes
-              },
-              {
-                value: "no",
-                text: "No"
-              }
-            ]
-          } %>
+          <div class="govuk-!-margin-top-8">
+            <%= render "govuk_publishing_components/components/radio", {
+              name: "#{current_facet['facet']['key']}-yesno",
+              heading: current_facet["question"],
+              description: formatted_description,
+              hint_text: current_facet["hint_title"],
+              is_page_heading: true,
+              items: [
+                {
+                  value: "yes",
+                  text: "Yes",
+                  conditional: checkboxes
+                },
+                {
+                  value: "no",
+                  text: "No"
+                }
+              ]
+            } %>
+          </div>
         <% else %>
-          <%= render "govuk_publishing_components/components/radio", {
-            name: "#{current_facet['facet']['key']}[]",
-            items: options
-          } %>
+          <div class="govuk-!-margin-top-8">
+            <%= render "govuk_publishing_components/components/radio", {
+              name: "#{current_facet['facet']['key']}[]",
+              heading: current_facet["question"],
+              description: formatted_description,
+              hint_text: current_facet["hint_title"],
+              is_page_heading: true,
+              items: options
+            } %>
+          </div>
         <% end %>
         <% filtered_params.each_pair do |key, value| %>
           <% if value.is_a?(Array) %>


app/views/qa/show.html.erb
```diff
@@ -31,7 +31,7 @@
       <% elsif single_wrapped_question? %>
         <% checkboxes = render "govuk_publishing_components/components/checkboxes", {
           name: "#{current_facet['facet']['key']}[]",
-          heading: "Select all that apply",
+          heading: current_facet["question"],
```
ChrisBAshton: Doesn't this repeat the heading twice?

What the docs seem to prefer is using the heading _as_ the legend, rather than just repeating it, if that makes sense.

DilwoarH: @ChrisBAshton - which docs are you referring to?

ChrisBAshton: https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/with_legend_as_page_heading

leenagupte: What's the value in having the visually-hidden class in this example?

DilwoarH: @ChrisBAshton - we cannot use this as the question is in a title component ( https://github.com/alphagov/finder-frontend/blob/29717a425b8a480738bdbbda51bba2fc2ed6ca83/app/views/qa/show.html.erb#L12)

@leenagupte - we have to hide it apparently rather than removing:

> When more than one checkbox is shown they are wrapped in a fieldset element, which requires a legend. This must be supplied to the component using the heading option.

ref: 
- https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/with_multiple_checkboxes
- https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/with_a_hidden_heading

app/views/qa/show.html.erb
```diff
@@ -33,14 +33,14 @@
         <% elsif multiple_question? %>
           <%= render "govuk_publishing_components/components/checkboxes", {
             name: "#{current_facet['facet']['key']}[]",
-            heading: "Select all that apply",
+            heading: current_facet["question"],
```
leenagupte: So this is the same as the title of the page? i.e. What's in the <H1>?

That looks fine to me, but I wonder if someone like @vanitabarrett or @andysellick could review.

andysellick: Have you considered only having the legend (perhaps setting it to 'Select all that apply'), and removing the hint text as in this example? https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/without_hint_text

DilwoarH: I don't think that works for us, it would change the design.

app/views/qa/show.html.erb
ChrisBAshton: I imagined we'd have to do something like this:

```html
<% if multiple_question? %>
  <!-- Let the checkboxes component render the title for us -->
  <% checkboxes = render "govuk_publishing_components/components/checkboxes", {
    name: "#{current_facet['facet']['key']}[]",
    heading: current_facet["question"],
    items: options
  } %>
<% else %>
  <!-- Render a separate title -->
  <%= render "govuk_publishing_components/components/title", {
    title: current_facet["question"],
  } %>
  <!-- Render the question here -->
<% endif %>
```

I appreciate this is a more substantial change than merely passing an argument to a component...!

app/views/qa/show.html.erb
ChrisBAshton: That would also introduce an a11y issue (as per Anika's comment on Slack). Documenting on GitHub:

> "Select all that apply" is not an adequate legend. The legend is there so that screen reader users can understand what a checkbox is about when they read it out of context. Screen reader users often just navigate from one form control to the next, so are likely to miss the question. It's the question that should be the legend.

app/views/qa/show.html.erb
leenagupte: Surely we always need to render the checkboxes?

ChrisBAshton: We always need to render the checkboxes (or radio buttons, or whatever the input may be) but we don't always need to render the title component, because sometimes the input component will render the title for us in an accessible way. Code sample is shortened for clarity ^

app/views/qa/show.html.erb
leenagupte: It looks like "Select all that apply" is the [default hint text](https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/with_legend_as_page_heading).

So perhaps, rather than `visually_hide_heading`, we can use `is_page_heading: true` instead?

DilwoarH:
I have made improvements to the component to make it fully accessible. Once the version is bumped up, I will update it here: https://github.com/alphagov/govuk_publishing_components/pull/971

app/views/qa/show.html.erb
```diff
@@ -31,38 +30,60 @@
...
-            heading: "Select all that apply",
-            visually_hide_heading: true,
-            items: options
-          } %>
+          <div class="govuk-!-margin-top-8">
+            <%= render "govuk_publishing_components/components/checkboxes", {
+              name: "#{current_facet['facet']['key']}[]",
+              heading: current_facet["question"],
+              description: render("govuk_publishing_components/components/govspeak", {
+                content: current_facet["description"].html_safe
```
leenagupte: What happens if the current_facet doesn't have a description? I'm asking because `multiple_grouped_question` has a guard around `current_facet["description"]`.

app/views/qa/show.html.erb
```diff
@@ -10,18 +10,17 @@
...
     <div class="govuk-grid-column-two-thirds">
-      <form action="<%= next_page_url %>" method="get" id="finder-qa-facet-filter-selection" class="gem-c-title" data-module="track-qa-choices">
-        <%= render "govuk_publishing_components/components/title", {
-          title: current_facet["question"],
-        } %>
-
-        <% if current_facet["description"] %>
-          <%= render "govuk_publishing_components/components/govspeak", {
-            content: current_facet["description"].html_safe
+      <form action="<%= next_page_url %>" method="get" id="finder-qa-facet-filter-selection" data-module="track-qa-choices">
```
ChrisBAshton: Can you explain in the commit message why the `gem-c-title` class was removed?

app/views/qa/show.html.erb
```diff
@@ -31,38 +30,60 @@
...
+              name: "#{current_facet['facet']['key']}[]",
+              heading: current_facet["question"],
+              description: render("govuk_publishing_components/components/govspeak", {
+                content: current_facet["description"].html_safe
+              }),
+              hint_text: current_facet["hint_title"],
+              is_page_heading: true,
+              items: options
+            } %>
+          </div>
```
ChrisBAshton: Could we refactor so that `<%= render "govuk_publishing_components/components/radio"` isn't repeated?

Something like:

```html
<%
  items = options
  if single_wrapped_question? {
    items = [
      { value: "yes"...},
      { value: "no"...}
    ]
  }
%>
<div class="govuk-!-margin-top-8">
  <%= render "govuk_publishing_components/components/radio", {
    name: "#{current_facet['facet']['key']}[]",
    heading: current_facet["question"],
    description: render("govuk_publishing_components/components/govspeak", {
      content: current_facet["description"].html_safe
    }),
    hint_text: current_facet["hint_title"],
    is_page_heading: true,
    items: items,
  } %>
</div>
```

leenagupte: I'd say this is nice to have. I'm not a fan of lots of logic in the views because they don't seem to be as well tested as a presenter.

ChrisBAshton: Agreed it's a conditional in a view, which is a bit "😬" but it's only moving a conditional around - rather than adding one - and simplifying the template in the process.

app/views/qa/show.html.erb
DilwoarH: I've added a guard on this too now

app/views/qa/show.html.erb
DilwoarH: I think it was put in by mistake. This is a class for the title gem, I don't think it was intended to be put into here.

app/views/qa/show.html.erb
DilwoarH: I say leave it as it is, the whole file needs to be refactored. We can improve this when there is a need.

app/views/qa/show.html.erb
ChrisBAshton: Probably worth noting that in the commit too 👍

DilwoarH:
Waiting on https://github.com/alphagov/govuk_publishing_components/pull/985

DilwoarH:
@leenagupte - tests now passing, waiting for e2e to finish

app/views/qa/show.html.erb
DilwoarH: Updated 

```
The `class="gem-c-title"` from form tag as this was done in error.
```

PR #1224, opened 2019-06-26
Use topic param for topic page filtering

bilbof:
When a user comes from a topic page, we'll set topic=id
rather than parent=topic_id. This will work on all pages
that have a topic param in the content item.

Example for filtering by brexit

`/search/services?topic=d6c2de5d-ef90-45d1-82d4-5f2438369eea`

---

- http://finder-frontend-pr-1224.herokuapp.com/search/all
- http://finder-frontend-pr-1224.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1224.herokuapp.com/search/research-and-statistics?topic=d6c2de5d-ef90-45d1-82d4-5f2438369eea
- http://finder-frontend-pr-1224.herokuapp.com/search/advanced?group=guidance_and_regulation&topic=%2Feducation
- http://finder-frontend-pr-1224.herokuapp.com/drug-device-alerts?topic=d6c2de5d-ef90-45d1-82d4-5f2438369eea
- http://finder-frontend-pr-1224.herokuapp.com/find-eu-exit-guidance-business?topic=d6c2de5d-ef90-45d1-82d4-5f2438369eea

### DIFF
--- a/app/lib/registries/full_topic_taxonomy_registry.rb
+++ b/app/lib/registries/full_topic_taxonomy_registry.rb
@@ -21,13 +21,13 @@ module Registries
     end
 
     def report_error
-      GovukStatsd.increment("#{NAMESPACE}.full_topic_taxonomy_api_errors")
+      GovukStatsd.increment("registries.full_topic_taxonomy_api_errors")
     end
 
     def format_taxon(taxon)
       {
         'title' => taxon['title'],
-        'content_id' => taxon['content_id']
+        'base_path' => taxon['base_path']
       }
     end
 
@@ -37,7 +37,7 @@ module Registries
 
     def flatten_taxonomy(taxons, output_hash)
       taxons.each do |taxon|
-        output_hash[taxon['base_path']] = format_taxon(taxon)
+        output_hash[taxon['content_id']] = format_taxon(taxon)
         unless taxon.dig('links', 'child_taxons').nil?
           flatten_taxonomy(taxon['links']['child_taxons'], output_hash)
         end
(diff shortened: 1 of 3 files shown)


app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -21,13 +21,13 @@ def cacheable_data
     end
 
     def report_error
-      GovukStatsd.increment("#{NAMESPACE}.full_topic_taxonomy_api_errors")
+      GovukStatsd.increment("registries.full_topic_taxonomy_api_errors")
```
barrucadu: Should this have changed?

bilbof: Yeah. It's an undefined constant if you run the spec for this file on its own. I meant to remove this for all registries, but missed this one.

app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -37,7 +37,7 @@ def level_one_taxons
 
     def flatten_taxonomy(taxons, output_hash)
       taxons.each do |taxon|
-        output_hash[taxon['base_path']] = format_taxon(taxon)
+        output_hash[taxon['content_id']] = format_taxon(taxon)
```
barrucadu: This makes data in the `FullTopicTaxonomyRegistry` a superset of the data in the `TopicTaxonomyRegistry`, which is also keyed by content ID but only has first level taxons.  We should consider removing `TopicTaxonomyRegistry` now.

bilbof: Yep that's a good idea.

PR #1227, opened 2019-06-26
Various refactorings

koetsier:
Some refactoring to make adding redirects from legacy advanced finders
easier.

- adds FactoryBot and some taxonomy related factories
- refactors taxonomy_spec_helper so that only two methods are used to set up taxonomy tests. 
- removes memoization of Service object which can cause problems with tests


app/lib/registries/full_topic_taxonomy_registry.rb
```diff
@@ -26,29 +26,34 @@ def report_error
...
+    def flatten_taxonomy(taxons)
+      return {} if taxons.empty?
+
+      taxons.inject({}) do |result, taxon|
+        child_taxons = taxon.dig('links', 'child_taxons') || []
+
+        taxon_hash = format_taxon(taxon)
+        child_taxon_hashes = flatten_taxonomy(child_taxons)
+
+        result.merge(taxon_hash).merge(child_taxon_hashes)
```
bilbof: Nice change.

spec/lib/advanced_search_finder_api_spec.rb
```diff
@@ -1,121 +1,4 @@
 require "spec_helper"
 
 describe AdvancedSearchFinderApi do
-  let(:finder_item) {
```
bilbof: Jumping the gun a bit here :D

bilbof:
Lets get this merged, as I think it'll help get https://github.com/alphagov/finder-frontend/pull/1204 done

PR #1228, opened 2019-06-27
Change wording for EU citizens

DilwoarH:
## What

Change the wording of the employing EU citizens in the business finder smart answer.

## Why

Research found that people were confused by the terminology of these questions. 

## Background

https://trello.com/c/uNWqplBr/165-change-wording-of-question-about-employing-eu-citizens

### Before
Screen Shot 2019-06-27 at 10 50 05

### After
Screen Shot 2019-06-27 at 10 49 00

## Demo
http://finder-frontend-pr-1228.herokuapp.com/prepare-business-uk-leaving-eu?business_activity-yesno=yes&sector_business_area%5B%5D=electronics-parts-machinery&page=3

https://trello.com/c/WDUH4kll/186-change-wording-for-eu-citizens

### DIFF
--- a/app/controllers/qa_controller.rb
+++ b/app/controllers/qa_controller.rb
@@ -64,13 +64,16 @@ private
   def current_facet
     current_facet = facets[page - 1]
     current_facet_config = qa_config["pages"][current_facet["key"]]
-    {
+    config = {
       "question" => current_facet_config["question"],
       "description" => current_facet_config["description"],
       "hint_title" => current_facet_config["hint_title"],
       "hint_text" => current_facet_config["hint_text"],
       "facet" => current_facet
     }
+    config.merge!("custom_options" => current_facet_config["custom_options"]) if current_facet_config["custom_options"].present?
+
+    config
   end
   helper_method :current_facet
 
@@ -93,7 +96,16 @@ private
     @facet_grouped_allowed_values ||= current_facet["facet"]["allowed_values"].group_by { |filter| filter["value"] }
   end
 
+  def custom_options?
+    question_type == "single" && current_facet['custom_options'].present?
+  end
+
   def options
+    if custom_options?
+      return current_facet["custom_options"].map do |option|
+        { label: option["label"], text: option["label"], value: option["value"] }
+      end
+    end
     allowed_values = current_facet["facet"]["allowed_values"]
     allowed_values.map do |option|
       { label: option["label"], text: option["label"], value: option["value"] }
--- a/lib/prepare_business_uk_leaving_eu.yaml
+++ b/lib/prepare_business_uk_leaving_eu.yaml
@@ -114,7 +114,12 @@ pages:
     question_type: "single"
     question: "Do you employ anyone from another European country?"
     description: |
-      <p>This includes <a href="/eu-eea">countries that are in the EU</a> and Norway, Switzerland, Iceland and Liechtenstein.</p>
+      <p>This includes <a href="/eu-eea">countries in the EU</a> and Norway, Switzerland, Iceland and Liechtenstein.</p>
+    custom_options:
+      - label: "Yes"
+        value: "yes"
+      - label: "No"
+        value: "no"
   personal_data:
     show_in_qa: true
     question_type: "single_wrapped"
--- a/features/fixtures/aaib_reports_example.json
+++ b/features/fixtures/aaib_reports_example.json
@@ -222,6 +222,24 @@
         ]
       },
       {
+        "key":"report_with_custom_labels",
+        "name":"Reports with custom labels",
+        "type":"text",
+        "preposition":"reports",
+        "display_as_result_metadata":true,
+        "filterable":true,
+        "allowed_values":[
+          {
+            "label":"Report A",
+            "value":"report_a"
+          },
+          {
+            "label":"Report B",
+            "value":"report_b"
+          }
+        ]
+      },
+      {
         "key":"date_of_occurrence",
         "name":"Date of occurrence",
         "short_name":"Occurred",
--- a/features/fixtures/aaib_reports_qa.yaml
+++ b/features/fixtures/aaib_reports_qa.yaml
@@ -35,6 +35,15 @@ pages:
     show_in_qa: true
     question_type: "multiple"
     question: "What is the multiple report type?"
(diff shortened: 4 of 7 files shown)


app/views/qa/show.html.erb
```diff
@@ -49,6 +49,11 @@
             }
           ]
         } %>
+      <% elsif single_custom_options_question? %>
```
leenagupte: I don't think you need this `elsif` as the code in it is identical to the `else`.

app/controllers/qa_controller.rb
```diff
@@ -89,6 +98,11 @@ def facet_grouped_allowed_values
   end
 
   def options
+    if single_custom_options_question?
```
leenagupte: Can you add a test for this new option? Perhaps something similar to: https://github.com/alphagov/finder-frontend/blob/master/features/qa.feature?

Also, what happens if the question is a `single_custom_options` question, but there aren't any custom options set?

app/controllers/qa_controller.rb
```diff
@@ -61,9 +61,13 @@ def breadcrumbs
   end
   helper_method :breadcrumbs
 
+  def current_facet_config
```
leenagupte: Do you need to separate this out?

Can you check if `custom_options` exist for current_facet instead?

app/views/qa/show.html.erb
DilwoarH: My bad, I was testing other ways and forgot to take it out

lib/prepare_business_uk_leaving_eu.yaml
```diff
@@ -117,10 +117,15 @@ pages:
...
     show_in_qa: true
-    question_type: "single"
+    question_type: "single_custom_options"
     question: "Do you employ anyone from another European country?"
     description: |
-      <p>This includes <a href="/eu-eea">countries that are in the EU</a> and Norway, Switzerland, Iceland and Liechtenstein.</p>
+      <p>This includes <a href="/eu-eea">countries in the EU</a> and Norway, Switzerland, Iceland and Liechtenstein.</p>
+    custom_options:
+      - label: "Yes"
+        value: "yes"
```
leenagupte: Can you a note in the commit message saying that this only works if the values match the values in the facet_group:  https://github.com/alphagov/content-tagger/blob/master/lib/data/find-eu-exit-guidance-business.yml#L243

app/controllers/qa_controller.rb
DilwoarH: No, it doesn't unfortunately, current_facet is built using the following function

```python
def current_facet
    current_facet = facets[page - 1]
    {
      "question" => current_facet_config["question"],
      "description" => current_facet_config["description"],
      "hint_title" => current_facet_config["hint_title"],
      "hint_text" => current_facet_config["hint_text"],
      "facet" => current_facet
    }
  end
```

leenagupte: Can't "custom_options" be added to current_facet?

e.g.

```
  def current_facet
    current_facet = facets[page - 1]
    current_facet_config = qa_config["pages"][current_facet["key"]]
    {
      "question" => current_facet_config["question"],
      "description" => current_facet_config["description"],
      "hint_title" => current_facet_config["hint_title"],
      "hint_text" => current_facet_config["hint_text"],
      "custom_options" => current_facet_config["custom_options"],
      "facet" => current_facet
    }
  end
```

DilwoarH: I don't think it's a good idea to add this into the current_facet function as `custom_options` is only available when using `single` question type. The current facet is generic across all types.

leenagupte: Or:

```
  def current_facet
    current_facet = facets[page - 1]
    current_facet_config = qa_config["pages"][current_facet["key"]]
     config = {
      "question" => current_facet_config["question"],
      "description" => current_facet_config["description"],
      "hint_title" => current_facet_config["hint_title"],
      "hint_text" => current_facet_config["hint_text"],
      "facet" => current_facet
    }
    config.merge("custom_options" => current_facet_config["custom_options"]) if current_facet_config["custom_options"].present?
  end
  
   config
```

lib/prepare_business_uk_leaving_eu.yaml
leenagupte: I see you've updated the commit message which is great, but I think it needs to be clear that it's the facet_values config (published by content tagger) that needs to match, not the finder content_item (published by search-api).

> It is important to note that each value must match a facet value
in the finder which it will redirect to, otherwise, it would not be
applied and ignored by the finder. These lists are not validated against
the finder, therefore, if the finder facet values change, this will need
to be changed too.

DilwoarH:
@leenagupte all issues have been addressed. Please let me know if you want further changes

app/controllers/qa_controller.rb
```diff
@@ -93,7 +96,16 @@ def facet_grouped_allowed_values
     @facet_grouped_allowed_values ||= current_facet["facet"]["allowed_values"].group_by { |filter| filter["value"] }
   end
 
+  def custom_options?
```
leenagupte: I wonder if this check is working? The page is no longer pulling out the custom label.
See: http://finder-frontend-pr-1228.herokuapp.com/prepare-business-uk-leaving-eu?business_activity-yesno=yes&sector_business_area%5B%5D=electronics-parts-machinery&page=3

app/controllers/qa_controller.rb
```diff
@@ -64,13 +64,16 @@ def breadcrumbs
...
     current_facet_config = qa_config["pages"][current_facet["key"]]
-    {
+    config = {
       "question" => current_facet_config["question"],
       "description" => current_facet_config["description"],
       "hint_title" => current_facet_config["hint_title"],
       "hint_text" => current_facet_config["hint_text"],
       "facet" => current_facet
     }
+    config.merge("custom_options" => current_facet_config["custom_options"]) if current_facet_config["custom_options"].present?
```
leenagupte: Sorry, my mistake, the `config.merge` needs an `!` so that the original `config` variable is updated.

The line should be: 
```
config.merge!("custom_options" => current_facet_config["custom_options"]) if current_facet_config["custom_options"].present?
``

PR #1229, opened 2019-06-27
Revert "Add blue header bar for users from topics"

bilbof:
Reverts alphagov/finder-frontend#1204

There's a wee issue with email-signup page css, so we'll roll back while we fix that.


bilbof:
Closing as we have a fix https://github.com/alphagov/finder-frontend/pull/1230

PR #1258, opened 2019-07-15
Refactoring

koetsier:
More refactoring, mainly:

1. Remove development code from content item model - separate class method for loading
   from the content store.
2. Using the ContentItem model instead of the raw hash in most parts of the application

1. Makes the ContentItem simpler and makes unit testing the work in 2 easier.
2. Makes the application more consistent - eventually hope to use mostly models instead of raw hashes.

---

- http://finder-frontend-pr-1258.herokuapp.com/search/all
- http://finder-frontend-pr-1258.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1258.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-1258.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1258.herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-1258.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- http://finder-frontend-pr-1258.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- http://finder-frontend-pr-1258.herokuapp.com/uk-nationals-living-eu
- http://finder-frontend-pr-1258.herokuapp.com/prepare-business-uk-leaving-eu

### DIFF
--- a/app/lib/search/query_builder.rb
+++ b/app/lib/search/query_builder.rb
@@ -60,7 +60,7 @@ module Search
     end
 
     def documents_per_page
-      finder_content_item['details']['default_documents_per_page'] || 1500
+      finder_content_item.default_documents_per_page
     end
 
     def return_fields_query
@@ -95,7 +95,7 @@ module Search
     end
 
     def raw_facets
-      @raw_facets ||= FacetExtractor.new(finder_content_item).extract
+      @raw_facets ||= FacetExtractor.new(finder_content_item.as_hash).extract
     end
 
     def unfilterise(field = '')
@@ -124,7 +124,7 @@ module Search
     end
 
     def remove_stopwords?
-      params["keywords"].present? && ["/find-eu-exit-guidance-business"].include?(finder_content_item["base_path"])
+      params["keywords"].present? && ["/find-eu-exit-guidance-business"].include?(finder_content_item.base_path)
     end
 
     def remove_stopwords
@@ -192,11 +192,11 @@ module Search
     end
 
     def base_filter
-      finder_content_item['details']['filter'].to_h
+      finder_content_item.filter
     end
 
     def base_reject
-      finder_content_item['details']['reject'].to_h
+      finder_content_item.reject
     end
 
     def facet_query
--- a/app/lib/search/query_builder.rb
+++ b/app/lib/search/query_builder.rb
@@ -95,7 +95,7 @@ module Search
     end
 
     def raw_facets
-      @raw_facets ||= FacetExtractor.new(finder_content_item.as_hash).extract
+      finder_content_item.raw_facets
     end
 
     def unfilterise(field = '')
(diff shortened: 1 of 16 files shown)


app/lib/search/query_builder.rb
```diff
@@ -106,7 +103,7 @@ def unfilterise(field = '')
     end
 
     def order_query
-      order_query_builder_class.new(
+      OrderQueryBuilder.new(
```
bilbof: Good stuff. Hangover from the Advanced Search stuff?

app/lib/content_item_builder.rb
```diff
@@ -0,0 +1,54 @@
+class ContentItemBuilder
```
bilbof: Could you add a note about the motivation for the `ContentItemBuilder` class? I'm not sure that making testing easier should be the reason we make this change.

app/models/content_item.rb
```diff
@@ -59,6 +59,14 @@ def base_path
     content_item_hash.dig('base_path')
   end
 
+  def raw_facets
+    @raw_facets ||= FacetExtractor.new(content_item_hash).extract
```
bilbof: What are raw facets here?

app/lib/content_item_builder.rb
koetsier: I think mainly to separate 'holding data' from 'fetching data' and 'caching data'. i think the improved testability is a result of the separation.

bilbof: I think I need to do some reading on the builder class pattern! Aside from testing benefits, I don't see the value in having a builder class here. At the moment there's only one way to get a content item (from the cache/content store call), so there's not much config work to be done.

I think that separating the caching and data model is not really necessary. If the caching logic is complex, we could create a module for cacheable models like we did for registries and include it.

app/models/content_item.rb
koetsier: Oh that is a refactor too many, I'll remove

app/lib/search/query_builder.rb
koetsier: think so. only place it is used so this looks clearer

app/lib/content_item_builder.rb
koetsier: So the alternative is we leave it as it is, but then whenever you want to test code involving the content_item model, you have to start stubbing out web request, which adds quite a bit of additional complexity throughout the testing code - and you have to 'know' that that that is how the content_item class gets its data. 

I agree it isn't a massive improvement of the model itself (although I do think it is ever so slightly better) but I do think the testability issue is important.

bilbof: Maybe you could do `ContentItem.for('/services')`.
```ruby
def for(path)
   # get hold of an item, do some fancy stuff.
   new(hash)
end

private
   
   :new
```

koetsier: Ok, did that - also removed all development finder stuff. 👍

bilbof: 🍰

PR #1270, opened 2019-07-22
Adds official-documents fixture

hannako:
This fixture was deleted in [a recent code clear up](https://github.com/alphagov/finder-frontend/pull/1199), but it's useful for local development so I've added it back.


hannako:
@andysellick i made the changes andy if you could 👍 pls :)

PR #1294, opened 2019-08-07
Add cancelled statistics to the research and stats finder

hannako:
Before:

If a stats announcement is cancelled, it is lost from the finder once its scheduled release date is passed. As it won't show up in the published statistics results, and its outside of the time period of the scoped upcoming statistics search (results are from today onwards)

After:

Screen Shot 2019-08-07 at 18 42 04

https://trello.com/c/TRKpC1SC/935-users-cant-find-cancelled-statistics-releases

---

## Search page example to sanity check:

- http://finder-frontend-pr-1294.herokuapp.com/search/research-and-statistics

### DIFF
--- a/app/presenters/statistics_sort_presenter.rb
+++ b/app/presenters/statistics_sort_presenter.rb
@@ -1,25 +1,25 @@
 class StatisticsSortPresenter < SortPresenter
   def initialize(content_item, filter_params)
-    @stats_grouping = filter_params['content_store_document_type']
+    @doc_type = filter_params['content_store_document_type']
     super(content_item, filter_params)
   end
 
 private
 
-  attr_reader :stats_grouping, :user_selected_order, :keywords, :content_item_sort_options
+  attr_reader :doc_type, :user_selected_order, :keywords, :content_item_sort_options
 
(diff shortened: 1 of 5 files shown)


app/presenters/statistics_sort_presenter.rb
```diff
@@ -1,25 +1,25 @@
...
 private
 
-  attr_reader :stats_grouping, :user_selected_order, :keywords, :content_item_sort_options
+  attr_reader :doc_type, :user_selected_order, :keywords, :content_item_sort_options
 
   RELEVANCE_OPTION_TYPES = %w(relevance -relevance).freeze
   EXCLUDED_OPTIONS = {
-    published: %w(-release_timestamp release_timestamp),
-    upcoming: %w(-public_timestamp public_timestamp),
+    public: %w(-release_timestamp release_timestamp),
```
sihugh: I find this bit slightly confusing.  I wonder if there's a better way that doesn't rely on us knowing what's in the content item.  That doesn't stop this PR though.

hannako: Could we have two different sort types for this finder's content item? `sort_published` which would be used for research and published stats, and `sort_unpublished` which would be used by cancelled and upcoming?

PR #1338, opened 2019-08-23
Prototype for the a mobile design with facets split into it's own page

injms:
This is the prototype for the new designs for smaller screens for the all finder. Emphasis on prototype, as it basically shows and hides things on the same page depending on which button the form is submitted with - this was the only way I could think of retaining the state of the search without JavaScript.

 - adds a filter page where facets can be selected
 - once the filter has been applied, a user can then go back to the search results

To get this working for the protoype, this is just showing and hiding certain elements of the same page to prevent a large-scale everything rewrite.

Trello card: https://trello.com/c/i8xtPcbY/947-build-the-prototype-for-improved-mobile-search-frontend

_I cannot emphasise enough how none of this should go into production! This is just to give an idea of how splitting the facets could work with real search results can be used when testing it_

WHAT WAS DONE

Phase one:
- Search box restyled on mobile, placeholder text added and a search button added
- Filter page added where facets can be selected on mobile - header and footer removed
- Filters hidden on search page with a "Filters" link added that opens the filter page and shows the number of currently-applied filters
- Facet tags not shown on search page, only on the new filters page
- Facet tag styles tweaked (no background, darker border) and displayed inline instead of block
- Keyword search facets not shown
- Filters restyled to look like accordions
- Topic selects and date range inputs moved into an "Expander" component
- Expander component restyled like an accordion
- A floating search button added to filters page with number of results available that dynamically updates with filter selection
- Clear filter link added to the filters page that goes back to the search page (thus removing all filters)
- Search results have an increased space and separator line removed
- Subscribe to feed button moved to the bottom of results


benthorner:
@injms just looking through the PRs and tidying up. Do you want to keep this open?

kr8n3r:
@benthorner need to keep this open for a while. will close once we start the actual work

kr8n3r:
closing but keeping the branch

PR #1487, opened 2019-09-05
Factories and validations

benthorner:
https://trello.com/c/o9XgkCxT/126-send-email-updates-for-actions

This PR has a fair few changes to make our psuedo-model classes more conventional. Please see the commits for more explanation. It aims to resolve a few pain points:

   - Test data can diverge from what's actually valid
   - Our initializers are repetitive/verbose
   - Our test data is verbose (due to lack of factories)

It also provides a foundation to improve validations for test data specifically - in particular, we'd like to validate that we're writing valid criteria expressions in our specs.

---

- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/search/all
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/uk-nationals-living-eu
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/prepare-business-uk-leaving-eu


lib/checklists/action.rb
```diff
@@ -1,44 +1,38 @@
...
-    @criteria = params['criteria']
-    @audience = params['audience']
-    @guidance_link_text = params['guidance_link_text']
-    @guidance_url = params['guidance_url']
-    @guidance_prompt = params['guidance_prompt']
-    @priority = params['priority']
+  validates_presence_of :id, :title, :consequence, :criteria
+  validates_inclusion_of :audience, in: %w(business citizen)
+  validates_presence_of :guidance_link_text, if: :guidance_url
+  validates_format_of :priority, with: /\d+/
```
theseanything: Would be nice to do this:
```suggestion
  validates_numericality_of :priority, : only_integer
```
But I assume priority is being parsed as a string from the YAML?

lib/checklists/action.rb
```diff
@@ -1,44 +1,38 @@
...
+    validate!
   end
 
   def show?(selected_criteria)
     Checklists::CriteriaLogic::Evaluator.evaluate(criteria, selected_criteria)
   end
 
+  def self.load(params)
+    parsed_params = params.dup
+    parsed_params['id'] = params['action_id']
```
theseanything: Possibly in the future - could we just rename the key in the YAML file and drop these two lines?

lib/checklists/action.rb
benthorner: Oh yeah - done 👍

lib/checklists/action.rb
benthorner: Definitely. I think it would make more sense if our YAML files weren't coupled to the spreadsheet import and were instead written to make them easy to load.

lib/checklists/change_note.rb
```diff
@@ -1,23 +1,40 @@
...
+
   CONFIG_PATH = Rails.root.join('lib/checklists/change_notes.yaml')
 
+  validates_presence_of :action_id
+  validates_inclusion_of :type, in: %w(addition content_change)
+  validates_format_of :time, with: /\d{4}-\d{2}-\d{2} \d{2}:\d{2}/
+
+  validate do
+    validates_presence_of :note if type == "content_change"
+  end
```
theseanything: I think this block can be
```suggestion
validates_presence_of :note, if: type == "content_change"
```

benthorner: I don’t think that would work, because it needs to lazily execute on the instance, not the class (i.e. `type` will be nil). I think you can do some lambda thing for some of these validations, but I was thinking the `validate` form was simpler.

benthorner: Sort of - updated as it does seem more readable.

lib/checklists/change_note.rb
```diff
@@ -1,23 +1,37 @@
...
   CONFIG_PATH = Rails.root.join('lib/checklists/change_notes.yaml')
 
+  validates_presence_of :action_id
+  validates_inclusion_of :type, in: %w(addition content_change)
+  validates_format_of :time, with: /\d{4}-\d{2}-\d{2} \d{2}:\d{2}/
+  validates_presence_of :note, if: -> { type == "content_change" }
+
+  validate do
+    errors.add("ID not a UUID") unless id.length == SecureRandom.uuid.length
+  end
```
theseanything:
```suggestion
validates_length_of :id, is: SecureRandom.uuid.length, message: "ID not a UUID"
```

theseanything: Or should this be a:
`validates_format_of?`

theseanything: Or:
 `validate { errors.add("ID not a UUID") unless UUID.validate(id) }`

lib/checklists/question.rb
```diff
@@ -34,10 +34,16 @@ def possible_values
     (options + sub_options).map(&:value).compact
   end
 
+  def self.load(params)
+    parsed_params = params.dup
+    parsed_params['text'] = params['question']
+    parsed_params['options'] = Option.load_all(params['options'])
+    parsed_params['type'] = params['question_type']
```
theseanything: Same here in the future we could rename `question -> text` and `question_type -> type`

benthorner: Yup :-).

lib/checklists/change_note.rb
benthorner: Nice - updated 👍 .

benthorner:
@theseanything thanks for the suggestions. I think this should be ready for another review now.

PR #1498, opened 2019-09-06
Add new module to track brexit questions

DilwoarH:
## What
Look into the current analytics that is being tracked and suggest improvements that can be made

## Why
Good analytics allow us to have a better understanding of how the product is being used.

## Changes
`Question Key` should be `event action` and `values label` should be `event label`.

Two new event categories - `brexit-checker-qa` and `brexit-checker-results`

https://trello.com/c/QTt0pDDW/211-change-question-event-tracking

### DIFF
--- a/app/assets/javascripts/modules/track-brexit-qa-choices.js
+++ b/app/assets/javascripts/modules/track-brexit-qa-choices.js
@@ -1,0 +1,43 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
+
+  GOVUK.Modules.TrackBrexitQaChoices = function () {
+    this.start = function (element) {
+      track(element)
+    }
+
+    function track (element) {
+      element.on('submit', function (event) {
+        var $checkedOption, eventLabel, options
+        var $submittedForm = $(event.target)
+        var $checkedOptions = $submittedForm.find('input:checked')
+        var questionKey = $submittedForm.data('question-key')
+
+        if ($checkedOptions.length) {
+          $checkedOptions.each(function (index) {
+            $checkedOption = $(this)
+            var checkedOptionId = $checkedOption.attr('id')
+            var checkedOptionLabel = $submittedForm.find('label[for="' + checkedOptionId + '"]').text().trim()
+            eventLabel = checkedOptionLabel.length
+              ? checkedOptionLabel
+              : $checkedOption.val()
+
+            options = { transport: 'beacon', label: eventLabel }
+
+            GOVUK.SearchAnalytics.trackEvent('brexit-checker-qa', questionKey, options)
+          })
+        } else {
+          // Skipped questions
+          options = { transport: 'beacon', label: 'no choice' }
+
+          GOVUK.SearchAnalytics.trackEvent('brexit-checker-qa', questionKey, options)
+        }
+      })
+    }
+  }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/modules/track-brexit-qa-choices.js
```diff
@@ -0,0 +1,40 @@
...
+
+        if ($checkedOptions.length) {
+          $checkedOptions.each(function (index) {
+            $checkedOption = $(this)
+            eventLabel = $checkedOption[0].labels.length ?
+              $($checkedOption[0].labels[0]).text().trim()
+              : $checkedOption.attr('name').replace('[]', '')
+            options = { transport: 'beacon', label: eventLabel }
+
+            GOVUK.SearchAnalytics.trackEvent('brexit-checker-qa', questionKey, options)
```
benthorner: This is mostly a lack of understanding on my part: it looks like we're sending three strings to GA - 'brexit-checker-qa', '<questionKey>' and '<eventLabel>' - but I think GA only has two fields to store them (category and label). Do you know what happens to the other string?

DilwoarH: It gets passed in as action:

Screen Shot 2019-09-06 at 14 18 11

Also in JS, you can pass any number of arguments to a function and it can be accessed via the `arguements` variable.

Screen Shot 2019-09-06 at 14 21 32
Screen Shot 2019-09-06 at 14 21 24

Try it out:
`var x = function(){console.log(arguments);}`

Screen Shot 2019-09-06 at 14 22 34

benthorner: Ah yes - forgot all about action!

spec/javascripts/modules/track-brexit-qa-choices.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  var $element
+
+  beforeEach(function () {
+    spyOn(GOVUK.SearchAnalytics, 'trackEvent')
+
+    $element = $(
+      '<div>' +
+        '<form onsubmit="event.preventDefault()" data-question-key="question-key">' +
+          '<div>' +
+            '<input name="sector_business_area[]" id="construction" type="checkbox" value="construction">' +
```
benthorner: Minor, but this is out of sync with our input naming.

spec/javascripts/modules/track-brexit-qa-choices.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+    expect(GOVUK.SearchAnalytics.trackEvent).toHaveBeenCalledWith(
+      'brexit-checker-qa', 'question-key', { transport: 'beacon', label: 'Accommodation label' }
+    )
+    expect(GOVUK.SearchAnalytics.trackEvent).toHaveBeenCalledWith(
+      'brexit-checker-qa', 'question-key', { transport: 'beacon', label: 'Construction label' }
+    )
+  })
+
+  it('track events sends value of checkbox when no label is set', function () {
```
benthorner: Could you help me understand when this scenario happens?

DilwoarH: This shouldn't ever happen if we use components but was done to handle any JS exceptions if labels are not set for checkboxes

spec/javascripts/modules/track-brexit-qa-choices.spec.js
DilwoarH: ah yes - shouldn't be an issue, just test data

PR #1502, opened 2019-09-06
Update facet remove button focus state

kr8n3r:
Follow-on from https://github.com/alphagov/finder-frontend/pull/1478

We are not likely to redesign the facet tags anytime soon, so best to update their focus state to match newer style.

This borrows the look from input focus styles

Example search
- http://finder-frontend-pr-1502.herokuapp.com/search/?keywords=pet+micro+pig+requirement&order=relevance

Before:
Screen Shot 2019-09-06 at 14 44 12

After: 
Screen Shot 2019-09-06 at 14 43 51

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -197,6 +197,12 @@
 
   .js-enabled & {
     display: inline-block;
+    border: 1px solid transparent;
+    &:focus {
+      -webkit-box-shadow: inset 0 0 0 2px;
+      box-shadow: inset 0 0 0 2px;
+      border: solid 1px govuk-colour("black");
+    }
   }
 }
 


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -197,6 +197,11 @@
 
   .js-enabled & {
     display: inline-block;
+    &:focus {
+      -webkit-box-shadow: inset 0 0 0 2px;
+      box-shadow: inset 0 0 0 2px;
+      border: 1px solid #000;
+    }
```
andysellick: Tiny nitpick: when the element is focused the 'x' moves very slightly, presumably related to the border addition? If you set a transparent 1px border on the default state it seems to fix it.

Normal:

Screen Shot 2019-09-06 at 16 01 59

Focus:

Screen Shot 2019-09-06 at 16 02 08

Actually it's easier to see this not as screenshots, oh well.

kr8n3r: yeah, I spotted it moved, but wasn't too bothered. 
have added the transparent border now

PR #1506, opened 2019-09-06
Add tracking for action and guidance links

DilwoarH:
## What
Look into the current analytics that is being tracked and suggest improvements that can be made

## Why
Good analytics allow us to have a better understanding of how the product is being used.

### Why do we track position of links?

The current reasoning for tracking the position of links is to understand where users are clicking on links. This information will be used to make informed decisions about the ordering of links and understand how far down the list users scroll and click. This is also something that is done in other finders across GOV.UK.

This is something that will need to be reviewed in the future and understand the user need more clearly.

### Action link
- Event Category: `brexit-checker-results`
- Event Action: `You or your family 1.02 - Action`
- Event Label: `[URL]`

### Guidance link
- Event Category: `brexit-checker-results`
- Event Action: `You or your family 1.02 - Guidance`
- Event Label: `[URL]`

Change `1.1` to `1.01` on action link tracking so that ordering in GA is done correctly

https://trello.com/c/Jg2VKa98/210-update-event-tracking-for-action-and-guidance-links

### DIFF
--- a/app/views/checklist/_action_list.html.erb
+++ b/app/views/checklist/_action_list.html.erb
@@ -1,4 +1,5 @@
 <% items.each.with_index do | action, index | %>
+  <% action_index = "#{audience_index}.#{'%02d' % (index + 1)}"%>
   <div class="govuk-grid-row govuk-!-margin-bottom-3">
     <div class="govuk-grid-column-three-quarters">
       <p class="govuk-body govuk-!-font-weight-bold">
@@ -7,8 +8,8 @@
             class="govuk-link"
             href="<%= action.title_url %>"
             data-module="track-click"
-            data-track-action="<%= audience_heading %> <%= audience_index %>.<%= index + 1 %>"
-            data-track-category="ContentItemClicked"
+            data-track-action="<%= audience_heading %> <%= action_index %> - Action"
+            data-track-category="brexit-checker-results"
             data-track-label="<%= action.title_url %>"
           ><%= action.title %></a>
         <% else %>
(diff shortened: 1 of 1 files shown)


app/views/checklist/_action_list.html.erb
```diff
@@ -1,4 +1,5 @@
 <% items.each.with_index do | action, index | %>
+  <% action_index = "#{audience_index}.#{'%02d' % (index + 1)}"%>
```
kevindew: Any chance this could go over 99? I assume you'll hit the same ordering problem if you had 1.99 then 1.100 but I'm guessing this risk is rather low.

DilwoarH: > Code looks good to me, any concerns about this not being backwards compatible with the previous data?

Performance analytics are happy with the change and are aware that some data might be out of order.

Regarding the 1.100 - I would like to hope that we don't ever get to a situation where there would be a 100 action as a finder would be a better use case for that scenario. If you think we should still make the change, happy to do so.

kevindew: I don't have enough context to know how crazy it would be for us to hit 100, so happy with a max of 99 if you're confident that is sufficient 👍

DilwoarH: I feel fairly confident that we won’t hit that, currently we haven’t hit 20 yet. This is also per category so should be fine.

DilwoarH: Spoke to the Performance analysis team, they don't think it will hit 100.

DilwoarH:
> I'm confused about the meaning of the index in this context, and I think it's worth clarifying the reason for using a dynamic index before we merge this, given that all actions have unique IDs. The description for the PR only states the change, but not the reasoning behind it. Happy to chat about it in person, if that's easier.

Screen Shot 2019-09-03 at 11 34 30

The numbers represent the category and position of the action in the results page.

So for example:
`1.01` is the first category ("1")  "You and your family" and the first action ("01") "Check your passport..."

`2.02` would be the second category ("2") and the second action ("02")

The reason why we can't just assume "You and your family" will always be the first category is that if there are no actions for the "You and your family" category, this will not show and "Your business" will become the first section.

All of this allows Performance analysts to work out the information needed to track where things are being clicked.

Hope this helps @benthorner

DilwoarH:
@benthorner 

Updated the commit message

> Why do we track position of links?
The current reasoning for tracking the position of links is to understand where users are clicking on links. This information will be used to make informed decisions about the ordering of links and understand how far down the list users scroll and click. This is also something that is done in other finders across GOV.UK.

This is something that will need to be reviewed in the future and understand the user need more clearly.

Issue #1323, opened 2019-08-21
Toggle functionality should be using a button


injms:
'Show more / fewer search options' toggle is an anchor element, and should be a button element as it's not a link to anywhere.

Code in question is in [_facet_collection.html.erb](https://github.com/alphagov/finder-frontend/blob/2f06ad0791d1df38a565e09ab3d8f3436e11988c/app/views/finders/_facet_collection.html.erb#L30)

```erb
<a href="#" class="facet-toggle" data-controls="facet-wrapper" data-expanded="false" data-track-category="filterClicked" data-track-action="accordion" data-toggled-text="- <%= t("finders.facet_disclosure.fewer", default: "Show fewer search options") %>">
  + <%= t("finders.facet_disclosure.more", default: "Show more search options") %>
</a>
```

PR #1542, opened 2019-09-11
Update element on toggle search options functionality

kr8n3r:
Use `button` instead of `a` for search options toggle
- Use a semantically correct HTMl element
- Add custom app styles to make it look like a link (retain previous visual)

Fixes: #1323

Visual check: 
- http://finder-frontend-pr-1542.herokuapp.com/search/all

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -305,3 +305,25 @@
     color: govuk-colour("white");
   }
 }
+
+.app-c-button-as-link {
+  padding: 0;
+  border-width: 0;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/finder_frontend.scss
```diff
@@ -305,3 +305,25 @@
     color: govuk-colour("white");
   }
 }
+
+.app-button-as-link {
```
injms: Should this be `.app-c-button-as-link`?

kr8n3r: it is now 😜

PR #1558, opened 2019-09-13
Pre-warm content items in cache

bilbof:
This should prevent requests from users requiring a content store call, meaning we will return search results to some users faster. This is a similar pattern to registries.

With this change, finder content items are cached ahead of users requests, and are periodically refreshed.

There are rake tasks to refresh the cache during a deploy and on a schedule:

```
# replaces finders in the cache, run on a schedule
rake content_store:refresh_cache_hard
# replaces finders in the cache if they're not already present, run at deploy
rake content_store:refresh_cache_soft
```

The reason we're doing this now is because there are more finder-frontend machines, which makes it more likely for an end user to hit a cold cache.

I added a new healthcheck for the cached content items.

Trello: https://trello.com/c/1PG0oqZg/992

---

- http://finder-frontend-pr-1558.herokuapp.com/search/all
- http://finder-frontend-pr-1558.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1558.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-1558.herokuapp.com/get-ready-brexit-check/questions
- http://finder-frontend-pr-1558.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1558.herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-1558.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- http://finder-frontend-pr-1558.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- http://finder-frontend-pr-1558.herokuapp.com/uk-nationals-living-eu
- http://finder-frontend-pr-1558.herokuapp.com/prepare-business-uk-leaving-eu

### DIFF
--- a/lib/fetch_finders.rb
+++ b/lib/fetch_finders.rb
@@ -1,0 +1,11 @@
+class FetchFinders
+  def self.from_search_api
+    GovukStatsd.time("fetch_finders.from_search_api.request_time") do
+      Services.rummager.search(
+        filter_rendering_app: 'finder-frontend',
+        fields: %w(link),
+        count: 1500,
+      ).dig('results')
+    end
+  end
+end
--- a/lib/fetch_finders.rb
+++ b/lib/fetch_finders.rb
@@ -1,6 +1,6 @@
 class FetchFinders
   def self.from_search_api
-    GovukStatsd.time("fetch_finders.from_search_api.request_time") do
+    GovukStatsd.time("rummager.fetch_finders") do
       Services.rummager.search(
         filter_rendering_app: 'finder-frontend',
         fields: %w(link),
--- a/lib/services/content_store.rb
+++ b/lib/services/content_store.rb
@@ -29,7 +29,7 @@ private
       content_store.content_item(base_path).to_h
     end
   rescue GdsApi::HTTPServerError, GdsApi::HTTPBadGateway
-    GovukStatsd.increment("content_store_service.connection_error")
+    GovukStatsd.increment("content_store.connection_error")
     raise
   end
 
--- a/lib/services/content_store.rb
+++ b/lib/services/content_store.rb
@@ -1,0 +1,43 @@
+class Services::ContentStore
+  def hard_refresh_cache
+    finder_paths.map do |base_path|
+      item = content_item(base_path)
+      Rails.cache.write(cache_key(base_path), item)
+    end
+  end
+
+  def soft_refresh_cache
+    finder_paths.map do |base_path|
+      cached_content_item(base_path)
+    end
+  end
+
+  def cached_content_item(base_path)
+    Rails.cache.fetch(cache_key(base_path)) do
+      content_item(base_path)
+    end
+  end
+
+  def cache_key(base_path)
+    "finder-frontend_content_items#{base_path}"
+  end
+
+private
+
+  def content_item(base_path)
+    GovukStatsd.time("content_store.fetch_request_time") do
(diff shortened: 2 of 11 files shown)


bilbof:
Yeah it'd be available in this format:

```
stats.timers.govuk.app.finder-frontend.*.<directory>.<stat>.metric
stats.timers.govuk.app.finder-frontend.*.content_store.fetch_request_time.upper_90
```

Thanks for your comment, that's much better than what I had - I've made those changes.

lib/services/content_store.rb
```diff
@@ -0,0 +1,43 @@
...
+    "finder-frontend_content_items#{base_path}"
+  end
+
+private
+
+  def content_item(base_path)
+    GovukStatsd.time("content_store.fetch_request_time") do
+      content_store.content_item(base_path).to_h
+    end
+  rescue GdsApi::HTTPServerError, GdsApi::HTTPBadGateway
```
kevindew: You probably want this to be `Gds::HTTPIntermittentServerError`. In the case above `GdsApi::HTTPBadGateway` has no effect as it is a subclass of `GdsApi::HTTPSserverError`.

lib/services/content_store.rb
```diff
@@ -0,0 +1,43 @@
+class Services::ContentStore
+  def hard_refresh_cache
+    finder_paths.map do |base_path|
+      item = content_item(base_path)
+      Rails.cache.write(cache_key(base_path), item)
```
kevindew: These should probably expire at some point in time - even to just make the service self cleaning up. 

The requests from content store will return a time that they can be cached until which would be consistent to acknowledge. That'd also resolve a problem that could happen here if any finder had an update scheduled to be published this would need a manual flush.

lib/services/content_store.rb
```diff
@@ -0,0 +1,43 @@
...
+
+private
+
+  def content_item(base_path)
+    GovukStatsd.time("content_store.fetch_request_time") do
+      content_store.content_item(base_path).to_h
+    end
+  rescue GdsApi::HTTPServerError, GdsApi::HTTPBadGateway
+    GovukStatsd.increment("content_store.connection_error")
+    raise
```
kevindew: Do you need to consider that any of these could return a 404 as that'd break the whole cache right?

lib/fetch_finders.rb
```diff
@@ -0,0 +1,11 @@
+class FetchFinders
+  def self.from_search_api
+    GovukStatsd.time("rummager.fetch_finders") do
+      Services.rummager.search(
+        filter_rendering_app: 'finder-frontend',
+        fields: %w(link),
+        count: 1500,
```
kevindew: Does this limit results from search-api? do we need a limit?

bilbof:
@kevindew you are right, this would not be worth the added complexity given the low impact it would have for users. I will close this for now.

There are 38 content items in search api rendered by finder frontend, and 30 machines with a cache expiry on content items every hour. So that would prevent a maximum of 1140 requests an hour to content store.

We want a shared memcached (https://trello.com/c/WFvKwlLN), so that would solve this problem.

I wanted the cache to patch over the content store response times, but it might be worth me looking into speeding up the already fast content store response time.

PR #1562, opened 2019-09-16
Remove PageService class

theseanything:
This PR aims to remove the page service and enable question resolution to support a backwards flow through the question.

The functionality of page service is being replaced with a class method `resolve` on the Question class to simplify logic. Also there was source of confusion over the meaning of the page parameter.  Now variables within the code accurately represent it as the previous question id.


spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:q4) { FactoryBot.build(:brexit_checker_question, criteria: %w(c)) }
+
+  let(:questions) { [q1, q2, q3, q4] }
+
+  describe '#resolve' do
+    before do
+      allow(BrexitChecker::Question).to receive(:load_all) { questions }
+    end
+
+    let(:subject) {
```
benthorner: Would it be possible to push some of this into the individual 'it' or 'context' blocks? I still find it confusing to see variables like `criteria_keys` set multiple times in the nesting structure, and I'm trying to avoid it myself.

spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    context 'the page on the form is 0' do
+      let(:previous_question) { 0 }
+      it 'returns question one ' do
+        expect(subject).to eq(q1)
+      end
+    end
+
+    context 'the page on the form is 1 and the criteria_keys match question_two' do
+      let(:previous_question) { 1 }
+      let(:criteria_keys) { %w[c d] }
```
benthorner: I wonder if we should stub the `show?` method of each question, so that the test isn't so tied to the Criteria::Evaluator? At first I didn't notice it, but when I got down here I was wondering why these particular keys would give question 3, and had to scroll up to the top and parse the expression. Some of that seems necessary, so maybe it's just a matter of moving the `let`'s a bit closer to the test.

spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      let(:criteria_keys) { %w[c] }
+      it 'returns question four' do
+        expect(subject).to eq(q4)
+      end
+    end
+
+    context 'the page on the form is 1 and the criteria_keys do not match anything' do
+      let(:previous_question) { 1 }
+      let(:criteria_keys) { %w[e] }
+      it 'returns nil' do
```
benthorner: Could we be more explicit here, and in the next test, to indicate that we've reached the end of the questions? I think it's important to be clear about the meaning of `nil` if we're using it to make decisions later.

spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    context 'the page on the form is 1 and the criteria_keys match question_two' do
+      let(:previous_question) { 1 }
+      let(:criteria_keys) { %w[c d] }
+      it 'returns question three' do
+        expect(subject).to eq(q3)
+      end
+    end
+
+    context 'the page on the form is 1 and the criteria_keys match question_three' do
```
benthorner: Minor: could we rephrase this to describe why this is a special case? The setup is pretty self-evident below, so we have the space to write something more descriptive.

lib/brexit_checker/question.rb
```diff
@@ -44,6 +45,12 @@ def self.load(params)
 
   def self.load_all
     @load_all = nil if Rails.env.development?
-    @load_all ||= YAML.load_file(CONFIG_PATH)['questions'] .map { |q| load(q) }
+    @load_all ||= YAML.load_file(CONFIG_PATH)['questions'].map.with_index { |q, i| load(i, q) }
+  end
+
+  def self.resolve(criteria_keys: [], previous_question: 0)
```
benthorner: I'm wondering if this method really needs to be in the Question class. This is partly my allergy of putting things in models (because so much can go into them), but also that a helper method might lend itself to a more descriptive name, such as `next_question`, whereas it would be weird to say `Question.next_question`.

spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    let(:previous_question) { nil }
+
+    context 'the page on the form is 0' do
+      let(:previous_question) { 0 }
+      it 'returns question one ' do
+        expect(subject).to eq(q1)
+      end
+    end
+
+    context 'the page on the form is 1 and the criteria_keys match question_two' do
```
emmabeynon: Should this be 'match question three'?

spec/lib/brexit_checker/question_spec.rb
emmabeynon: Should this be 'match question four'?

lib/brexit_checker/question.rb
theseanything: Yup - moved to helper.

spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'spec_helper'
+
+describe BrexitChecker::Question do
+  let(:q1) { FactoryBot.build(:brexit_checker_question) }
```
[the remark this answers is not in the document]
theseanything: Scoped as tests now moved to helper sec.

spec/lib/brexit_checker/question_spec.rb
theseanything: removed nested lets.

spec/lib/brexit_checker/question_spec.rb
theseanything: I've stubbed `show?`. Also refactored the examples so hopefully they make a bit more sense now.

spec/lib/brexit_checker/question_spec.rb
theseanything: Updated the description of the example - is it a bit clearer?

spec/helpers/brexit_checker_helper_spec.rb
```diff
@@ -57,4 +57,71 @@
       expect(subject).to contain_exactly('A', 'B')
     end
   end
+
+  describe "#next_question" do
+    let(:q1) { FactoryBot.build(:brexit_checker_question) }
+    let(:q2) { FactoryBot.build(:brexit_checker_question, criteria: [{ "all_of" => %w(a b) }]) }
```
benthorner: Minor: do we need to specify the criteria here any more?

theseanything:
@benthorner 
> I'd like to avoid introducing dynamic IDs into the questions, since this information is already stored as part of the ordered list of questions. A smaller change could be to have the `resolve` method return the index of the next question, which can be set as a param and used to get the actual question.

Agree, not a huge fan of dynamically adding in the IDs. I tried having `next_question/resolve` output an array index - but it ends up being the same complex logic we had in the PageService. Alternatively, we can explicit add a key to the YAML with a sequential "ID" - which I also prefer as it prevents having to rely implicit assumption that array order is question order.

> As mentioned in a lower comment, I'd like to know your thoughts about putting this method into a helper, partly as I think it would help expand the naming options for it; and partly because I think the logic is closer to the view than it is to the model, since we'll likely need a `previous_question(_index)` method soon.

theseanything:
@emmabeynon 
> Just a couple of comments, this may be because I find the interaction between questions and criteria a bit difficult to understand!

Ah sorry - it was my mistake - I copied the tests from the page service but didn't update the descriptions. Hopefully they are less confusing now!

spec/lib/brexit_checker/question_spec.rb
benthorner: Thanks for updating - it's a bit more clear. I realise I wasn't specific about what was confusing me. The thing I'm struggling with is really the contexts. For example, when I'm reading "previous question id is one and the criteria matches question three", I'm wondering "what does that really mean?", or "how would it appear to the user?". In this case, what we're testing is that we "skip questions with un-met criteria" - this means I'm learning about the overall functionality, and understanding the particular question number is just a detail.

Some other examples I thought of while reading the other contexts:

   - we "return nil when there are no further questions" (vs.)
   - we "return nil when the remaining questions have un-met criteria"

spec/helpers/brexit_checker_helper_spec.rb
theseanything: We need for the premise of the example on line 103 - i.e. a question has criteria, but it doesn't match so we skip it.

benthorner:
@theseanything thanks for giving the index suggestion a go. Perhaps it's worth separating the two aspects I'm concerned about here, as I think one of them should be easier to look at:

   - We should avoid the term 'ID' here, as questions already have a 'key' and introducing some other identifier is confusing, especially when it's so transient. If we do keep the concept, then can we consider changing it to 'number'? This would better reflect how we're interpreting it.
   - I was struggling to understand why the index approach wasn't working out. I've had a go at it myself, and I've [put something up](https://github.com/alphagov/finder-frontend/compare/update-page-service...update-page-service-ben?expand=1) as a demo, which hopefully looks simple enough - I like that it's a smaller change in total, and avoids changing how we think about the YAML file

I agree that specifying the number explicitly would clarify the origin of the new field. What I'm worried about there is that it would open the door to questions being harder to follow in the YAML file, since they could be written in a strange order, whereas an array ordering has a well-defined interpretation - I think it would be more of a smell if we were somehow reordering it in the code.

As a separate thing, we should think about what's going to happen to the 'page' param, since the main change in this PR is that the concept of 'page' is going, and now the param is referring to a question index. I suppose we could still call it 'page', but it seems unnecessarily vague. Do you have any thoughts on changing the param, perhaps in a subsequent PR?

theseanything:
> As a separate thing, we should think about what's going to happen to the 'page' param, since the main change in this PR is that the concept of 'page' is going, and now the param is referring to a question index. I suppose we could still call it 'page', but it seems unnecessarily vague. Do you have any thoughts on changing the param, perhaps in a subsequent PR?

The value of the parameter has always represented the previous question index - it was just poorly named parameter from the start. For the purpose of this PR - to maintain compatibility with existing URL the parameter name will remain `page`.

theseanything:
> @theseanything thanks for giving the index suggestion a go. Perhaps it's worth separating the two aspects I'm concerned about here, as I think one of them should be easier to look at:
> 
> * We should avoid the term 'ID' here, as questions already have a 'key' and introducing some other identifier is confusing, especially when it's so transient. If we do keep the concept, then can we consider changing it to 'number'? This would better reflect how we're interpreting it.
> * I was struggling to understand why the index approach wasn't working out. I've had a go at it myself, and I've [put something up](https://github.com/alphagov/finder-frontend/compare/update-page-service...update-page-service-ben?expand=1) as a demo, which hopefully looks simple enough - I like that it's a smaller change in total, and avoids changing how we think about the YAML file
> 
> I agree that specifying the number explicitly would clarify the origin of the new field. What I'm worried about there is that it would open the door to questions being harder to follow in the YAML file, since they could be written in a strange order, whereas an array ordering has a well-defined interpretation - I think it would be more of a smell if we were somehow reordering it in the code.

I've implemented using the question index - will look in subsequent PRs to further simplify the logic and explore the possibility adding a "number" attribute to questions.

spec/helpers/brexit_checker_helper_spec.rb
benthorner: I'm not sure what you mean here. Since my suggestion around the wording of the contexts is quite lengthy, perhaps we it would be easier to discuss in person?

app/controllers/brexit_checker_controller.rb
```diff
@@ -9,16 +9,16 @@ class BrexitCheckerController < ApplicationController
...
-    @page_service = BrexitChecker::PageService.new(questions: @questions,
-                                                criteria_keys: criteria_keys,
-                                                current_page_from_params: page)
-
-    if @page_service.redirect_to_results?
-      redirect_to brexit_checker_results_path(c: criteria_keys)
-    else
-      @current_question = @questions[@page_service.current_page]
-    end
+    all_questions = BrexitChecker::Question.load_all
```
benthorner: Minor thing: I was slightly surprising to see this here, mostly because we don't inject it in [other places](https://github.com/alphagov/finder-frontend/blob/master/spec/lib/brexit_checker/criteria/filter_spec.rb#L34). I'm wondering if there's a particular reason for injecting it here?

theseanything: Not sure what you mean by "it" here? We need to load all the questions to pass them to the `next_question_index` method?

PR #1570, opened 2019-09-18
Add actions S030 and S031

emmabeynon:
Fpr https://trello.com/c/u3qRc236/157-create-an-action-for-ukn-living-in-ir-and-eu-they-have-to-exchange-their-uk-driving-licence-s030-and-s031

### DIFF
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -1340,3 +1340,20 @@ actions:
       - living-eu
       - living-ie
   audience: citizen
+- action_id: S031
+  priority: 4
+  title: Test your vehicle for roadworthiness in the EU country you live in
+  title_url: https://www.gov.uk/guidance/driving-in-the-eu-after-brexit-driving-licence-exchange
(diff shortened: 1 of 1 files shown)


lib/brexit_checker/actions.yaml
```diff
@@ -1323,3 +1323,41 @@ actions:
...
+  guidance_link_text: 'Driving in the EU after Brexit: driving licence exchange'
+  guidance_url: https://www.gov.uk/guidance/driving-in-the-eu-after-brexit-driving-licence-exchange
+  criteria:
+  - any_of:
+    - all_of:
+      - nationality-uk
+      - living-eu
+    - all_of:
+      - nationality-uk
+      - living-ie
```
theseanything: Nit: this could be simplified to 
```yaml
- all_of:
  - nationality-uk
  - any_of:
    - living-eu
    - living-ie
```

theseanything: Same with action below

emmabeynon: Thanks - I've updated in the spreadsheet.

PR #1573, opened 2019-09-19
Add index attribute to Question

theseanything:
This PR poses a change which introduces the index attribute explicitly to the Question class.

Previously, we relied on array order to implicitly define question order. Adding an explicit reference to order in the Question class allows us to further simplify the logic resolve the question to be displayed and decouples the data structure from the underlying persistence layer.

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -38,10 +38,10 @@ module BrexitCheckerHelper
       hint_text: option.hint_text }
   end
 
-  def next_question_index(all_questions:, criteria_keys: [], previous_question_index: 0)
-    available_questions = all_questions[previous_question_index..] || []
+  def next_question(all_questions:, criteria_keys: [], previous_question_index: 0)
+    ordered_questions = all_questions.sort_by(&:index)
+    potential_questions = ordered_questions[previous_question_index..] || []
 
-    relative_next_question_index = available_questions.find_index { |question| question.show?(criteria_keys) }
-    relative_next_question_index ? previous_question_index + relative_next_question_index : nil
+    potential_questions.find { |question| question.show?(criteria_keys) }
   end
 end
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -13,13 +13,11 @@ class BrexitCheckerController < ApplicationController
 
   def show
     all_questions = BrexitChecker::Question.load_all
-    @question_index = next_question_index(
+    @current_question = next_question(
       all_questions: all_questions,
       criteria_keys: criteria_keys,
       previous_question_index: page
     )
-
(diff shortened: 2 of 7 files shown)


app/helpers/brexit_checker_helper.rb
```diff
@@ -38,10 +38,9 @@ def format_question_option(option, criteria_keys)
       hint_text: option.hint_text }
   end
 
-  def next_question_index(all_questions:, criteria_keys: [], previous_question_index: 0)
+  def next_question(all_questions:, criteria_keys: [], previous_question_index: 0)
     available_questions = all_questions[previous_question_index..] || []
```
benthorner: It still looks like we are relying on the order of the questions in the array at this point. To move away from that, we’ll need to reorder the questions by their index.

theseanything:
> I think we need to be clearer about the problem this is solving. What problem does the reliance of array order currently cause?
> 
> I’m also struggling to see this as a reduction in complexity, since this PR is actually a net addition to the codebase. I reckon this feature could be helpful as a way to make our configuration more flexible, so perhaps it’s worth focusing on that instead?

I'm coming from a view of tidying up the logic to make it easier for developers to understand how we resolve the next question. It also has the added bonus we don't have a concept that is tied to how we store the data.

I also would be wary about conflating "number of lines" with complexity. The variables `relative_next_question_index` and `question_index` add extra concepts that developer has to reason with.

app/helpers/brexit_checker_helper.rb
theseanything: 🤦‍♂ - yup!

benthorner:
Thanks for explaining your thoughts a bit more. I like that we're making an effort to tidy up our code, and you're right that lines != complexity - that was a low shot! I think I haven't been clear about the concerns I have with this change, so I'll try to outline them again:

   - For me, I don't find the current code difficult to follow. It's not clear to me who is struggling with the current code, and that this particular change will make things better for them
   - This will make it harder to add or reorder the questions in the YAML file, since a developer will now have to manually alter all of the indices to ensure they are still a sequence
   - This could also make understanding the order of the questions harder, as it's no longer guaranteed to be the top-to-bottom ordering in the YAML file, which I currently find helpful

I see your point that there are some non-trivial variables in the code we're talking about, which isn't ideal and perhaps we can think of better names for them. But removing them doesn't reduce the conceptual load, since we're also introducing a new concept (index). In a way, we've just moved the current concept somewhere else - sometimes that's a more natural fit and does indeed make the codebase more readable, but I don't think that's the case here.

As a way to proceed, I'd suggest getting someone who's having trouble understanding the current code to review this PR. We should also have a think about how much of a pain point maintaining the manual numbering is going to be, and whether we need to do anything to mitigate this as part of making this change.

benthorner:
@theseanything this one's been around for a while. Do you mind if we close?

PR #1577, opened 2019-09-19
Rename question and action attributes

theseanything:
This is a quick PR to tidy up keys in the YAML to prevent stutter and remove extra logic to deal with irregular mapping to object attributes.

### DIFF
--- a/lib/brexit_checker/action.rb
+++ b/lib/brexit_checker/action.rb
@@ -21,18 +21,12 @@ class BrexitChecker::Action
     BrexitChecker::Criteria::Evaluator.evaluate(criteria, selected_criteria)
   end
 
-  def self.load(params)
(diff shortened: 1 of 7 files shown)


lib/brexit_checker/action.rb
```diff
@@ -23,7 +23,6 @@ def show?(selected_criteria)
 
   def self.load(params)
```
benthorner: Can we just call `new` now and lose this method?

theseanything: Ah yeah - good shout!

theseanything:
> Yaaas! Just to check: we're updating the spreadsheet so the columns match the new CSV fixture?

Yup - I'll do the update as soon as this gets merged! :)

PR #1583, opened 2019-09-23
Add back link for questions

emmabeynon:
For https://trello.com/c/LVSlwXmo/246-add-back-link-for-questions

This adds a back link to the question pages in the Brexit Checker so that users can navigate to previous questions.  The exception to this is the first question which directs users to the 'start' page of the checker.  Since we now have 'Back' links on question pages, as per the design system we remove breadcrumbs from these pages.

We also ensure that the answers that a user has already selected are passed as params when they click the back link so that the answers are pre-selected on previous pages.  Answers that are changed on previous pages are carried forward to the results and previous answers are removed.

Screen Shot 2019-09-26 at 14 41 29

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -44,4 +44,9 @@ module BrexitCheckerHelper
     relative_next_question_index = available_questions.find_index { |question| question.show?(criteria_keys) }
     relative_next_question_index ? previous_question_index + relative_next_question_index : nil
   end
+
+  def previous_question_index(all_questions:, criteria_keys: [], current_question_index: 0)
+    previous_questions = all_questions[0...current_question_index]
+    previous_questions.rindex { |question| question.show?(criteria_keys) }
+  end
 end
--- a/spec/features/brexit_checker/question_navigation_spec.rb
+++ b/spec/features/brexit_checker/question_navigation_spec.rb
@@ -12,6 +12,7 @@ RSpec.feature "Navigating Brexit Checker questions", type: :feature do
 
   def when_i_visit_the_brexit_checker_flow
     visit brexit_checker_questions_path
+    expect(page).to have_link("Back", href: "/get-ready-brexit-check")
   end
 
   def and_i_answer_the_nationality_question
--- a/spec/features/brexit_checker/question_navigation_spec.rb
+++ b/spec/features/brexit_checker/question_navigation_spec.rb
@@ -1,13 +1,12 @@
 require "spec_helper"
 
 RSpec.feature "Navigating Brexit Checker questions", type: :feature do
-  scenario "navigate to previous page" do
+  scenario "navigate to previous question" do
     when_i_visit_the_brexit_checker_flow
-    and_i_answer_the_nationality_question
-    then_i_should_see_the_living_question
-
-    when_i_click_on_the_back_link
-    then_i_should_see_the_nationality_question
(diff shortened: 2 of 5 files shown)


spec/features/brexit_checker/question_results_spec.rb
```diff
@@ -25,6 +25,16 @@
     then_i_should_see_the_no_results_page
   end
 
+  scenario "navigate to previous page" do
```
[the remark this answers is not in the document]
emmabeynon: I've split this into a `question_navigation_spec.rb` file now.

emmabeynon:
@benthorner This should be ready to review again.

app/helpers/brexit_checker_helper.rb
```diff
@@ -44,4 +44,9 @@ def next_question_index(all_questions:, criteria_keys: [], previous_question_ind
     relative_next_question_index = available_questions.find_index { |question| question.show?(criteria_keys) }
     relative_next_question_index ? previous_question_index + relative_next_question_index : nil
   end
+
+  def previous_question_index(all_questions:, criteria_keys: [], current_question_index: 0)
+    previous_questions = all_questions[0...current_question_index]
+    previous_questions.rindex { |question| question.show?(criteria_keys) }
```
benthorner: Ruby can be so helpful sometimes!

spec/features/brexit_checker/question_navigation_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    when_i_visit_the_brexit_checker_flow
+    and_i_answer_the_nationality_question
+    then_i_should_see_the_living_question
+
+    when_i_click_on_the_back_link
+    then_i_should_see_the_nationality_question
+    and_my_previous_nationality_choice_is_selected
+  end
+
+  scenario "navigate to previous question with criteria requirements" do
```
benthorner: I'd suggest we only need one scenario here, as there's no difference in the visuals presented from the user's point of view, and it looks like we're covering another potential scenario (clicking back on the first question) in the unit tests. My other worry with these feature tests (including the ones we have already) is that they're already quite brittle, being coupled to the current structure of the questions; for this additional reason, I think we should minimise the number of scenarios where possible. Fortunately, our unit tests look solid for this.

app/helpers/brexit_checker_helper.rb
emmabeynon: Yep, `rindex` is a new one for me!

spec/features/brexit_checker/question_navigation_spec.rb
emmabeynon: Fair point 👍

DilwoarH:
With this change, can we support changing of answers? @benthorner @emmabeynon

emmabeynon:
@DilwoarH Yes some work was done previously to support changing answers so as long as we pass the criteria back to previous pages users can change their answers and this will be reflected in the results.

DilwoarH:
Seems to work as expected.

spec/features/brexit_checker/question_navigation_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "spec_helper"
+
+RSpec.feature "Navigating Brexit Checker questions", type: :feature do
+  scenario "navigate to previous question" do
```
benthorner: Just a minor thing: we conventionally drop the scenario text if there's only one scenario.

PR #1592, opened 2019-09-25
Add results page action count tracking

DilwoarH:
# What
A custom dimension that tracks numbers of results presented to users

# Why
If we knew how many results were presented it would help us work out how many users are seeing an unreasonably large number of results. One problem with the Business Readiness finder was that it often gave people an overwhelming number of results - we want to avoid that outcome.

---
Card created based on discussion on: https://trello.com/c/LyUDFNZE/191-discuss-tracking-with-performance-analysis-team

https://trello.com/c/hH1EGTGT/277-track-number-of-results-being-displayed-to-users


DilwoarH:
Ash has confirmed it is okay to use.

PR #1598, opened 2019-09-26
Add email tracking for dynamic list emails

DilwoarH:
Currently we send out notifications without any tracking on the links they contain. A given notification may have up to 2 links: the title of the action; and a link to some guidance. Note that some links are to external websites and we won't be able to track clicks on those.

After talking to Paul Cronk, we propose to manually inject the following query params into links that go to gov.uk:

   - utm_source = `<uuid of the change note>`
   - utm_medium = email
   - utm_campaign = govuk-brexit-checker

Normally we also set `utm_content = <subscription frequency>`, but we don't have the data to do that upfront. We've also changed the campaign from the normal `govuk-notifications`, which we should bear in mind in for other links in the email e.g. the link to the results list.

Normally an 'Update on GOV.UK' has a title link with tracking attributes automatically injected into it e.g.

https://www.gov.uk/government/news/government-backs-cutting-edge-tech-to-drive-down-shipping-emissions?utm_source=cb68c5c0-974f-46e3-a1db-2287917b17cd&utm_medium=email&utm_campaign=govuk-notifications&utm_content=immediate

Currently our notifications don't have a (title) link, and instead have other custom links in the freeform body, which don't benefit from this automatic injection mechanism.

https://trello.com/c/5UPZU8jc/242-collect-analytics-for-when-users-click-on-action-links-in-notifications

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -1,3 +1,5 @@
+require "addressable/uri"
+
 module BrexitCheckerHelper
   def format_criteria_list(criteria)
     criteria.map { |criterion| { readable_text: criterion.text } }
@@ -48,5 +50,18 @@ module BrexitCheckerHelper
   def previous_question_index(all_questions:, criteria_keys: [], current_question_index: 0)
     previous_questions = all_questions[0...current_question_index]
     previous_questions.rindex { |question| question.show?(criteria_keys) }
+  end
+
+  def change_note_email_link(non_tracked_url, change_note)
+    url = Addressable::URI.parse(non_tracked_url)
+    return non_tracked_url unless url.host == "www.gov.uk"
+
+    url.query_values = (url.query_values || {}).merge(
+      utm_source: change_note.id,
+      utm_medium: "email",
+      utm_campaign: "govuk-brexit-checker",
+    )
+
+    url.to_s
   end
 end
--- a/app/mailers/brexit_checker_mailer.rb
+++ b/app/mailers/brexit_checker_mailer.rb
@@ -1,4 +1,6 @@
 class BrexitCheckerMailer < ApplicationMailer
+  add_template_helper(BrexitCheckerHelper)
+
   def change_notification(change_note)
     @change_note = change_note
     @action = @change_note.action
--- a/app/views/brexit_checker_mailer/change_notification.text.erb
+++ b/app/views/brexit_checker_mailer/change_notification.text.erb
@@ -1,7 +1,7 @@
 <%= t("brexit_checker_mailer.change_notification.#{@change_note.type}") %>
 
 <% if @action.title_url.present? %>
-[<%= @action.title %>](<%= @action.title_url %>)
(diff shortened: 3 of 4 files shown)


lib/brexit_checker/action.rb
```diff
@@ -29,4 +29,14 @@ def self.load_all
     @load_all = nil if Rails.env.development?
     @load_all ||= YAML.load_file(CONFIG_PATH)["actions"].map { |a| new(a) }
   end
+
+  def add_email_tracking_to_url(non_tracked_url, change_note_id)
+    url = Addressable::URI.parse(non_tracked_url)
+    url.query_values = (callback.query_values || {}).merge({
```
benthorner: The `callback` is mysterious here - perhaps it's a feature of this particular library. To avoid confusion, can we do `url.query_values ||= {}` and then set the other params separately?

DilwoarH: sorry callback was a test variable - I'll update now.

DilwoarH:
> This is a good start @DilwoarH - I like that we're taking care to preserve any existing query params. I think there are a couple of aspects we should look at further before we get this in:
> 
> * Can we define the `add_email_tracking_to_url` as a (suitably renamed) helper, instead of adding it to the `Action` class? Normally, we avoid putting view-specific code into model classes.
> * One of the other problems we have is avoiding tracking for non-GOV.UK links - this is an AC on the card. Can we look at addressing that in this PR?

Hey @benthorner - I've updated based on your feed, please let me know if you have any further feedback.

app/helpers/brexit_checker_helper.rb
```diff
@@ -49,4 +51,16 @@ def previous_question_index(all_questions:, criteria_keys: [], current_question_
     previous_questions = all_questions[0...current_question_index]
     previous_questions.rindex { |question| question.show?(criteria_keys) }
   end
+
+  def change_note_email_link(change_note_id, non_tracked_url)
+    url = Addressable::URI.parse(non_tracked_url)
+    return non_tracked_url unless url.host == Plek.current.parent_domain
```
DilwoarH: This line doesn't seem to be working as expected. Will need to look into this.

lib/brexit_checker/change_notes.yaml
```diff
@@ -42,3 +42,7 @@ change_notes:
      note: "Content updated to say that you may need to apply for plant variety rights in both the UK and the EU. New arrangements will have to be agreed between the UK and the EU before seeds can be marketed in the EU after Brexit."
      action_id: T086
      date: 2019-09-26
+   - uuid: "5fe018d7-edb1-4d7a-858e-f758cbabc397"
+     type: addition
+     action_id: T092
+     date: 2019-09-09
```
theseanything: This needs to be removed before merge.

benthorner:
@DilwoarH I've updated this and it should be ready to go - how does it look?

DilwoarH:
> @DilwoarH I've updated this and it should be ready to go - how does it look?

Changes look good to me, thanks 👍

PR #1604, opened 2019-09-27
Add criteria rules to change notes

huwd:
https://trello.com/c/fMFEA6y9/294-support-custom-criteria-rules-for-change-notes

If criteria rules are added to change notes they overwrite the critera
rules from the action, used to determine who is notified of changes.

If no criteria rules are specified, defaults back to criteria rules
from action.


theseanything:
@huwd Just thinking we might want to restrict this to just "addition" type change notes. And we might want to support another type "deletion" i.e. for #1608

PR #1608, opened 2019-09-30
Update criteria for S001 and S003

theseanything:
This updates the criteria for S001 and S003 and remove irrelevant guidance for S003. No change note required.


benthorner:
After discussing with Martin and the Content people, we've agreed to get these updates out, while the decisions and work on 'criteria change notifications' are still being made/done.

PR #1615, opened 2019-10-02
T064

huwd:
Relates to: https://trello.com/c/1lvJh9Zi/297-t064-new-haulier-handbook-dft

Please note, this is a first content change PR from me!
Ruthless scrutiny appreciated.
I can confirm `bundle exec rspec spec/integration/brexit_checker_spec.rb` passes.

Questions:
- [x] [Did i generate the UUID correctly](https://github.com/alphagov/finder-frontend/pull/1615/files#r330619822)?
- [x] [Did i correctly understand where the note came from](https://github.com/alphagov/finder-frontend/pull/1615/files#r330620185)?
- [x] I [separated the commits](https://github.com/alphagov/finder-frontend/pull/1615/commits) for the typo change and the more substantive change. Not really sure why, kinda made sense to me. Should i squash these commits before merging or is that worth keeping apart?

[T064](https://trello.com/c/1lvJh9Zi/297-t064-new-haulier-handbook-dft) ticket seems to have a different template to previous [T086](https://trello.com/c/c3KOPd3Z/265-t086-seed-marketing-defra) Ticket. T086 specifically asks if users should be notified. There is no yes or no on this ticket.

- [x] Has there been a new standard template created since [T086](https://trello.com/c/c3KOPd3Z/265-t086-seed-marketing-defra)?
- [x] Is the absence of notification proof that we don't need one or should we always ask?

### DIFF
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -1007,11 +1007,12 @@ actions:
   title: Check if your customers can give you the right documents for trade with the
     EU, if you’re a haulier
   consequence: You may not be able to transport goods through the EU if your customers
-    do not give you the right documents
+    do not give you the right documents.
   lead_time: It takes more than 4 weeks
   guidance_prompt: Read the guidance
-  guidance_link_text: 'International road haulage: no-deal Brexit haulier checklist'
-  guidance_url: https://www.gov.uk/guidance/transport-goods-out-of-the-uk-by-road-if-the-uk-leaves-the-eu-without-a-deal-checklist-for-hauliers
+  guidance_link_text: 'Transporting goods between the UK and EU in a no-deal Brexit:
+    guidance for hauliers'
+  guidance_url: https://www.gov.uk/guidance/transporting-goods-between-the-uk-and-eu-in-a-no-deal-brexit-guidance-for-hauliers
   criteria:
   - any_of:
     - road-passenger-freight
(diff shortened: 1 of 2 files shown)


lib/brexit_checker/change_notes.yaml
```diff
@@ -37,3 +37,8 @@ change_notes:
      note: "Content updated to say that you may need to apply for plant variety rights in both the UK and the EU. New arrangements will have to be agreed between the UK and the EU before seeds can be marketed in the EU after Brexit."
      action_id: T086
      date: 2019-09-26
+   - uuid: "d2335a6a-ea96-4647-86f1-164cc1144964"
```
huwd: Wasn't sure where this might have come from.
I assumed these are uuids _just_ for a content change and so can be randomly generated.
So i yanked open a new terminal with pry in it and ran:
```
SecureRandom.uuid
```
Is that correct?

lib/brexit_checker/change_notes.yaml
```diff
@@ -37,3 +37,8 @@ change_notes:
...
+     type: content_change
+     action_id: T064
+     note: "Link changed to provide hauliers with updated guidance."
```
huwd: Took this straight off the language in the ticket.
Also checked a previous change which seemed to also take the `note` from the same place on a ticket.

lib/brexit_checker/actions.yaml
```diff
@@ -724,7 +724,11 @@ actions:
     deal Brexit
   guidance_url: https://www.gov.uk/guidance/fresh-fruit-and-vegetable-marketing-standards-if-the-uk-leaves-the-eu-without-a-deal
   criteria:
-  - forestry
+  - any_of:
```
benthorner: It looks like this criteria change has crept in as part of this PR. We should avoid making changes that are unrelated to the card (i.e. T064), as they may not be completed yet, and makes future reasoning about the work we did hard. Can we revert this part of the change for now?

huwd: Yep, will do!

huwd: ah ha, so the CSV can cover multiple changes and we need to `git add -p` just out the ones related to a ticket?

PR #1623, opened 2019-10-03
Change priority of bus document actions (T078 & T034)

huwd:
A part of: https://trello.com/c/owxTSb2l/298-t063-and-t078-scheduled-and-unscheduled-coach-services-dft

- Change priority of T078
- Change prioirty of T034


huwd:
Done

PR #1628, opened 2019-10-04
Rename ChangeNote to Notification

huwd:
## Background
Change notes have been used in the application as a log of changes we will notify subscribers about.

Each ones also has an optional 'note' that content can use to describe the nature of the change to subscribers.

This has generated historic confusion for developers and makes these concepts difficult for new starters to instantly grasp.

## Pain point example
Pain point here arose in relation to [this ticket](https://trello.com/c/fMFEA6y9/294-send-separate-notifications-when-criteria-change), when writing up the documentation there were several points of confusion around the difference between a `change_note` and note describing a change _on a change note_

## Change
As 'Change notes' are really a log of notified changes. This PR renames all references to 'change_note' to "notification'

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -52,12 +52,12 @@ module BrexitCheckerHelper
     previous_questions.rindex { |question| question.show?(criteria_keys) }
   end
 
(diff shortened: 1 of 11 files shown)


huwd:
Done and done!

PR #1643, opened 2019-10-10
Show search spelling suggestions

kr8n3r:
**_Back by popular demand!_**

We used to have spelling suggestions. We [blogged](https://insidegovuk.blog.gov.uk/2015/12/30/site-search-improvements-quick-wins/) about them as well

This works enables spelling suggestions on search/all finder only.

Screen Shot 2019-09-13 at 15 43 12

Changes made to design 
image

image

---

- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/search/all
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/get-ready-brexit-check/questions
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/uk-nationals-living-eu
- http://finder-frontend-pr-[THIS PR NUMBER].herokuapp.com/prepare-business-uk-leaving-eu


kr8n3r:
note: pr extracts page variables in it's own method so we need to update the controller here  once that gets merged https://github.com/alphagov/finder-frontend/pull/1656/commits/6effc0a8ac8761d8270c24c1bd34eebbfa83417c

app/views/finders/_show_header.html.erb
```diff
@@ -32,6 +32,15 @@
             value: result_set_presenter.user_supplied_keywords,
           } %>
         </div>
+        <div id="js-spelling-suggestions" class="spelling-suggestions" 
+          data-analytics-ecommerce
+          data-ecommerce-start-index="<%= result_set_presenter.start_offset %>"
```
barrucadu: Is it useful to track what page the user is on?  Spelling suggestions are a separate list to search results

app/controllers/finders_controller.rb
```diff
@@ -141,13 +144,14 @@ def search_results
...
-        keywords: keywords,
-        link: finder_url_builder.url(keywords: keywords),
-      }
-    end
+  def spelling_suggestion_presenter
+    suggested_queries = search_results.fetch("suggested_queries", [])
+    SpellingSuggestionPresenter.new(
+      suggested_queries,
+      finder_url_builder.url(keywords: suggested_queries.first),
+      #todo maybe back and make sure that search api returns only 1 thing and not and array with one item
```
barrucadu: ?

kr8n3r: search-api looks to always return an array, even as there is always one item. @laurentqro suggested a pr in search api to update that, but check with other devs first. 
that's why we left a comment, but i forgot to check 😊

laurentqro: I guess it all depends on how many spelling suggestions we want to show users, and if so, which one. So it might actually be quite a bit of work.

app/views/finders/_show_header.html.erb
kr8n3r: do you mean it's missing and we need to, or it's there but it's redundant?

app/controllers/finders_controller.rb
barrucadu: search-api is set up to return at most one suggestion, so I agree an array is a bit unnecessary.

But maybe once upon a time someone thought it was more convenient to have the value always returned, and have consumers check if the array is empty or not, than for consumers to check if the value exists or not.  Returning an array *does* mean we could return more suggestions in the future without much change

app/views/finders/_show_header.html.erb
barrucadu: I think the `data-ecommerce-start-index` is redundant for spelling suggestions

kr8n3r: that determines the position in the list, if i take it out, it says 
Impression List 1 Product 1 Position : NaN
I can just say 1 - as it's just one for now

app/controllers/finders_controller.rb
kr8n3r: I've updated the comment, happy to take suggestions for code to be updated to accept an array  and loop over, instead of doing `.first`

app/views/finders/_show_header.html.erb
barrucadu: I think just saying `1` would be good, as `start_offset` will be change if the user goes to a different page, but the page they're on doesn't have any bearing on the spelling suggestion they get.

kr8n3r: cool, updated

PR #1654, opened 2019-10-14
Minor change to T068 text

huwd:
Relates to: https://trello.com/c/aspUdbOk/81-t068-clinical-trials-dhsc-mhra-approved-11-oct


huwd:
> Looks good, although this does seem like it might be a major update. Are we sure it doesn't need a change note?

Yeah good shout, I checked with Liz.
Turns out the new link is really just to the parent page of the previous link. So both that and text fall under the header of "Clarifying guidance" or making existing guidance easier to access.

So no major changes or notifications required.

PR #1656, opened 2019-10-15
Error messages on date filter

hannako:
We don't currently display error messages to a user if they enter a date that we cannot parse.

1.  Adds two fields to the JSON response so that the livesearch function that replaces the page with fresh search results is aware that the date filter is incorrect. 

    On the frontend if there is an error on page render and the finder has collapsed filters, we click 
    on the toggle and focus on the input with the error.

   On live search we check for error fields in the JSON and either construct or remove the error 
   from the fields.

## Before:
Screen Shot 2019-10-15 at 10 54 15

## After:
Screenshot 2019-11-08 at 09 10 43

2. Allows error message to be displayed if date params are bad. This is required because the `livesearch.js` function does not run when the page is first loaded. It's a bit belt and braces as probably quite rare for a link with a bad date param to be shared.

Screenshot 2019-11-08 at 09 11 38

[Trello card](https://trello.com/c/cO3nIze1/1052-spike-investigate-validation-on-user-input-on-date-filter-date-entry)

---

- https://finder-frontend-pr-1656.herokuapp.com/search/all
- https://finder-frontend-pr-1656.herokuapp.com/search/research-and-statistics
- https://finder-frontend-pr-1656.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-1656.herokuapp.com/get-ready-brexit-check/questions
- https://finder-frontend-pr-1656.herokuapp.com/drug-device-alerts
- https://finder-frontend-pr-1656.herokuapp.com/find-eu-exit-guidance-business
- https://finder-frontend-pr-1656.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- https://finder-frontend-pr-1656.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- https://finder-frontend-pr-1656.herokuapp.com/uk-nationals-living-eu
- https://finder-frontend-pr-1656.herokuapp.com/prepare-business-uk-leaving-eu


kr8n3r:
as a note so it's not forgotten: filters are collapsed on page load, so we'll need to toggle the "Show more search options" with JS if there's an error

kr8n3r:
@hannako are we able to combine errors into a single property in the json? 
say 
```
public_timestamp_errors : {
from:true,
to: true
}
```

kr8n3r:
> @hannako are we able to combine errors into a single property in the json?
> say
> 
> ```
> public_timestamp_errors : {
> from:true,
> to: true
> }
> ```
> 
> this will make it simpler to check for error and manipulate the correct element

we might need to update the property name to take into account that the actual application property is called differently on different finders

kr8n3r:
i've rebase with master and [started writing a cucumber](https://github.com/alphagov/finder-frontend/pull/1656/commits/5e083b7326bbe73c42a8da7db97821d97508c025) test but got stuck

hannako:
@laurentqro How would that be better? We would still have to have a custom validator, ie the code currently in the date errors presenter. JSON response method also would have to access the error message(s) but only the specific date errors. Feels like it would add complexity?

app/assets/javascripts/live_search.js
```diff
@@ -385,5 +388,52 @@
...
+      errors[prop] ? this.buildErrorMessage(prop) : this.removeErrorMessage(prop)
+    }
+  }
+
+  LiveSearch.prototype.buildErrorMessage = function (type) {
+    var $input = this.$form.find('input[name*="[' + type + ']"]')
+    var errorMessageElement = $('<span />', {
+      id: 'error-' + type,
+      class: 'gem-c-error-message govuk-error-message',
+      html: '<span class="govuk-visually-hidden">Error:</span> Enter a real date</span>'
```
barrucadu: There are two closing tags here but only one opening tag

app/assets/javascripts/live_search.js
```diff
@@ -385,5 +388,52 @@
...
+    })
+
+    // only attach the error message if not present
+    if ($input.siblings('.gem-c-error-message').length <= 0) {
```
barrucadu: Can this be `< 0`?

app/assets/javascripts/live_search.js
kr8n3r: good spot 👍

app/assets/javascripts/live_search.js
kr8n3r: it can be exactly 0 though. updated

app/controllers/finders_controller.rb
```diff
@@ -68,6 +68,7 @@ def json_response
       sort_options_markup: render_component("finders/sort_options", sort_presenter.to_hash),
       next_and_prev_links: render_component("govuk_publishing_components/components/previous_and_next_navigation", pagination_presenter.next_and_prev_links),
       suggestions: render_component("finders/spelling_suggestion", suggestions: spelling_suggestion_presenter.suggestions),
+      public_timestamp_errors: search_query.date_errors_hash,
```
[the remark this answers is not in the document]
hannako: Good idea, thanks. Will do that.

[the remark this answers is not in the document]
hannako: Good idea thanks will do that

app/validators/date_validator.rb
```diff
@@ -0,0 +1,53 @@
...
+  # The date facet on search/all filters on public_timestamp, but the specialist
+  # finders use a variety of different types. Error messages are
+  # currently only implemented on search/all.
+
+  def date_type
+    :public_timestamp
+  end
+
+  def invalid_date?(user_input)
+    user_input.present? && DateParser.parse(user_input).nil?
```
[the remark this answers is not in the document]
hannako: Thanks

app/assets/javascripts/live_search.js
```diff
@@ -382,5 +385,52 @@
...
+    var $facetToggle = $container.find('.facet-toggle')
+    var isfacetToggleExpanded = $facetToggle.attr('aria-expanded')
+    var $inputWithError = $container.find('input[class*=--error]')
+    if ($facetToggle.length > 0 && isfacetToggleExpanded === 'false' && $inputWithError.length > 0) {
+      $facetToggle.click()
+      $inputWithError.focus()
+    }
+  }
+
+  LiveSearch.prototype.checkForJsonErrorMessage = function (errors) {
```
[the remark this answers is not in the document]
kr8n3r: sounds good

app/assets/javascripts/live_search.js
```diff
@@ -382,5 +385,52 @@
     return ''
   }
 
+  LiveSearch.prototype.dealWithInputErrorOnLoad = function ($container) {
+    var $facetToggle = $container.find('.facet-toggle')
+    var isfacetToggleExpanded = $facetToggle.attr('aria-expanded')
+    var $inputWithError = $container.find('input[class*=--error]')
+    if ($facetToggle.length > 0 && isfacetToggleExpanded === 'false' && $inputWithError.length > 0) {
+      $facetToggle.click()
+      $inputWithError.focus()
```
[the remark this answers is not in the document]
kr8n3r: they are scoped, but will have a look if there's a cleaner way to do it

app/assets/javascripts/live_search.js
```diff
@@ -382,5 +385,52 @@
...
+  }
+
+  LiveSearch.prototype.checkForJsonErrorMessage = function (errors) {
+    if (!errors) return
+    for (var prop in errors) {
+      errors[prop] ? this.buildErrorMessage(prop) : this.removeErrorMessage(prop)
+    }
+  }
+
+  LiveSearch.prototype.buildErrorMessage = function (type) {
```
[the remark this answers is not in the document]
kr8n3r: 👍

app/controllers/finders_controller.rb
```diff
@@ -20,9 +20,9 @@ def show
         @search_query = initialize_search_query
         if content_item.is_search? || content_item.is_finder?
           @spelling_suggestion_presenter = spelling_suggestion_presenter
-          render json: json_response
+          render json: json_response, status: response_status
```
hannako: Hey @bilbof If we raise a 400 the JSON fetch fails :-( Is there another way I can log that the parameters are bad other than by using a status code?

hannako: Would statsd be a viable alternative?

bilbof: I think that's probably the result of the current Ajax error handling:

```
      return $.ajax({
        url: this.action,
        data: this.state,
        searchState: searchState
      }).done(function (response) {
        liveSearch.cache($.param(liveSearch.state), response)
        liveSearch.displayResults(response, this.searchState)
      }).error(function () {
        liveSearch.showErrorIndicator()
      })
```

In the `.error` callback you could handle the 400 status code, and check whether there's an error message to be rendered for an input field.

hannako: Thanks @bilbof 

I discussed that option with @kr8n3r earlier and the concern was that there might be legitimate reasons for a 400 meaning we shouldn't do the fetch.

Maybe I could raise a more specific 4xx and handle just that one type?

hannako: Oh sorry @bilbof I misread your comment. Yep good idea thank you.

bilbof: In the error callback, it's mostly been handling 500 errors from search-api timing out. So you'd need to change this to handle the 400 error and keep the current behaviour for other status codes.

That might look like this:

```
      }).error(function (response) {
        if (response.status == 400) { return handleInputError() }
        liveSearch.showErrorIndicator()
      })
```

Is there a more specific error code you had in mind?

kr8n3r: at the moment we provide a json fail regardless of the 400 error, as the results still come back from the api
if we were to read the error message for the field in the JSON, we'd still need the status to be 200.

unless i'm understanding the requirement incorrectly

bilbof: The response body can still be JSON even if the status code is not 200. If the status code is 400 jQuery will call the `error` callback, so `error` handles the response. Within the `error` callback you can treat the response in the same way as you treat it in the `success` handler.

kr8n3r: sure, we can do that. let me think how to handle that and other 400 cases

kr8n3r: i'm now battling with the state not being updated

hannako: Does that matter? It's updating if it's valid. Maybe that's fine??

kr8n3r: when firing 400 status, ajax does not return a new state back so the function that updates the url does not fire for this or subsequent (valid) ajax requests until the fields with the error is fixed, leaving a stale url

hannako: @kr8n3r 

Following pairing effort with @bilbof it looks like error handling is happening [upstream in Slimmer](https://github.com/alphagov/slimmer/blob/b56bd87a09824262aac298490e0d14122c86df5c/lib/slimmer/app.rb#L85) 
ie when we return a 400 if the search query is invalid, slimmer decides that we show the default gov.uk 400 error page. We can skip slimmer, but then we also lose the headers and formatting of the page that slimmer brings over from static. 

Decision: Return 200's for both the application and JSON response.

bilbof:
I get some JS errors when loading the page, which prevents me from showing facets:

Do you see the same thing? http://finder-frontend-pr-1656.herokuapp.com/search/all

Screenshot 2019-11-12 at 14 11 16


```
Uncaught TypeError: Cannot read property 'version' of null
    at window.GOVUK.Modules.GlobalBar.start (header-footer-only-06d2bb189e4062967da9ced4270c6d23ab4c20eb6d2928b3ab20f37e3dfbbaef.js:1)
    at Object.start (application-40ab8219b3d6eb7908638d0bf6587970c51f7e9163fa2d5de7f67d76fd621c6f.js:4)
    at HTMLDocument.<anonymous> (header-footer-only-06d2bb189e4062967da9ced4270c6d23ab4c20eb6d2928b3ab20f37e3dfbbaef.js:2)
    at u (jquery-1.12.4-c731c20e2995c576b0509d3bd776f7ab64a66b95363a3b5fae9864299ee594ed.js:2)
    at Object.fireWith [as resolveWith] (jquery-1.12.4-c731c20e2995c576b0509d3bd776f7ab64a66b95363a3b5fae9864299ee594ed.js:2)
    at Function.ready (jquery-1.12.4-c731c20e2995c576b0509d3bd776f7ab64a66b95363a3b5fae9864299ee594ed.js:2)
    at HTMLDocument.o (jquery-1.12.4-c731c20e2995c576b0509d3bd776f7ab64a66b95363a3b5fae9864299ee594ed.js:1)
```

hannako:
@bilbof Yours is a weird link missing the `s` on `http`

https://finder-frontend-pr-1656.herokuapp.com/search/all seems to be working fine

PR #1658, opened 2019-10-15
Exclude living question choice from travel options

emmabeynon:
https://trello.com/c/IwUKz9tE/80-spike-travel-question-is-there-a-solution-thats-not-just-not-logic

Users have fed back that it is confusing to see the place that they live as an
option when asked where they intend to travel to. This captures the option that
a user selects when asked where they live and removes that location as an
option when they see the travel question. Since this is a presentation issue,
it made sense to construct the options that are displayed in a helper that can
be used by the view.

### DIFF
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -27,6 +27,8 @@ class BrexitCheckerController < ApplicationController
       current_question_index: page,
     )
 
+    @living_criterion = find_living_criterion
+
     redirect_to brexit_checker_results_path(c: criteria_keys) if @current_question.nil?
   end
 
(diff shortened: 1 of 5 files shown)


koetsier:
Could we add something to the questions.yaml file to encapsulate this logic? i.e.

''' - key: travelling
    type: multiple
    text: 'Where do you plan to travel after 31 October 2019?'
    description: |
      <p>This includes travel that may have started on or before 31 October 2019.</p>
    hint_text: Select all that apply. If you do not plan to travel, select next.
    options:
      - label: To the UK
        value: visiting-uk
        exclude_if: living-uk
      - label: To Ireland
        value: visiting-ie
        exclude_if: living-ie
      - label: To another EU country, Iceland, Liechtenstein, Norway or Switzerland
        value: visiting-eu
        exclude_if: living-eu
      - label: To the rest of the world
        value: visiting-row
        exclude_if: living-row
'''

That might make it a bit more generic.

koetsier:
Reworked into: https://github.com/alphagov/finder-frontend/pull/1676

PR #1669, opened 2019-10-22
Clarify brexit checker business questions

huwd:
Relates to: https://trello.com/c/K8jXUOLz/58-update-business-questions-flow-in-checker

---

Last point to clarify:
[This change](https://github.com/alphagov/finder-frontend/pull/1669/files#r337433192) has resulted in a change of criteria. Confirmed with mark it needs to be here.
Therefore we'd need to make changes to existing SQL. @huwd to confirm with @benthorner and @koetsier how and when that should happen.

---

Changes are intended to do three things:
1) User research has shown that business questions need to be "hardened",
we need to clarify that this only applies to UK businesses, as this has
been a reoccurring point of confusion in user testing.

2) User research shows that Users are often not ticking that they "Sell or
provide services in the UK" even when it should apply to them. If we have
clarified (1), we can assume that this is true and reduce the friction.

3) The descriptive text has proven more confusing than helpful, so we're
removing it

### DIFF
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -337,8 +337,7 @@ actions:
   guidance_url: https://www.gov.uk/guidance/prepare-to-use-the-ukca-mark-after-brexit
   criteria:
   - any_of:
-    - sell-goods-provide-services-in-uk
-    - export-to-eu
+    - owns-operates-business-organisation
   audience: business
 - id: T002
   priority: 8
@@ -392,7 +391,6 @@ actions:
   - any_of:
     - sell-public-sector
     - sell-public-sector-contracts
-    - sell-goods-provide-services-in-uk
   audience: business
 - id: T006
   priority: 8
@@ -482,7 +480,6 @@ actions:
   guidance_url: https://www.gov.uk/government/publications/exhaustion-of-intellectual-property-rights/exhaustion-of-intellectual-property-rights
   criteria:
   - any_of:
-    - sell-goods-provide-services-in-uk
     - import-from-eu
     - export-to-eu
     - provide-services-do-business-in-eu
@@ -752,7 +749,6 @@ actions:
   criteria:
   - any_of:
     - food-drink-tobacco
-    - sell-goods-provide-services-in-uk
     - export-to-eu
   audience: business
 - id: T038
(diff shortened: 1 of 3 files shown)


lib/brexit_checker/actions.yaml
```diff
@@ -337,7 +337,7 @@ actions:
   guidance_url: https://www.gov.uk/guidance/prepare-to-use-the-ukca-mark-after-brexit
   criteria:
   - any_of:
-    - sell-goods-provide-services-in-uk
+    - owns-operates-business-organisation
```
huwd: Need to check implications of this.

benthorner: Hmm...this is suspicious. If we have this criteria, we shouldn't need the export one, since it's not possible to have the latter without the former 🤔.

huwd: That's a great point, I guess this is implicit by virtue of question criteria.
Let me check in again with Mark, not sure if this choice is about the results page.

huwd: @benthorner Ah ha ok, caught up with mark so the criteria goes something like this.
Given that you _do_ own or operate a business in the UK
And that you _do not_ export to the eu

You still need to check CE guidance as there could be a new UK CA stamp that could be relevant.

huwd: Guidance is "In some cases you will need to apply the UKCA marking to goods being sold in the UK immediately after Brexit"

benthorner: I'm still confused, but maybe I wasn't clear before. I'll try explain slightly differently: 'owns-operates-...' is a superset of 'export-to-eu', since it's not possible to 'export-to-eu' unless you 'owns-operates-...'. Given this action uses an `any_of` predicate, it's not necessary to have 'export-to-eu'. 

I notice that the English version in your comment above is different to the actual criteria expression, so maybe someone needs to fix that?

huwd: Could I gather you, mark and I together in person on this one?
I think we're losing quite a bit in translation?

huwd: Have amended the copy above so it's an exact quote of the guidance and not my sloppy typing of what was being read out loud

huwd: Ok resolved. We do not need the exporting to EU. 
And as this enlarges the scope of notifications we won't need to change any subscriptions in SQL

PR #1675, opened 2019-10-22
Update action language to match the landing page

vanitabarrett:
Update the actions with the new content to match the landing page. This should only be rewording of consequences, nothing else.

See the highlighted boxes on [this spreadsheet](https://docs.google.com/spreadsheets/d/1D-QAXlYgytZHvWVKjgk5hDr79se9Quw6PaGp2wfp20A/edit#gid=0) for the consequences that should have changed.

### DIFF
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -109,8 +109,8 @@ actions:
   priority: 8
   title: Check a passport for travel to Europe after Brexit
   title_url: https://www.gov.uk/check-a-passport-for-travel-to-europe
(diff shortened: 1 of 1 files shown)


lib/brexit_checker/actions.yaml
```diff
@@ -1484,21 +1487,6 @@ actions:
   criteria:
   - chemical
   audience: business
-- id: T096
```
benthorner: Are we sure we want to delete this?

benthorner: Never mind!

PR #1676, opened 2019-10-22
Add exclude_if to options

koetsier:
co-author @emmabeynon 

Add 'exclude_if' on options so we can exclude an option

Adds a very simple optional 'exclude_if' to options in the questions.
This stops an option from showing if the given value has been
previously selected by the user.

This commit uses this functionality to update the questions so that
users are no longer asked if they want to travel to the place
they have indicated they live.

Removed functionality to only show an option if a criterion is present
as we're not using this.

trello: https://trello.com/c/IwUKz9tE/80-spike-travel-question-is-there-a-solution-thats-not-just-not-logic

image
image

- http://finder-frontend-pr-1676.herokuapp.com/get-ready-brexit-check/questions

### DIFF
--- a/docs/brexit-checker.md
+++ b/docs/brexit-checker.md
@@ -117,7 +117,7 @@ A option has the following attributes:
 | value       | string           | criteria key in which to apply to the user if the option is selected                                            | yes      |
 | sub_options | array of Options | additional options that are available if option is selected, primarily used for the 'single_wrapped' type questions | no       |
 | hint_text   | string           | addition context to help the user understand the option                                                         | no       |
-| criteria    | Criteria Rule    | defines which sets of criteria the option should be shown                                                       | no       |
+| exclude_if  | string           | the option will not be shown if the given criterion has been selected.                                                    | no       |
 
 ### Notification
 A notification is used to record changes to actions and to send emails to relevant subscribers.
--- a/spec/features/brexit_checker/travel_question_spec.rb
+++ b/spec/features/brexit_checker/travel_question_spec.rb
@@ -1,6 +1,8 @@
 require "spec_helper"
 
-RSpec.feature "Brexit Checker travel question", type: :feature do
+RSpec.feature "Filtering options based on criteria", type: :feature do
+  include BrexitCheckerHelper
+
   scenario "User's living question choice is not displayed as a travel question option" do
     when_i_visit_the_brexit_checker_flow
     and_i_choose_uk_for_the_living_question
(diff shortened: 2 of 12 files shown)


docs/brexit-checker.md
```diff
@@ -117,7 +117,7 @@ A option has the following attributes:
 | value       | string           | criteria key in which to apply to the user if the option is selected                                            | yes      |
 | sub_options | array of Options | additional options that are available if option is selected, primarily used for the 'single_wrapped' type questions | no       |
 | hint_text   | string           | addition context to help the user understand the option                                                         | no       |
-| criteria    | Criteria Rule    | defines which sets of criteria the option should be shown                                                       | no       |
+| exclude_if  | string           | the option will not be shown if the given criterion has been selected.                                                    | no       |
```
huwd: So this replaced criteria? There continue to be criteria on questions but not on options?

spec/features/brexit_checker/travel_question_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "spec_helper"
+
+RSpec.feature "Brexit Checker travel question", type: :feature do
+  scenario "User's living question choice is not displayed as a travel question option" do
```
benthorner: Just a minor thing, but the phrasing sounds like the 'feature' is a 'bug'! How about 'filter options by location'?

spec/features/brexit_checker/travel_question_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    click_on "Next"
+  end
+
+  def then_i_should_not_see_uk_as_an_option_for_the_travel_question
+    question = BrexitChecker::Question.find_by_key("travelling")
+    expect(page).to have_content(question.text)
+    expect(page).not_to have_content("To the UK")
+  end
+
+  def answer_question(key, *options)
```
benthorner: This is the third duplicate of this helper. Should we consider refactoring it centrally?

spec/features/brexit_checker/travel_question_spec.rb
koetsier: 👍

spec/features/brexit_checker/travel_question_spec.rb
koetsier: refactored this 👍

spec/features/brexit_checker/brexit_checker_spec_helper.rb
```diff
@@ -0,0 +1,8 @@
+module BrexitCheckerSpecHelper
```
benthorner: Thanks @koetsier +1. Helpers generally go in `spec/support`, though.

koetsier: 👍 saves a ```require``` as well

PR #1684, opened 2019-10-24
Change copy for T003 EU funded projects

huwd:


huwd:
Yep, we checked apparently they had a conversation about it and decided this made most sense.

PR #1691, opened 2019-10-28
Hacky prototype for dynamic facets

barrucadu:
This makes *some* facets dynamic - ie, they only contain values which, when selected, will return at least 1 additional result.

I thought that elements disappearing from the page could be confusing, which has some consequences:

- The brexit checkbox on https://www.gov.uk/search/news-and-communications should always be there.
- The radio buttons on https://www.gov.uk/search/research-and-statistics should always be there
- Invalid values can be removed from the topic, sub-topic, organisations, people (etc) facets but the component should always be drawn, even if it has nothing which can be selected.

This PR makes registry-backed facets (organisations, people, etc) dynamic.  Things which would be nice to be dynamic but which are trickier are:

- Facets which have a list of allowed values (eg, document type in https://www.gov.uk/search/policy-papers-and-consultations)
- The topic and sub-topic facets.

However, those facets have a relatively small number of options available.

Topic/sub-topic would require changes in search-api I think, as search-api doesn't treat first- and second-level taxons specially, so we can only aggregate over the entire taxonomy, which is too much.

This PR implements it in a very hacky way, detailed in d9fff518562816a142248b434154815f38025dfb, and I'd need some help from a frontend developer to make it nice.

---

- http://finder-frontend-pr-1691.herokuapp.com/search/all
- http://finder-frontend-pr-1691.herokuapp.com/search/research-and-statistics
- http://finder-frontend-pr-1691.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- http://finder-frontend-pr-1691.herokuapp.com/get-ready-brexit-check/questions
- http://finder-frontend-pr-1691.herokuapp.com/drug-device-alerts
- http://finder-frontend-pr-1691.herokuapp.com/find-eu-exit-guidance-business
- http://finder-frontend-pr-1691.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- http://finder-frontend-pr-1691.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- http://finder-frontend-pr-1691.herokuapp.com/uk-nationals-living-eu
- http://finder-frontend-pr-1691.herokuapp.com/prepare-business-uk-leaving-eu

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -14,6 +14,7 @@
     this.$resultsWrapper = this.$form.find('.js-live-search-results-block')
     this.$suggestionsBlock = this.$form.find('#js-spelling-suggestions')
     this.$resultsBlock = options.$results.find('#js-results')
+    this.$facetWrapper = this.$form.find('#facet-wrapper')
     this.$countBlock = options.$results.find('#js-result-count')
     this.$facetTagBlock = options.$results.find('#js-facet-tag-wrapper')
     this.$loadingBlock = options.$results.find('#js-loading-message')
@@ -116,7 +117,6 @@
   LiveSearch.prototype.formChange = function formChange (e) {
     var pageUpdated
     if (this.isNewState()) {
-      this.getTaxonomyFacet().update()
       this.saveState()
       this.updateOrder()
       this.updateLinks()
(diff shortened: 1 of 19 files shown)


kr8n3r:
when we do this properly I reckon we split `facet_collection` partial into 2 parts:
- furniture
- facets_only

That way we can reuse the new facets_only to generate the json data without needing to update the rest of the markup (search-box, show/hide toggle) as we do now.

kr8n3r:
> Message doesn't look vertically centred :(
> 
> Screenshot 2019-10-31 at 15 36 53

there is some magic javascript that sets the height of that container and cuts half of the checkbox so that users see that more options are available.

i could look into using css grid or flexbox as an styling enhancement

PR #1700, opened 2019-10-30
[Do not merge] Change EU/UK funding question

huwd:
[Trello card](https://trello.com/c/Yvl6gb8v/139-eu-programme-funding-update-funding-question-on-checker)

---

Reverts alphagov/finder-frontend#1698
Which itself reverted the initial version of this change.

---

Leaving this here as a basis for whatever we decide to do instead.


huwd:
replaced by: https://github.com/alphagov/finder-frontend/pull/1704

PR #1702, opened 2019-10-30
Add optional explanitory detail to questions

huwd:
[Trello card](https://trello.com/c/0fa05Nes/179-help-users-to-understand-why-they-are-being-asked-questions-in-the-checker)

---

Replaces previous use of 'hint' detail components, that are not currently used as no hint titles exist in questions.yaml


app/views/brexit_checker/_detail_text.html.erb
```diff
@@ -0,0 +1,7 @@
+<% unless detail_text.nil? %>
+  <%= render "govuk_publishing_components/components/details", {
+    title: detail_title  || "Why am I being asked this question?"
```
huwd: @vanitabarrett thanks for this idea, i think it makes a lot of sense.
Does this look good to you right now?

vanitabarrett:
Note: we're going to deploy this change without tracking to start with, as we're under time pressure to get this deployed. There is an [open PR](https://github.com/alphagov/govuk_publishing_components/pull/1187) to allow the details component to support tracking, but we're having a few issues with the tests which need investigating. As soon as those tests are fixed, we can come back and add tracking to this.

PR #1704, opened 2019-10-30
Change content and simplify criteira for funding Q

huwd:
Trello Card: https://trello.com/c/Yvl6gb8v/139-eu-programme-funding-update-funding-question-on-checker

---

Replaces [previous work here](https://github.com/alphagov/finder-frontend/pull/1700)

---

- Simplify criteria for funding question
- Adjust content


lib/brexit_checker/questions.yaml
```diff
@@ -205,15 +205,10 @@ questions:
     type: "single_wrapped"
```
benthorner: Should we change the question type here?

huwd: Ah interesting, what does question type do?
Because this has switched to radio buttons.

benthorner: It's probably a minor thing, as I can't think of any objective issue with leaving as-is. Perhaps we should condense the question types here: https://github.com/alphagov/finder-frontend/blob/master/app/views/brexit_checker/show.html.erb#L58

huwd: Should I bung this in the backlog? or a ticket on planning?

benthorner: I'd say only if it's causing actual pain in development. Not sure if that's the case?

PR #1714, opened 2019-11-04
Reverse actions array, so business actions come first

huwd:
Quick request by Mark and Stephen.
Would need to go live before PEP starts

~I'm unsure if using `.reverse` like that is a terrible hack (currently there are only two audiences, not impossible there are more in future!).~  

Turns out the linter doesn't like me chaining `end.reverse`, gone for a slightly more verbose method though I'm _still_ not sure if that's a terrible hack. A proper sort method would make sense, but I'm wondering if it's overkill?

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -6,15 +6,21 @@ module BrexitCheckerHelper
   end
 
   def format_action_audiences(actions)
-    action_groups = actions.group_by(&:audience)
+    business, citizen = actions.partition { |action| action.audience == "business" }
+    business_results = {
+      heading: I18n.t("brexit_checker.results.audiences.business.heading"),
+      actions: order_actions_by_priority(business),
+    }
+    citizen_results = {
+      heading: I18n.t("brexit_checker.results.audiences.citizen.heading"),
+      actions: order_actions_by_priority(citizen),
+    }
+    [business_results, citizen_results]
+  end
 
-    action_groups.map do |key, action_group|
-      {
-        heading: I18n.t("brexit_checker.results.audiences.#{key}.heading"),
-        actions: action_group.sort_by.with_index do |action, index|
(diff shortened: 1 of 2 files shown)


app/helpers/brexit_checker_helper.rb
```diff
@@ -8,14 +8,16 @@ def format_criteria_list(criteria)
   def format_action_audiences(actions)
     action_groups = actions.group_by(&:audience)
 
-    action_groups.map do |key, action_group|
+    action_groups = action_groups.map do |key, action_group|
```
benthorner: It's hard to tell from reading this why we're reversing the groups. Can we do a `sort_by` here, so the ordering predicate is clear?

app/helpers/brexit_checker_helper.rb
```diff
@@ -8,14 +8,16 @@ def format_criteria_list(criteria)
...
       {
         heading: I18n.t("brexit_checker.results.audiences.#{key}.heading"),
         actions: action_group.sort_by.with_index do |action, index|
           [-action.priority, index]
         end,
       }
     end
+
+    action_groups.reverse
```
koetsier: Maybe to be certain about the order of the keys, you could use ```sort_by```? https://apidock.com/ruby/Enumerable/sort_by

app/helpers/brexit_checker_helper.rb
koetsier: could you call this something else so that we know action_groups has changed from hash to array

huwd: Went through with Jos and simplified massively. @benthorner hopefully the result is much easier to grok?

app/helpers/brexit_checker_helper.rb
huwd: Thanks for running through jos, partition is a much better idea!

PR #1718, opened 2019-11-05
Fix Audience headers rendering without content

huwd:
I introduced a bug in #1714
The view layer was only filtering by any action, it is now more nuanced and filters on a per-group basis, that fits the new data structure.

Screen Shot 2019-11-05 at 3 07 53 PM

Also wrote some tests.

### DIFF
--- a/spec/features/brexit_checker/question_results_spec.rb
+++ b/spec/features/brexit_checker/question_results_spec.rb
@@ -3,22 +3,25 @@ require "spec_helper"
 RSpec.feature "Brexit Checker workflow", type: :feature do
   include BrexitCheckerHelper
 
-  scenario "business questions" do
+  scenario "business and citizen questions" do
     when_i_visit_the_brexit_checker_flow
     and_i_answer_citizen_questions
     and_i_answer_business_questions
-    then_i_should_see_the_results_page
-    and_i_should_see_a_pet_action
-    and_i_should_see_a_tourism_action
+    then_i_see_citizen_and_business_results
   end
 
-  scenario "citizen questions" do
+  scenario "business questions only" do
+    when_i_visit_the_brexit_checker_flow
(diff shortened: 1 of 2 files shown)


spec/features/brexit_checker/question_results_spec.rb
```diff
@@ -3,20 +3,34 @@
 RSpec.feature "Brexit Checker workflow", type: :feature do
   include BrexitCheckerHelper
 
-  scenario "business questions" do
+  scenario "business and citizen questions" do
     when_i_visit_the_brexit_checker_flow
     and_i_answer_citizen_questions
     and_i_answer_business_questions
     then_i_should_see_the_results_page
+    and_i_should_see_the_citizens_action_header
```
benthorner: To avoid adding more methods in here, could we look at combining them into a single 'then' step where they 'see_citizen_and_business_results', etc?

spec/features/brexit_checker/question_results_spec.rb
```diff
@@ -27,10 +41,33 @@
     then_i_should_see_the_no_results_page
   end
 
+  def and_i_should_see_the_citizens_action_header
+    expect(page).to have_content I18n.t!("brexit_checker.results.audiences.citizen.heading")
```
vanitabarrett: I would normally check that the text is contained within a specific element, either by class or (for example) a h2. Otherwise it could be somewhere else on the page and your test would still pass.

spec/features/brexit_checker/question_results_spec.rb
huwd: How do these rspec feature tests work, are they just looking for each line to be `true`?

So you'd use `!see_citizen_and_business_results` and `see_citizen_and_business_results` in the main method?

PR #1737, opened 2019-11-14
Enable AB testing for Levenshtein algorithm for suggestions

Tetrino:
This PR allows us to use the Levenshtein algorithm for search
suggestions over the default Elasticsearch algorithm, which relies on
an "optimised" system that may not be too accurate for our purposes.

Trello card: https://trello.com/c/c2YUDVg3/1117-try-different-ways-of-doing-spelling-suggestions


Tetrino:
> Would one need to be removed before the other could be enabled?

Spoke to Andrew about this, as Shingles is turned off now we should be good to go, but for safety I'll remove the Shingles line before I merge.

PR #1742, opened 2019-11-14
Update email signup component

emmabeynon:
For https://trello.com/c/U5eKec0W/241-change-signup-to-get-email-alerts-component

This updates the email signup component as follows:
- it accepts a title
- the link text is no longer bold
- the bell icon is no longer a link

We also update the email signup link text.

**Before**
before

**After**
after


app/views/components/_email_link.html.erb
```diff
@@ -3,10 +3,15 @@
   data_attributes ||= {}
 %>
 <div class="app-c-email-link">
+  <% if link_title %>
+    <div class="app-c-email-link__title">
```
vanitabarrett: This could be a `p` tag rather than a `div`

huwd:
@vanitabarrett How does the link tracking work here, was there anything else we needed to add?

vanitabarrett:
@huwd I think the tracking can stay the same : https://github.com/alphagov/finder-frontend/pull/1742/files#diff-6e38c2fb80d2131332b6e498435430a9R68

app/views/components/_email_link.html.erb
emmabeynon: Done 😸

PR #1755, opened 2019-11-20
[DO NOT MERGE] Stop sending ecommerce tracking content id

vanitabarrett:
Trello Card: https://trello.com/c/2iCWKnNV/274-adjust-existing-ecommerce-tracking-logic-to-handle-passing-page-path-as-name

## What
Follows on from this change: https://github.com/alphagov/static/pull/1941

We no longer send `id` with our ecommerce tracking on finders. Instead, the page path is sent as `name` (see linked static PR).

One exception are spelling suggestions: we continue to send `id` for these.


vanitabarrett:
@sihugh @kr8n3r inviting you to review as this will affect tracking on finders

PR #1757, opened 2019-11-20
Enable non-modifiable email facet choices

laurentqro:
For specialist finders using `email_filter_facets`, the default behaviour is currently to allow the user to modify email filters before subscribing to the alerts.

This PR introduces a flag in the schema that allows us to skip this intermediate step and go straight to sign up when that makes sense, e.g. when the list of facets is too long to display as checkboxes.

I am not a fan of the negatively phrased `cannot_modify_choices`; the intention here is to allow existing finders to keep their default behaviour without changing anything to their schemas. Naming suggestions / alternate solutions welcome. :)

**Update 26/11/2019**

I have replaced the `cannot_modify_choices` boolean flag with the ability to add `"all_selected_facets"` value to `email_filter_by`. This is a tactical solution too, but a better one, in that it avoids the negative flag and doesn't require an intrusive change to the email sign up content item schema.

--

### Card

https://trello.com/c/rTCRCbiM/930-fix-the-email-subscriptions-for-residential-property-tribunal-decisions-et-al-%F0%9F%8D%90

--

- https://finder-frontend-pr-1757.herokuapp.com/search/all
- https://finder-frontend-pr-1757.herokuapp.com/search/research-and-statistics
- https://finder-frontend-pr-1757.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-1757.herokuapp.com/get-ready-brexit-check/questions
- https://finder-frontend-pr-1757.herokuapp.com/drug-device-alerts
- https://finder-frontend-pr-1757.herokuapp.com/find-eu-exit-guidance-business
- https://finder-frontend-pr-1757.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- https://finder-frontend-pr-1757.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- https://finder-frontend-pr-1757.herokuapp.com/uk-nationals-living-eu
- https://finder-frontend-pr-1757.herokuapp.com/prepare-business-uk-leaving-eu

### DIFF
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -35,7 +35,7 @@ class SignupPresenter
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any? && !(content_item["details"]["email_filter_by"] == "all_selected_facets")
+    choices? && choices_formatted.any? && content_item["details"]["email_filter_by"] != "all_selected_facets"
   end
 
   def hidden_choices
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -35,7 +35,7 @@ class SignupPresenter
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any?
+    choices? && choices_formatted.any? && !content_item["details"]["cannot_modify_choices"]
   end
 
   def hidden_choices
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -35,7 +35,7 @@ class SignupPresenter
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any? && !content_item["details"]["cannot_modify_choices"]
+    choices? && choices_formatted.any? && !(content_item["details"]["email_filter_by"] == "all_selected_facets")
   end
 
   def hidden_choices
--- a/app/presenters/subscriber_list_params_presenter.rb
+++ b/app/presenters/subscriber_list_params_presenter.rb
@@ -28,11 +28,21 @@ private
   end
 
   def facets
-    allowed_facets.reject { |facet| facet.key?("facet_choices") }
+    return allowed_facets if can_modify_choices?
+
+    email_filter_facets
+  end
+
+  def can_modify_choices?
+    !content_item["details"]["cannot_modify_choices"]
+  end
+
+  def email_filter_facets
+    content_item["details"].fetch("email_filter_facets", [])
   end
 
   def allowed_facets
-    content_item["details"].fetch("email_filter_facets", [])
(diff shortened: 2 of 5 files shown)


app/presenters/signup_presenter.rb
```diff
@@ -35,7 +35,7 @@ def email_filter_by
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any?
+    choices? && choices_formatted.any? && !content_item["details"]["cannot_modify_choices"]
```
barrucadu: There are other finders which don't present a list of facets to change, eg https://www.gov.uk/search/news-and-communications/email-signup?level_one_taxon=e491505c-77ae-45b2-84be-8c94b94f6a2b&organisations%5B%5D=academy-for-social-justice&order=updated-newest, how do they do it?  Can we reuse the same mechanism?  Is that impossible because specialist finders do email in a different way?

I'm really not a fan of a negatively-named boolean, but I see your point about not wanting to change all the existing content items.  If this is necessary for the specialist finders, then I think it should be documented in the card about making email signups consistent as something to remove when that work is done.

Maybe it's even worth having a card about making specialist and non-specialist finder email signups consistent, and leaving the wider email signup consistency problems for after this smaller problem is solved.

laurentqro: @barrucadu Maxwell did indeed mention that some followup work will have to be done to make the email alert signups consistent.

As far as I can see, the divide isn't so much between finders and specialist finders, but between finders that use so-called dynamic filters (like News and Communications), and the rest: https://github.com/alphagov/finder-frontend/blob/master/lib/email_alert_title_builder.rb#L123.

app/presenters/signup_presenter.rb
```diff
@@ -35,7 +35,7 @@ def email_filter_by
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any?
+    choices? && choices_formatted.any? && !(content_item["details"]["email_filter_by"] == "all_selected_facets")
```
sihugh: Is this so we wouldn't need to have `cannot_modify_choices` in the content item?

(The build's failing on linting at the mo)

app/presenters/signup_presenter.rb
sihugh: I don't mind doing something tactical here.  We're flagging that there's a general lack of consistency (and that the signup journey needs work) across email journeys.  So long as we can keep a note of what we've done in commit messages and some product docs, that's probs OK.

sihugh: (Linting's failing by the way)

laurentqro: Thanks @sihugh, I'll document that.

PR #1758, opened 2019-11-20
Replace deprecated govuk-lint with rubocop-govuk

laurentqro:
Replace deprecated govuk-lint with rubocop-govuk https://github.com/alphagov/rubocop-govuk.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ inherit_gem:
 AllCops:
   Exclude:
     - 'lib/tasks/cucumber.rake' # Automatically generated
+    - 'tmp/**/*'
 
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -11,3 +11,7 @@ Metrics/BlockLength:
     - 'features/**/*.rb'
     - 'spec/**/*'
     - 'config/routes.rb'
+
+inherit_gem:
```
benthorner: I wonder if all the additional linting errors would disappear if we moved this to the top of the file?

laurentqro: @benthorner it works now that I've excluded temp files.

PR #1766, opened 2019-11-27
Next to continue

huwd:
Resurrected version of: #1763

---

Relates to: Trello (https://trello.com/c/Zut0LKKK/262-do-not-deploy-change-button-label-in-all-questions-of-the-checker-to-continue)

--- 

## What

Changes button copy and hints from "Next" -> "Continue", on the Brexit checker only.
Non standard "Next" buttons remain in finder-frontend questions/answers.

## Why
Because the design system says `Make sure your ‘Continue’ button is labelled ‘Continue’, not ‘Next’`

https://design-system.service.gov.uk/patterns/question-pages/

We should be using common components whenever possible to be consistent across GOV.UK.

### DIFF
--- a/app/views/brexit_checker/show.html.erb
+++ b/app/views/brexit_checker/show.html.erb
@@ -73,7 +73,7 @@
         %>
         <div class="govuk-form-footer">
           <%= render "govuk_publishing_components/components/button", {
-            text: "Next"
+            text: "Continue"
           } %>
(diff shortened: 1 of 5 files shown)


huwd:
Gonna merge into our holding branch as was approved here: https://github.com/alphagov/finder-frontend/pull/1765

PR #1769, opened 2019-11-28
Refactor sub groups results

huwd:
Refactors to improve the grouping PR


benthorner:
Would be great to have an overall description of the change being made. Since this is also a visual change, would also be good to have a screenshot or two.

PR #1731, opened 2019-11-08
[DO NOT MERGE] Post Election Changes

huwd:
Trello: https://trello.com/c/uLLaEVPV/186-dont-deploy-update-checker-results-page

## PEP restricted PR
🚨 Do not merge until product has approved. 🚨

## Included Changes:
- Add Subgroups to results page (this PR + #1754)
- Changed brexit checker buttons from next to continue (#1766)

PR #1771, opened 2019-11-29
Moves logic from checker results view to helper

huwd:
A part of: https://trello.com/c/uLLaEVPV/186-dont-deploy-update-checker-results-page

---

## What

In advance of larger changes to the view, to help users parse large numbers of actions by grouping them, start by simplifying the view by moving logic to the helper. This gives a cleaner canvas to begin working on the subgrouping feature.

## Why
Attempts to [deliver all of the subgrouping](https://github.com/alphagov/finder-frontend/pull/1731) at once led to too much complexity at once. This is the first in a sequence of PRs that breaks out useful work that can be delivered in more manageable chunks.

This particular set of changes assumes that the sub-groupings will inevitably add complexity to the results page, and pre-emptively reduces complexity in the view layer by moving logic into a helper.

## When can it go live?
As soon as merged, this is just a refactor of existing code in live. No new copy, features or user facing changes. Can be merged into master before the rest of the grouping work begins.

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -70,4 +70,32 @@ module BrexitCheckerHelper
 
     url.to_s
   end
+
+  def brexit_results_email_link_label(actions)
+    if actions.any?
+      t("brexit_checker.results.email_sign_up_link")
+    else
+      t("brexit_checker.results.email_sign_up_link_no_actions")
+    end
+  end
+
+  def brexit_results_title(actions, criteria_keys)
+    if actions.any?
+      t("brexit_checker.results.title")
+    elsif criteria_keys.any?
+      t("brexit_checker.results.title_no_actions")
+    else
+      t("brexit_checker.results.title_no_answers")
+    end
+  end
+
+  def brexit_results_description(actions, criteria_keys)
+    if actions.any?
+      t("brexit_checker.results.description")
+    elsif criteria_keys.present?
+      t("brexit_checker.results.description_no_actions")
+    else
+      t("brexit_checker.results.description_no_answers")
+    end
+  end
 end
--- a/app/views/brexit_checker/results.html.erb
+++ b/app/views/brexit_checker/results.html.erb
@@ -1,19 +1,7 @@
 <%
-  action_based_email_link_label = @actions.any? ? t('brexit_checker.results.email_sign_up_link') : t('brexit_checker.results.email_sign_up_link_no_actions')
-  action_based_title            = if @actions.any?
-                                    t('brexit_checker.results.title')
-                                  elsif criteria_keys.present?
-                                    t('brexit_checker.results.title_no_actions')
-                                  else
-                                    t('brexit_checker.results.title_no_answers')
-                                  end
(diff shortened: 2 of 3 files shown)


app/views/brexit_checker/results.html.erb
```diff
@@ -1,19 +1,7 @@
...
-                                  end
-  action_based_description      = if @actions.any?
-                                    t('brexit_checker.results.description')
-                                  elsif criteria_keys.present?
-                                    t('brexit_checker.results.description_no_actions')
-                                  else
-                                    t('brexit_checker.results.description_no_answers')
-                                  end
+  action_based_email_link_label = email_link_label(@actions)
+  action_based_title            = title(@actions, criteria_keys)
```
benthorner: Minor, but I find this kind of spacing irritating when I need to make a change, because it tends to cause a lot of extra typing. We don't tend to do this anywhere else (probably for that reason).

spec/helpers/brexit_checker_helper_spec.rb
```diff
@@ -199,4 +199,58 @@
...
+    it "Should return the no results email link copy if there are no actions" do
+      expect(email_link_label([])).to eq(t("brexit_checker.results.email_sign_up_link_no_actions"))
+    end
+  end
+
+  describe "#title" do
+    let(:actions) { [FactoryBot.build(:brexit_checker_action)] }
+    let(:criteria_keys) { %w"nationality-eu" }
+
+    it "Should return the title if there are actions and answers" do
```
benthorner: Minor: the word "should" implies some level of doubt 😆. Could we just say `it "returns..."`.

app/helpers/brexit_checker_helper.rb
```diff
@@ -70,4 +70,32 @@ def notification_email_link(non_tracked_url, notification)
 
     url.to_s
   end
+
+  def email_link_label(actions)
```
benthorner: Since all helper methods are global, we should scope these generic methods to avoid clashes.

app/views/brexit_checker/results.html.erb
huwd: Yeah wasn't sure about that, largely just went with the flow as it was there.
If that's not something we do elsewhere then i don't want it either.

Removed and pushed amended change.

spec/helpers/brexit_checker_helper_spec.rb
huwd: fair cop!
Amended

app/helpers/brexit_checker_helper.rb
huwd: ah great point, I'd guessed they were more specifically scoped.

Humm `brexit_results_title` and so on?

app/views/brexit_checker/results.html.erb
benthorner: Ah yeah. Sorry, I didn't check the "before" 😬

app/helpers/brexit_checker_helper.rb
huwd: Cool, amended that too

PR #1774, opened 2019-11-29
Add grouping_criteria to csv importer

huwd:
A part of: https://trello.com/c/uLLaEVPV/186-dont-deploy-update-checker-results-page

---

## What

grouping criteria are a new column in the [brexit checker spreadsheet](https://docs.google.com/spreadsheets/d/1wIeBTitJVfkWa7oKrGmusIo2r4TsvXVdlne_xG6YjYs/edit), that will allow some actions to be added to subgroupings.

## Why

This sets some of the background for adding grouping to the results page, by allowing data to be added from the spreadsheets to actions.

## When can it go live?
It's a backend data processing task with no copy or user facing change so could be merged straight away.

### DIFF
--- a/lib/brexit_checker/action.rb
+++ b/lib/brexit_checker/action.rb
@@ -12,7 +12,8 @@ class BrexitChecker::Action
 
   attr_reader :id, :title, :consequence, :exception, :title_url, :title_path,
               :lead_time, :criteria, :audience, :guidance_link_text,
-              :guidance_url, :guidance_path, :guidance_prompt, :priority
+              :guidance_url, :guidance_path, :guidance_prompt, :priority,
+              :grouping_criteria
 
   def initialize(attrs)
     attrs.each { |key, value| instance_variable_set("@#{key}", value) }
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -21,6 +21,8 @@ actions:
       - living-uk
       - join-family-uk-yes
   audience: citizen
+  grouping_criteria:
+  - living-uk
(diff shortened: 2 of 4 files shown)


benthorner:
This is a nice, simple change 👍. I'm not sure we should merge it, though. Although it's separable, it's also functionally unused, which makes it liable to change when the code that uses it comes in. I think it would make more sense to see this as a commit in a larger PR.

huwd:
Humm ok, I'd wanted to go on to deal with some other backend changes.
There's a groups.rb and a few other bits that i thought i could add silently to the backend. But they might start interacting with this code...

So would you create a new feature branch and merge in there? Or branch any further changes off this one?

Trying to make sure this doesn't become merge spaghetti again!

benthorner:
I'd suggest doing this as part of the overall change, as it'll make more sense to the person reviewing it. My main question in reviewing this is: "what's it for?". If this were in a PR to implement the associated feature, then I'd be able to answer the question for myself.

benthorner:
Is there a way we could move to the new feature without this? Perhaps with some default?

huwd:
> Is there a way we could move to the new feature without this? Perhaps with some default?

I think this would be a pre-requisite for the new feature, I'd need the data in the model in order to render anything.

By a default do you mean building the feature around something like 'grouping_criteria' = nil?
I could do that, but it might be more complicated by the fact only business results can have no grouping_criteira. Ideally I want to validate them for any citizens action

huwd:
I guess if not I'll just close this PR and then open a larger one with this as a commit for perhaps... all the non-breaking backend changes?

PR #1786, opened 2019-12-06
Fix validation of criteria separating extraction/validation

huwd:
We import criteria from strings in a spreadsheet. As such they're untrusted data, that we should validate. However criteria that is embedded in a hash can vary between having symbolic keys and string based keys.

This has led to a bug where a non-existing criteria `export-from-eu` snuck in to two business results.

This PR both fixes that, and separates criteria extraction from validation so it can be properly unit tested. This also lays some useful groundwork for us to re-use this criteria extraction on the new citizens grouping feature.

### DIFF
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -1568,7 +1568,7 @@ actions:
   - any_of:
     - environment
     - import-from-eu
-    - export-from-eu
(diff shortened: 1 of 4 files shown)


benthorner:
@huwd I think this is superseded by https://github.com/alphagov/finder-frontend/pull/1787

PR #1790, opened 2019-12-09
[On hold until Post Election] - Update email signup component

huwd:
Cherry picked version of this PR: https://github.com/alphagov/finder-frontend/pull/1742

---

For https://trello.com/c/U5eKec0W/241-change-signup-to-get-email-alerts-component

---

This updates the email signup component as follows:
- it accepts a title
- the link text is no longer bold
- the bell icon is no longer a link

We also update the email signup link text.

**Before**
before

**After**
after


vanitabarrett:
I think the `Update .app-c-email-link__icon CSS` commit can be squashed into `Stop linking to email link icon`

app/assets/stylesheets/components/_email-link.scss
```diff
@@ -1,16 +1,19 @@
...
   background: govuk-colour('grey-4');
 
+  .app-c-email-link__title {
+    font-weight: bold;
+  }
+
+  .app-c-email-link__icon {
+    position: relative;
+    margin-right: govuk-spacing(2);
+    top: govuk-spacing(1);
```
vanitabarrett: I would set this to an explicit `5px`. Only because some larger `govuk-spacing` numbers have different values on mobile vs desktop which is not what we want in this case. It doesn't affect `govuk-spacing(1)` but it makes it easier for someone to simply change it to `govuk-spacing(6)`, for example, without considering that mobile vs desktop difference.

vanitabarrett:
Look good, I only have a few commit tidy-up comments:

- I think the `Update .app-c-email-link__icon CSS` commit can be squashed into `Stop linking to email link icon`
- `Add css for new results page group styling` - not sure this commit belongs in this branch?

app/assets/stylesheets/components/_email-link.scss
huwd: Great shout, will amend now!

huwd: done

huwd:
Done, thanks very much @vanitabarrett

app/assets/stylesheets/components/_email-link.scss
vanitabarrett: @huwd Looks like this is still `govuk-spacing(1)`?

huwd: Ah, looks like I got the wrong end of the stick and changed the margin-right above. Have reverted that and edited like `13`

PR #1798, opened 2019-12-11
change brexit checkbox to small variant

SamJamCul:
Change based on this https://trello.com/c/PiJGPil8/1179-tweak-the-sidebar-design-of-the-news-comms-type-page-template


SamJamCul:
> I think you want the option `small: true`, see [this page of the component guide](https://components.publishing.service.gov.uk/component-guide/checkboxes/with_small_checkboxes).
> 
> Also I'm not sure the class of `govuk-checkboxes` is needed on the wrapping div, can you try removing that?

Hmm, I made the change you suggested, but it didn't work! I've been working from the info in [this link](https://design-system.service.gov.uk/components/checkboxes/), and the only way I've gotten it working is by adding that modifier to the class. Any ideas?

andysellick:
Seems to be a problem with the component where the small option doesn't work if there's only one checkbox. This has now [been fixed](https://github.com/alphagov/govuk_publishing_components/pull/1221) and a new version of the gem should be on its way.

Just to note that the design system documentation is only sort-of applicable to the components, it's best to use the component guide and not the design system documentation.

PR #1808, opened 2019-12-17
Aligns filter elements into a neat column

SamJamCul:
Change from this card https://trello.com/c/UiD3ey0z/1181-tweak-the-sidebar-design-of-the-research-and-statistics-type-page-template

Screenshot below shows the aligning of all the elements - I haven't been able to use govuk-spacing because its really specific pixel numbers. 

image

### DIFF
--- a/app/assets/stylesheets/components/_expander.scss
+++ b/app/assets/stylesheets/components/_expander.scss
@@ -19,9 +19,9 @@
 .js-enabled {
   .app-c-expander__heading {
     position: relative;
-    padding: 10px 8px 5px 40px;
+    padding: 10px 8px 5px 43px;
   }
-  
+
   .app-c-expander__content {
     display: none;
   }
@@ -35,7 +35,7 @@
   display: none;
   position: absolute;
   top: 0;
-  left: 6px;
+  left: 9px;
   width: 30px;
   height: 40px;
   fill: govuk-colour("black");
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -385,3 +385,8 @@ mark {
   margin-bottom: govuk-spacing(7);
   padding-bottom: govuk-spacing(0);
 }
+
+.facet-container {
+  padding-left: 13px;
+  padding-right: 13px;
+}
--- a/app/assets/stylesheets/components/_option-select.scss
+++ b/app/assets/stylesheets/components/_option-select.scss
@@ -87,7 +87,7 @@
   display: none;
   position: absolute;
   top: 0;
-  left: 6px;
+  left: 9px;
   width: 30px;
   height: 40px;
   fill: govuk-colour("black");
@@ -128,7 +128,7 @@
 .js-enabled {
   .app-c-option-select__heading {
     position: relative;
-    padding: 10px 8px 5px 40px;
+    padding: 10px 8px 5px 43px;
   }
 
   .app-c-option-select__icon--up {
--- a/app/views/finders/_radio_facet.html.erb
+++ b/app/views/finders/_radio_facet.html.erb
@@ -1,5 +1,7 @@
-<%= render "govuk_publishing_components/components/radio", {
-  name: radio_facet.key,
-  small: true,
-  items: radio_facet.options
-} %>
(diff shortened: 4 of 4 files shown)


app/assets/stylesheets/components/_expander.scss
```diff
@@ -87,3 +88,7 @@
     display: block;
   }
 }
+
+.app-c-option-select__icon {
```
kr8n3r: option-select styles should go in the options-select partial. this file is the expander partial

half pixels should be avoided as some browser round them up, some round them down

app/assets/stylesheets/components/_expander.scss
```diff
@@ -39,6 +39,7 @@
   width: 30px;
   height: 40px;
   fill: govuk-colour("black");
+  padding-left: 3.5px;
```
kr8n3r: we can adjust the `left` property and then we don't seen the padding declaration. 
see the comment below about half pixels

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -385,3 +385,7 @@ mark {
   margin-bottom: govuk-spacing(7);
   padding-bottom: govuk-spacing(0);
 }
+
+.govuk-radios {
```
kr8n3r: the padding is specific to the small radios, but the styling is applied to the generic radios class which will affect the large (normal) radios as well - producing unwanted results, suggest we scope the adjustment to `govuk-radios--small`

app/assets/stylesheets/components/_expander.scss
SamJamCul: Moved this into the option-select partial, and changed it to 3px

app/assets/stylesheets/finder_frontend.scss
SamJamCul: changed it so that it's only radio smalls

app/assets/stylesheets/components/_expander.scss
kr8n3r: > we can adjust the left property and then we don't seen the padding declaration.

removing `padding-left: 3px` and changing `left` from 6 to 9px should produce the same visual result

app/assets/stylesheets/finder_frontend.scss
SamJamCul: I've changed the others so that the movement is all in the `left` attribute - but for radio-smalls I was only able to get it to work with the `padding-left`. Hope that's alright!

app/assets/stylesheets/components/_expander.scss
SamJamCul: Sorry, I pushed the wrong commit when I resolved these, it should be better now

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -385,3 +385,7 @@ mark {
   margin-bottom: govuk-spacing(7);
   padding-bottom: govuk-spacing(0);
 }
+
+.govuk-radios--small {
```
andysellick: Making changes like this that affect components inside apps are against our component principles - if a component needs to appear differently, that change needs to be made on the component itself, either as the default behaviour or as an option.

andysellick: Related to this, why is everything indented by 12px, instead of making it flush with the left edge of the column? (which would align everything with the left edge of the search box and make this change unnecessary)

kr8n3r: the design change here doesn't need to be on the component itself, so no need for the option. If components allowed passing through custom classes, that would be useful.
We're going to wrap these radios in a `facet-container` and apply the padding there

andysellick: That's a better option, thanks.

Why is everything indented though?

kr8n3r: because of the hover effect on radios and checkboxes. padding on option-select and expander that was there before hasn't changed

SamJamCul: It's done, we've now got the radio buttons inside a container that's being used to position them correctly, rather than modifying the component itself.

PR #1819, opened 2019-12-20
Add share links to brexit checker results

huwd:
Trello card: https://trello.com/c/fB9KKRrJ/242-add-share-your-results-component

---

Adds social share links that pass on your results URL onto various social media platforms.

## Desktop
Screenshot_2019-12-20 Brexit check what you need to do if there is no deal(2)

## Intermediate
Screenshot_2019-12-20 Brexit check what you need to do if there is no deal

## Tiny Mobile
Screenshot_2019-12-20 Brexit check what you need to do if there is no deal(1)

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -1,6 +1,10 @@
 require "addressable/uri"
 
 module BrexitCheckerHelper
+  def encoded_results_url
+    CGI.escape(request.original_url)
+  end
+
   def format_criterion(criteria_key)
     BrexitChecker::Criterion.load_by(criteria_key).first.text
   end
--- a/app/views/brexit_checker/_share_links.html.erb
+++ b/app/views/brexit_checker/_share_links.html.erb
@@ -1,0 +1,38 @@
+<section class="brexit-checker__share">
+  <% hidden_text = capture do %>
+    <span class="govuk-visually-hidden">Share on</span>
+  <% end %>
+  <h4 class="govuk-heading-m">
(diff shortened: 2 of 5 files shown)


app/views/brexit_checker/_share_links.html.erb
```diff
@@ -0,0 +1,39 @@
+
```
vanitabarrett: Extra new line

huwd: atch! i _really_ need to get proper linting set up in ruby.

Thanks for spotting.

app/helpers/brexit_checker_helper.rb
```diff
@@ -1,6 +1,10 @@
 require "addressable/uri"
 
 module BrexitCheckerHelper
+  def encoded_results_url
```
huwd: no test for this, but it's a fairly simple alias.
I think I'd need to stub the request to unit test that here?

huwd:
🤔 can't work out what's going on with the e2e tests here, everything runs fine on local... and I've re-run em.

huwd:
seems to be: 
```
13:46:00  Failed examples:
13:46:00  
13:46:00  rspec ./spec/publisher/publishing_content_to_government_frontend_spec.rb:7 # Publishing content from Publisher to Government Frontend Publishing a Help guide
13:46:00  
13:46:00  Finished in 30.76 seconds (files took 1.93 seconds to load)
13:46:00  1 example, 0 failures
```

Unrelated to this change?

huwd:
oh weird, that change went away...

PR #1820, opened 2019-12-20
Change sizing of brexit checker results notification window

huwd:
Cross referenced with the prototype here https://govuk-dynamic-list.herokuapp.com/get-ready-for-brexit/results/uk-national-and-business-mvp

The notification promps were larger and more prominent.

### Before
Screenshot_2019-12-20 Brexit check what you need to do if there is no deal(3)

### After 
Screenshot_2019-12-20 Brexit check what you need to do if there is no deal(4)

### After (mobile)
Screenshot_2019-12-20 Brexit check what you need to do if there is no deal(5)

### DIFF
--- a/app/assets/stylesheets/components/_email-link.scss
+++ b/app/assets/stylesheets/components/_email-link.scss
@@ -3,6 +3,10 @@
   padding: govuk-spacing(4);
   background: govuk-colour('grey-4');
 
+  .govuk-heading-m {
+    margin-bottom: govuk-spacing(2);
+  }
+
   .app-c-email-link__title {
     font-weight: bold;
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_email-link.scss
```diff
@@ -14,6 +18,7 @@
   }
 
   .app-c-email-link__link {
+    @include govuk-font(19);
```
benthorner: Will this affect other email links in the app?

huwd: it would yes, but this class is only actually used in one place in the whole of finder-frontend ([here](https://github.com/alphagov/finder-frontend/blob/f2f5a582d113868f0e3278b24825e800a53f37ab/app/views/components/_email_link.html.erb#L12)) 

Which is the target of our change.

benthorner: Indeed. My question was where this component then gets used. Looking into it, I can see it's only used by the checker - it's a generic component, so could be used anywhere.

PR #1825, opened 2019-12-20
[WIP] Browse search prototype (OscarSearch)

bilbof:
This is an experimental browse page. The intent of this change is to provide two new features to users:

- a single predicted document as an answer to a query
- a set of search results grouped by topics that we think the user is interested in

This work was done by @oscarwyatt in alphagov/collections#858 and has been moved here so we can make some performance improvements using the features we have added to finder-frontend for that purpose.

This PR includes some ideas we'd like to test:

- apply filters that we predict the user would find useful (e.g filter by services if the user uses a verb)
- highlight a result when we have high confidence in it being the right answer
- display results grouped by topic and subtopic

Still to do ahead of user testing:

- make speed optimisations to the prototype
- refactor, add tests

View the work in progress browse page at https://finder-front-search-bro-pz5kjw.herokuapp.com/browse?q=universal+credit

https://trello.com/c/RN2A2b8w/1229-next-steps-for-testing-oscarsearch


kr8n3r:
browse page not found?

bilbof:
This commit needs some refactoring and better test coverage before it can merged. We can use this branch as a review app for user testing in the meantime.

This branch should respond to requests a bit faster than the initial prototype. It does some extra caching / avoiding requests, and batches search requests, but otherwise is unchanged.

sihugh:
Nice work @bilbof (and @oscarwyatt) 👍 .  I wonder if we need to do some polishing on the mobile view before testing (or whether we're OK just looking at desktop for now)?  I might have a chat to Jeremy about who we think is most likely to use this

PR #1827, opened 2019-12-23
Results accessibility audit

huwd:
Trello: [A part of](https://trello.com/c/Kmp4WW27/339-accessibility-audit-of-checker-results-page)

Small accessibility tweaks, including:
- Adding aria labels to repeating 'aside' elements that hold criteria, uses group heading or business string as a label. Uses a new helper plus tests.
- Removes role from main elements. Those are now redundant with proper adoption of HTML5 in all major browsers
- Promotes the heading of the share buttons from h4 to h2, as it's really a higher heading level.

### DIFF
--- a/app/helpers/brexit_checker_helper.rb
+++ b/app/helpers/brexit_checker_helper.rb
@@ -5,6 +5,12 @@ module BrexitCheckerHelper
     CGI.escape(request.original_url)
   end
 
+  def criteria_aside_label(heading = nil)
+    return "you-and-your-family-actions-group-#{heading.downcase.gsub(' ', '-')}-criteria" if !heading.nil?
+
+    "business-actions-criteria"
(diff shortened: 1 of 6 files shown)


app/helpers/brexit_checker_helper.rb
```diff
@@ -5,6 +5,12 @@ def encoded_results_url
     CGI.escape(request.original_url)
   end
 
+  def criteria_aside_label(heading)
+    return "you-and-your-family-actions-group-#{heading.downcase.gsub(" ", "-")}-criteria" if !heading.nil?
+
+    "business-actions-critiera"
```
koetsier: criteria

huwd: fixed!

PR #1831, opened 2019-12-27
Refactor email signup presenter

bilbof:
This change refactors finder-frontend to reflect changes to specialist publisher (https://github.com/alphagov/specialist-publisher/pull/1555) and content schemas (https://github.com/alphagov/govuk-content-schemas/pull/948).

Those changes make the email signup content items published by Specialist Publisher consistent with those published by Search API.

Making the content items consistent will help us make the signup flow for email subscriptions consistent. It'll also make working with the email signup flow in finder-frontend a bit easier.

Users shouldn't see any differences as a result of this change.

This should be merged and deployed after https://github.com/alphagov/specialist-publisher/pull/1555 is deployed and the content-items republished.

https://trello.com/c/otcTG22f/1246-email-alert-signup-tech-debt

More details about the changes in specialist publisher:

Specialist Publisher used to use other keys for email signup page facets. E.g. it put facets into `details.email_signup_choice` while search api would use `email_filter_facets`. The only reason for this is that we didn't get round to making them consistent.

Two keys are being removed from the finder email signup schema (https://github.com/alphagov/govuk-content-schemas/pull/948): `email_signup_choice` and `email_filter_name`. More could be deleted, like `email_filter_by`, but we can do it later.

### DIFF
--- a/app/presenters/signup_presenter.rb
+++ b/app/presenters/signup_presenter.rb
@@ -25,11 +25,11 @@ class SignupPresenter
   end
 
   def email_filter_by
-    content_item["details"].fetch("email_filter_by", nil)
+    content_item.dig("details", "email_filter_by")
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any? && content_item["details"]["email_filter_by"] != "all_selected_facets"
+    choices? && choices_formatted.any? && email_filter_by != "all_selected_facets"
   end
 
   def hidden_choices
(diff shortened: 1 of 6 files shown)


laurentqro:
Very nice work!

From the top of my head, `email_filter_by` is used by residential property tribunal decisions, and export health certificates. with a value of `all_selected_filters`, in order to override the default behaviour.

It is indeed something that should be removed entirely once all finders behave the same way.

app/presenters/signup_presenter.rb
```diff
@@ -25,11 +25,11 @@ def beta?
   end
 
   def email_filter_by
-    content_item["details"].fetch("email_filter_by", nil)
+    content_item.dig("details", "email_filter_by")
```
laurentqro: 💯

features/fixtures/cma_cases_signup_content_item.json
```diff
@@ -91,61 +91,63 @@
   "description": "You'll get an email each time a case is updated or a new case is published.",
   "details": {
     "beta": false,
-    "email_signup_choice": [
+    "email_filter_facets": [
```
laurentqro: 🥇

app/presenters/signup_presenter.rb
```diff
@@ -25,11 +25,11 @@ def beta?
...
   end
 
   def can_modify_choices?
-    choices? && choices_formatted.any? && content_item["details"]["email_filter_by"] != "all_selected_facets"
+    choices? && choices_formatted.any? && email_filter_by != "all_selected_facets"
```
laurentqro: 💯

PR #1836, opened 2020-01-03
Allowlist for email alert filters

bilbof:
This implements an allowlist for the filters that we permit users to use in their subscriber lists (on the finder frontend side).

Users shouldn't see a difference as a result of this change.

When signing up for email alerts, users can submit filters they'd like to sign up with:

```
{ "people": ["albus-dumbledore"], "organisations": ["hogwarts"] }
```

This change ensures that the keys and values submitted are correct and will result in a correctly working email subscription.

For now we'll show users a generic error if they've made an invalid request. I've also used a 200 rather than a 400 in the response, because we want to re-render the same signup page, and slimmer renders an error page if we respond with a bad request status. It would be awesome to improve on how we respond to errors in the future!

I'd recommend viewing by commit, since the first commit contains the bulk of the change (updating content item fixtures).

Should be merged after https://github.com/alphagov/search-api/pull/1884 is merged and deployed.

---

- https://finder-frontend-pr-1836.herokuapp.com/search/all
- https://finder-frontend-pr-1836.herokuapp.com/search/research-and-statistics
- https://finder-frontend-pr-1836.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-1836.herokuapp.com/get-ready-brexit-check/questions
- https://finder-frontend-pr-1836.herokuapp.com/drug-device-alerts
- https://finder-frontend-pr-1836.herokuapp.com/find-eu-exit-guidance-business
- https://finder-frontend-pr-1836.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- https://finder-frontend-pr-1836.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- https://finder-frontend-pr-1836.herokuapp.com/uk-nationals-living-eu
- https://finder-frontend-pr-1836.herokuapp.com/prepare-business-uk-leaving-eu

https://trello.com/c/LMsj6yPk/1247
https://trello.com/c/tIR0DvbN/1248-business-finder-email-signup-errors-if-no-facet-options-selected

### DIFF
--- a/app/views/email_alert_subscriptions/new.html.erb
+++ b/app/views/email_alert_subscriptions/new.html.erb
@@ -22,9 +22,9 @@
         items: @signup_presenter.choices_formatted,
         visually_hide_heading: true
       } %>
-      <% if @error_message.present? %>
-        <p class="signup-choices__message"><%= @error_message %></p>
-      <% end %>
+    <% end %>
+    <% if @error_message.present? %>
+      <p class="signup-choices__message"><%= @error_message %></p>
     <% end %>
 
     <% @signup_presenter.hidden_choices.each do |hidden_choice| %>
--- a/lib/parameter_parser/email_alert_parameter_parser.rb
+++ b/lib/parameter_parser/email_alert_parameter_parser.rb
@@ -1,0 +1,98 @@
+# Ensures that email alert filter keys and values provided by users
+# and that Finder Frontend sends to Email Alert API are within the
+# set of permitted keys and values provided in the content item or
+# registries of people, organisations, taxons, and world locations.
+module ParameterParser
+  class EmailAlertParameterParser
+    include ActiveModel::Validations
+
+    validates :applied_filters, email_alert_params: true
+
+    def initialize(content_item, filter_params, params)
+      @content_item = content_item
+      @filter_params = parsed_params(filter_params, params).deep_stringify_keys
+    end
+
+    def applied_filters
+      @applied_filters ||= begin
+        permitted_filter_keys.each_with_object({}) { |key, filter_hash|
+          filter_hash[key] = permitted_values_for_filter_key(key)
+        }.compact
+      end
+    end
+
+  private
+
+    attr_reader :filter_params, :content_item
+
+    def permitted_facets
+      content_item.dig("details", "email_filter_facets") || []
+    end
+
+    def permitted_filter_keys
+      allowed = permitted_facets.map { |facet| key_for_facet(facet) }
+      given = filter_params.keys
+      allowed & given
+    end
+
+    def permitted_values_for_filter_key(key)
+      facet = find_facet_by_key(key)
+      return if facet.blank?
+
+      given_values = Array(filter_params[key])
+      allowed_values = allowed_values_for_facet(facet)
+      permitted = allowed_values & given_values
+      permitted if permitted.any?
+    end
+
+    def find_facet_by_key(key)
+      permitted_facets.find { |facet| key_for_facet(facet) == key }
+    end
+
+    def key_for_facet(facet)
+      facet["filter_key"] || facet["facet_id"]
+    end
+
+    # TODO: this could be improved upon
+    def allowed_values_for_facet(facet)
+      # Facet choices used by /cma-cases/email-signup
+      facet_choices = facet.fetch("facet_choices", [])
+      facet_choice_values = facet_choices
+        .map { |choice| choice["content_id"] || choice["key"] }
+        .uniq
+        .compact
+
+      return facet_choice_values if facet_choice_values.any?
+
+      # Option lookup used by /search/policy-papers-and-consultations/email-signup
+      # TODO: remove option_lookup and make it consistent with other finders.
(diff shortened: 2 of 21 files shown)


lib/parameter_parser/email_alert_parameter_parser.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+      []
+    end
+
+    def could_be_a_dynamic_facet?(key)
+      %w(organisations people world_locations all_part_of_taxonomy_tree part_of_taxonomy_tree).include? key
+    end
+
+    def parsed_params(filter_params, params)
```
hannako: I was just looking at this code in its current home in the `applied_filters` method. I think it's causing a couple of bugs. One of which is https://trello.com/c/tIR0DvbN/1248-business-finder-email-signup-errors-if-no-facet-options-selected.

On the signup page, the user can make a change to the filters they selected on the finder page. Currently their final choice isn't being reflected in what we send to email alert api. This sort of thing...
```
  def final_filter_choices
    filter_choices_from_finder_page = filter_params.fetch("subscriber_list_params", {}).keys
    filter_choices_from_signup_page   = filter_params.fetch("filter", {}).keys
    filter_choices_from_finder_page && filter_choices_from_signup_page
  end
```

bilbof: Thanks @hannako, let's get this one merged and then rebase your draft PR.

lib/parameter_parser/email_alert_parameter_parser.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    def permitted_filter_keys
+      allowed = permitted_facets.map { |facet| key_for_facet(facet) }
+      given = filter_params.keys
+      allowed & given
+    end
+
+    def permitted_values_for_filter_key(key)
+      facet = find_facet_by_key(key)
+      return if facet.blank?
```
barrucadu: `facet.blank?` can't ever be true here, right?  because `permitted_values_for_filter_key` is only called with values in `permitted_facets.keys`

lib/parameter_parser/email_alert_parameter_parser.rb
```diff
@@ -0,0 +1,98 @@
+# Ensures that email alert filter keys and values provided by users
+# and that Finder Frontend sends to Email Alert API are within the
+# set of permitted keys and values provided in the content item or
+# registries of people, organisations, taxons, and world locations.
+module ParameterParser
```
barrucadu: A bit of a design question: do we want this to be *invalid* if there's an unrecognised filter given?  That would send the user back to the signup page with the message that something's wrong, rather than sign them up for something which doesn't match what they entered (even if they did somehow enter a bad facet)

barrucadu: Same question for unrecognised values

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -22,9 +22,9 @@
         items: @signup_presenter.choices_formatted,
         visually_hide_heading: true
       } %>
-      <% if @error_message.present? %>
-        <p class="signup-choices__message"><%= @error_message %></p>
-      <% end %>
+    <% end %>
+    <% if @error_message.present? %>
```
barrucadu: Is the intent of this change to catch users submitting weird parameters in the URL?

lib/parameter_parser/email_alert_parameter_parser.rb
bilbof: Yeah the intent is to remove any invalid filter keys and values and continue to sign users up. What solution do you prefer?

My initial implementation raised an error if there were any invalid keys and returned users to the signup page. I think that'd be a worse user experience because users who have bad query params won't be able to fix them on the signup page. The user would see their final filters in the title of their subscriber list, so they have a final chance to go back and fix their subscription.

I think ideally we'd display facets on email signup pages, so users could modify their subscription. It may not be obvious to users that their filters from the finder will be used in their subscription, e.g. when signing up to [News from HMRC](https://www.gov.uk/search/news-and-communications/email-signup?organisations%5B%5D=hm-revenue-customs) it's not obvious that they'll only get news from HMRC.

app/views/email_alert_subscriptions/new.html.erb
bilbof: Yes

lib/parameter_parser/email_alert_parameter_parser.rb
barrucadu: Ah, I'd forgotten the list title includes what you're actually signing up for.  That's good, my concern was users being linked to a page which has a URL suggesting one thing, but they actually get signed up for a different thing.

bilbof:
@barrucadu thank you for your review - I've made those changes, please could I have a re-review?

PR #1841, opened 2020-01-06
Handle 422 from email alert api

bilbof:
Email Alert API can return a 422 (unprocessable) error. This adds more error handling to finder-frontend for this case.

Co-Authored-by: @MahmudH

Trello: https://trello.com/c/so20SdvH/808

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -5,15 +5,22 @@ class EmailAlertSubscriptionsController < ApplicationController
   helper_method :subscriber_list_params
 
   def create
-    if valid_choices?
-      redirect_to email_alert_signup_api.signup_url
-    else
-      @error_message = "Please choose an email alert"
-      render action: :new
-    end
+    validate_choices!
+    redirect_to email_alert_signup_api.signup_url
+  rescue MissingFiltersError
+    render_error "Please choose an email alert"
+  rescue EmailAlertSignupAPI::UnprocessableSubscriberListError
+    render_error("An error occured. Please check your filters and try again.")
   end
 
 private
+
+  class MissingFiltersError < StandardError; end
+
+  def render_error(error_message)
+    @error_message = error_message
+    render action: :new
+  end
 
   def content
     @content ||= fetch_content_item(request.path)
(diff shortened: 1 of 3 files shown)


app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -5,16 +5,25 @@ class EmailAlertSubscriptionsController < ApplicationController
...
-    else
-      @error_message = "Please choose an email alert"
-      render action: :new
-    end
+    validate_choices
+    redirect_to email_alert_signup_api.signup_url
+  rescue MissingFiltersError
+    render_error "Please choose an email alert"
+  rescue EmailAlertSignupAPI::UnprocessableSubscriberListError
+    render_error "Unprocessable request. Please check your filters and try again."
```
barrucadu: Is this message displayed to users?  If it is, I don't think "unprocessable request" is a very useful message to give.  Even "An error occurred" is probably better.

app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -31,6 +40,10 @@ def subscriber_list_params
     SubscriberListParamsPresenter.new(content, filter_params).subscriber_list_params
   end
 
+  def validate_choices
```
barrucadu: I think this should be `validate_choices!`, as the convention of using `!` for exception-throwing validation stuff seems pretty common

bilbof:
@barrucadu thanks for your comments! I made those changes.

app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -5,16 +5,23 @@ class EmailAlertSubscriptionsController < ApplicationController
...
-      redirect_to email_alert_signup_api.signup_url
-    else
-      @error_message = "Please choose an email alert"
-      render action: :new
-    end
+    validate_choices!
+    redirect_to email_alert_signup_api.signup_url
+  rescue MissingFiltersError
+    render_error "Please choose an email alert"
+  rescue EmailAlertSignupAPI::UnprocessableSubscriberListError
```
barrucadu: Should the `StandardError` case have been removed?

bilbof: Yeah, I think I'd prefer returning a 500 than a 200 in this case.

PR #1868, opened 2020-01-16
[WIP] Add "as you type" title suggestions to search/all

kr8n3r:
## What

Enhance search on search/all with suggestions from titles that match the keywords user is entering. 
Uses [Accessible Autocomplete](https://github.com/alphagov/accessible-autocomplete) engine.

## Why 
Hypothesis: We think that users don’t always know what they’re looking for, so if we provide autocomplete suggestions, they will refine their search less often.

## Performance
<details>
    <summary>Strategies</summary>

- we make use of a debounce function, so that we only trigger the lookup once 300ms have elapsed between keystrokes

- we cache last shown suggestions, so if users and more characters that result in a matching query or if they delete characters, we serve those instead of fetching from the API

</details>

[Trello card](https://trello.com/c/boqrYuXi/686-make-autocomplete-production-ready)

- https://finder-front-add-autoco-jp2dlr.herokuapp.com/search/all

Thanks to @kevindew and @sihugh for helping with CSP and Docker yarn stuff.

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -15,6 +15,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$label = this.$module.querySelector('label')
     this.$submitButton = this.$module.querySelector('button[type=submit]')
     this.$searchContainer = this.$module.querySelector('.gem-c-search')
+    // store user entered partial query for further use
+    this.userEnteredQuery = ''
     // use to store suggestions
     this.cachedSuggestions = []
 
@@ -128,12 +130,20 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     window.GOVUK.SearchAnalytics.trackEvent(
       'suggestionClicked',
-      trackingDataOptions.position
+      'click',
+      {
+        'dimension333': this.userEnteredQuery,
+        'dimension444': trackingDataOptions.text,
+        'dimension555': trackingDataOptions.position,
+        'dimension666': trackingDataOptions.numberOfSuggestions
+      }
     )
+    document.querySelector('meta[name="govuk:publishing-application"]').setAttribute('content', 'suggestion selected')
   }
 
   Autocomplete.prototype.handleSearchQuery = function (query, populateResults) {
     statusMessage = 'Searching...'
+    this.userEnteredQuery = query
 
     // check if current string exists in the cached array
     // if it does exists, return matching items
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -0,0 +1,175 @@
...
+    window.GOVUK.SearchAnalytics.trackEvent(
+      'suggestionClicked',
+      trackingDataOptions.position
+    )
+  }
+
+  Autocomplete.prototype.handleSearchQuery = function (query, populateResults) {
+    statusMessage = 'Searching...'
+    var matchingCachedResults = this.cachedSuggestions.filter(function (suggestion) {
+      return suggestion.toLowerCase().indexOf(query.toLowerCase()) !== -1
```
barrucadu: If you get suggestions for a query, and then delete a bunch of characters, won't this just return the same highly-specific suggestions, rather than widening to give more generic suggestions?

I was imagining something like this:

```js
if (!(query in cache)) {
    cache[query] = fetch_suggestions(query);
    // maybe some logic to keep the cache under a certain number of queries
}
return cache[query];
```

kr8n3r: hmm 
- When you type `breakfast` quickly, you get 2 results. 
If you then start deleting characters, suggestions don't change until you the base word (`break`)

- When you type `break` quickly, you get 8 results. 

when characters are deleted, the query variable gets updated and is checked against the strings stored in the cache

as both phrases contain the same base word (`break` in this case) and that exists in the cache, how can we know best to trigger the fetching of new suggestions?

barrucadu: I think fetching suggestions whenever that exact query isn't cached is probably fine, as the delay will prevent it from being every character (unless someone is typing slowly I guess)

kr8n3r: > ```js
> if (!(query in cache)) {
>     cache[query] = fetch_suggestions(query);
>     // maybe some logic to keep the cache under a certain number of queries
> }
> return cache[query];
> ```

this is essentially what we're doing here https://github.com/alphagov/finder-frontend/blob/c31e0f7c58fa34d717be3c09caaddb2380054fa2/app/assets/javascripts/components/autocomplete.js#L175
just not adding to the array, replacing it.

we could push to the array (at the start or at the end) and limit the array to N items, but I wonder if it's worth complicating the logic to ensure we have the latest items in the array.
It might not be needed, as I wonder hoe many queries will the user do before visiting a result and thus clearing the cached array?

Tetrino: I don't think we need a queue - I doubt the user will attempt more than a couple phrases before clicking through, so at most 40 suggestions assuming we're hitting the max of 8 returned suggestions per phrase. No need to overcomplicate.

benthorner:
@kr8n3r are we OK to close this?

PR #1870, opened 2020-01-16
Show facets by default on tablet / desktop

kr8n3r:
## What
Always show facets (filters) on finders on tablet/desktop. This PR removes code associated with the facet toggle functionality.

## Why
Following on from the[ A/B test we ran ](https://github.com/alphagov/finder-frontend/pull/1736) where our hypothesis was 

> that users who wouldn't use any other filter options are being confused by the keyword facets and if it weren't there they would either use the search bar to refine their search or not at all. 

The results tell us that users who did not see Facets after removing a facet did not then amend their search more. (removing a facet does not trigger and refinement but removing a facet and retyping does)

> When analysing the test we noticed that Search click through rate in variant B had increased by quite a large factor. This proved to be worth investigating as it seems that removal of Facets allows users to see results more quickly and click. Specifically on mobile. 

In conclusion,

> we will be making this change permanently as the results show it had a positive impact on users including some unexpected but welcome impacts.

[Trello ticket](https://trello.com/c/PtSyQ5hi/1285-show-facets-by-default)

- https://finder-frontend-pr-1870.herokuapp.com/search/all
- https://finder-frontend-pr-1870.herokuapp.com/search/research-and-statistics
- https://finder-frontend-pr-1870.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-1870.herokuapp.com/get-ready-brexit-check/questions
- https://finder-frontend-pr-1870.herokuapp.com/drug-device-alerts
- https://finder-frontend-pr-1870.herokuapp.com/find-eu-exit-guidance-business
- https://finder-frontend-pr-1870.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance

### DIFF
--- a/config/locales/en/finders/facet_collection.yml
+++ b/config/locales/en/finders/facet_collection.yml
@@ -1,5 +1,0 @@
-en:
-  facet_collection:
-    facet_disclosure:
-      more: "Show more search options"
-      fewer: "Show fewer search options"
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -401,11 +401,8 @@
   }
 
   LiveSearch.prototype.focusErrorMessagesOnLoad = function ($container) {
-    var $facetToggle = $container.find('.facet-toggle')
-    var facetsHidden = $facetToggle.attr('aria-expanded') === 'false'
     var $inputWithError = $container.find('input[class*=--error]')
-    if (facetsHidden && $inputWithError.length) {
-      $facetToggle.click()
+    if ($inputWithError.length) {
       $inputWithError.focus()
     }
   }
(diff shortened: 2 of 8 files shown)


config/locales/en/finders/facet_collection.yml
```diff
@@ -1,5 +1,2 @@
 en:
   facet_collection:
-    facet_disclosure:
-      more: "Show more search options"
-      fewer: "Show fewer search options"
```
bilbof: Could this file be deleted?

kr8n3r: done

PR #1871, opened 2020-01-17
Add a registry for roles

thomasleese:
This will be used to support filtering documents on the role.

Depends on https://github.com/alphagov/search-api/pull/1922 to include `title`s and `link`s in the registry.

### DIFF
--- a/lib/registries/roles_registry.rb
+++ b/lib/registries/roles_registry.rb
@@ -1,0 +1,50 @@
+module Registries
+  class RolesRegistry < Registry
+    include CacheableRegistry
+
+    def [](slug)
+      roles[slug]
+    end
+
+    def roles
+      @roles ||= fetch_from_cache
+    end
+
+    def values
+      roles
+    end
+
+    def cache_key
+      "#{NAMESPACE}/roles"
+    end
+
+  private
+
+    def report_error
+      GovukStatsd.increment("registries.roles_api_errors")
+    end
+
+    def cacheable_data
+      roles_as_hash
+    end
+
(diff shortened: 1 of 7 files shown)


lib/registries/roles_registry.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    def roles_as_hash
+      GovukStatsd.time("registries.roles.request_time") do
+        roles = fetch_roles_from_rummager || {}
+
+        roles.reject { |result| result.dig("value", "slug").blank? || result.dig("value", "title").blank? }
+          .each_with_object({}) { |result, orgs|
+            slug = result["value"]["slug"]
+            orgs[slug] = result["value"].slice("title", "slug", "content_id")
```
bilbof: The object here could be changed from `orgs` to `roles`

lib/registries/roles_registry.rb
```diff
@@ -0,0 +1,51 @@
...
+          }
+      end
+    end
+
+    def fetch_roles_from_rummager
+      params = {
+        facet_roles: "1500,examples:0,order:value.title",
```
bilbof: I'd use `aggregate_roles` since `facet_` is [to be deprecated](https://github.com/alphagov/search-api/blob/859884e85a087a9f75e05f7c84f30b2a7f2b032b/lib/parameter_parser/aggregates_parameter_parser.rb#L15). Elasticsearch has replaced facets with aggregations since we started using it: https://www.elastic.co/guide/en/elasticsearch/reference/7.5/search-facets.html

PR #1873, opened 2020-01-21
Spike into querying for eligibility on knowledge graph

oscarwyatt:
This is a spike into querying the knowledge graph to find content for which a person might be eligible for.

What's interesting is that on starting I realised what we're actually finding is things that we're not _ineligible_ for - if someone has only entered their age it might filter some things out but we're not necessarily eligible for everything else because we haven't entered enough information yet. As users enter more information the number of results will fall. This does mean that occasionally we end up with a double negative like 'not_ineligible' which is slightly eggy. I think this is something that needs a little more thought and clarification, perhaps with a product steer

There are ways to make this faster - for example making a query to find things that we know we're ineligible for (eg AgeRequirement where age < min_age or age > max_age), instantiating the content item and then doing a full filter to see if we can be certain we're not ineligible for that service (as some filters contain OrGroupRequirements, being ineligible for one requirement does not guarantee that, although if all parents of a FooRequirement are AndRequirementGroups that does guarantee that the user is not eligible). For the time being, such approaches add a lot more complexity and as for the short to medium term we will only have a small number of content items with eligibility criteria, there will be precious little speed improvement in doing so and simply iterating over any Cids with eligibility will be ok.

NB: The neo4j gem is quite large so I haven't had a proper look into it as I felt this is something to do if/when we think this approach is worth persuing

### To try it out (with docker)

* Add an environment variable for the key NEO4J_URL with the URL to a remote instance of the knowledge graph (or leave it blank to use a local copy (untested)). This takes the form `http://neo4j:password@localhost:7474`. 
* Run this branch of Finder Frontend with a bash console
* run `rake neo4j:migrate` (this only has to be done once a day, it may be possible to not have to do this but I didn't want to get too bogged down in it)
and open a console. 
* Running `Cid.test_toy_example` will delete any existing eligibilities, insert some test data into the database and run a few tests to ensure everything works as expected. You can run the query `MATCH p=()-[r:HAS_REQUIREMENT|HAS_ELIGIBILITY]->() RETURN p LIMIT 25` in the database to see the things you just inserted.

### DIFF
--- a/app/models/age_requirement.rb
+++ b/app/models/age_requirement.rb
@@ -1,0 +1,32 @@
+class AgeRequirement
+  include Neo4j::ActiveNode
+  has_many :out, :requirements, type: :HAS_REQUIREMENT, model_class: %i(OrRequirementGroup AndRequirementGroup)
+  property :min_age
+  property :max_age
+
+  def ineligible?(attributes)
+    return false if attributes[:age].blank?
+
+    met_requirements = requirements_present.map { |requirement| send(requirement, attributes) }
+    !met_requirements.all?
+  end
+
+private
+
+  def requirements_present
+    criteria = { min_age: "meets_min_age_requirement?", max_age: "meets_max_age_requirement?" }
+    %i(min_age max_age).each_with_object([]) do |requirement, requirements|
+      if read_attribute(requirement).present?
+        requirements << criteria[requirement]
+      end
+    end
+  end
+
+  def meets_min_age_requirement?(attributes)
+    attributes[:age] >= min_age
+  end
+
+  def meets_max_age_requirement?(attributes)
+    attributes[:age] <= max_age
+  end
+end
--- a/app/models/cid.rb
+++ b/app/models/cid.rb
@@ -1,0 +1,21 @@
+class Cid
+  include Neo4j::ActiveNode
+
+  property :documentType
+  property :contentID
+  property :name
+  property :description
+  property :pagerank
+  property :title
+
+  has_many :out, :eligibilities, type: :HAS_ELIGIBILITY, model_class: :Eligibility
+
+  def self.where_not_ineligible(attributes)
+    Eligibility.
+      all.
+      reject { |eligibility| eligibility.ineligible?(attributes) }.
+      map(&:cids).
+      flatten.
+      compact
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gem "govuk_publishing_components", "~> 21.20.0"
 gem "rails", "~> 6.0.2"
 gem "slimmer", "~> 13.2.0"
 
+gem "neo4j", "~> 9.6", ">= 9.6.1"
 gem "sass-rails", "~> 6.0.0"
 gem "uglifier", "~> 4.2"
 gem "whenever", "~> 1.0.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -120,11 +120,18 @@ GEM
       dotenv (= 2.7.5)
       railties (>= 3.2, < 6.1)
     erubi (1.9.0)
+    ethon (0.12.0)
+      ffi (>= 1.3.0)
     execjs (2.7.0)
     factory_bot (5.1.1)
       activesupport (>= 4.2.0)
     faraday (0.17.3)
       multipart-post (>= 1.2, < 3)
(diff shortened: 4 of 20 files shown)


app/models/cid.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    eligibility = Eligibility.create()
+    content_page = Cid.find_by(contentID: "f4b96a38-5247-4afd-b554-8a258a0e8c93")
+    content_page.eligibilities << eligibility
+
+    base_requirement = AndRequirementGroup.create()
+    eligibility.requirement = base_requirement
+
+    # be at least 16 and under 19 on 31 August 2019
+    age_requirement = AgeRequirement.create(min_age: 16, max_age: 18)
```
karlbaker02: Perhaps a bit pedantic with language here, but is _under 19_ technically the same as `18 <= 18`? (for example, if you're nearly 19, you're over 18y 0m).

app/models/age_requirement.rb
```diff
@@ -0,0 +1,27 @@
+class AgeRequirement
+  include Neo4j::ActiveNode
+  has_many :out, :requirements, type: :HAS_REQUIREMENT, model_class: [:OrRequirementGroup, :AndRequirementGroup]
+  property :min_age
+  property :max_age
+
+  def ineligible?(attributes)
+    return false unless attributes[:age].present?
+
+    met_requirements = requirements_present.map { |requirement| send("meets_#{requirement}_requirement?", attributes) }
```
bevanloon: Personal preference - I don't really like using `send` because it obfuscates where and when methods are being called, but I can see why you've used it here. Is it optimising too soon though? 🤷‍♂

app/models/age_requirement.rb
```diff
@@ -0,0 +1,27 @@
...
+    !met_requirements.all?
+  end
+
+  private
+
+  def requirements_present
+    %i(min_age max_age).select { |requirement| read_attribute(requirement).present? }
+  end
```
bevanloon: As a suggestion, could we eliminate the string interpolation `ineligible?` by doing something like:

```
criteria = {min_age: "meets_min_age_requirement?", max_age: "meets_max_age_requirement?}
%(min_age max_age).select do |requirement|
  criteria[requirement] if read_attribute(requirement).present?
end
```

then in `ineligible?` we could just `send(requirement)`? 

Haven't tested it, so I'm not sure if that would work.

app/models/cid.rb
```diff
@@ -0,0 +1,127 @@
...
+  def self.where_not_ineligible(attributes)
+    Eligibility.
+      all.
+      reject { |eligibility| eligibility.ineligible?(attributes) }.
+      map { |eligibilty| eligibilty.cids }.
+      flatten.
+      compact
+  end
+
+  def self.test_toy_example
```
bevanloon: I know this is only a spike, but is this perhaps better extracted and expressed as a rake task or a spec just for ease of running it?

oscarwyatt: True, I've moved it all to one giant test, obvs this'd want splitting up at some point but not yet

app/models/age_requirement.rb
oscarwyatt: This is a much better idea, changed

app/models/age_requirement.rb
oscarwyatt: Yeah, I'm not a huge fan. I think your suggestion is an excellent way to deal with it. The only alternative I could think of was mildly hideous nested if statements and this felt slightly better but....

bevanloon: Yep, it does feel a bit "six of one, half a dozen of the other".  

I think I probably would have done something like:

```
def ineligible?(attributes)
  return false unless met_min_age_requirement?(attributes)
  return false unless met_max_age_requirement?(attributes)
  true
end

def meets_min_age_requirement?(attributes)
  return false unless read_attribute(:min_age).present?
  attributes[:age] >= min_age
end
.
.
.
```

to avoid any nested `ifs`. Trade offs though - `read_attribute` would be duplicated in each `_requirement?` method, but there's no meta-programming. Also scaling this with more criteria might be annoying. 🤷‍♂ 

It's not really a better solution though, just different.

benthorner:
@oscarwyatt are we OK to close this?

oscarwyatt:
@benthorner Go ahead

PR #1874, opened 2020-01-21
Ensure required filters have values at email signup

bilbof:
This will enforce a requirement that required facets (such as case type on the CMA cases email signup) have a value.

It should be deployed once the finders using the required field have been republished.

Example: http://finder-frontend-pr-1874.herokuapp.com/cma-cases/email-signup

Related:
https://github.com/alphagov/search-api/pull/1926
https://github.com/alphagov/specialist-publisher/pull/1564
https://github.com/alphagov/govuk-content-schemas/pull/952

https://trello.com/c/BOUerRPp/1252

### DIFF
--- a/spec/controllers/email_alert_subscriptions_controller_spec.rb
+++ b/spec/controllers/email_alert_subscriptions_controller_spec.rb
@@ -91,21 +91,15 @@ describe EmailAlertSubscriptionsController, type: :controller do
         stub_content_store_has_item("/cma-cases/email-signup", cma_cases_signup_content_item)
       end
 
-      context "when no filters are provided" do
-        it "redirects to the subscription url" do
-          email_alert_api_has_subscriber_list(
-            "tags" => {
-              "format" => { any: %w[cma_case] },
-            },
-            "subscription_url" => "http://www.gov.uk/default-subscription-to-cma-cases",
-          )
-
+      context "when a required filter is not provided" do
+        it "returns the user to the signup page with an error" do
           post :create, params: { slug: "cma-cases" }
-          expect(subject).to redirect_to("http://www.gov.uk/default-subscription-to-cma-cases")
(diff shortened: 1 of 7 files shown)


app/models/content_item.rb
```diff
@@ -6,6 +6,10 @@ def initialize(content_item_hash)
   end
 
   def self.from_content_store(base_path)
+    if base_path == "/cma-cases/email-signup"
+      return self.new(JSON.parse(File.read(Rails.root.join("features/fixtures/cma_cases_signup_content_item.json"))))
```
barrucadu: ?

bilbof: 🤕 oops!

spec/controllers/email_alert_subscriptions_controller_spec.rb
```diff
@@ -91,21 +91,15 @@
...
-            },
-            "subscription_url" => "http://www.gov.uk/default-subscription-to-cma-cases",
-          )
-
+      context "when a required filter is not provided" do
+        it "returns the user to the signup page with an error" do
           post :create, params: { slug: "cma-cases" }
-          expect(subject).to redirect_to("http://www.gov.uk/default-subscription-to-cma-cases")
+          expect(response).to be_successful
+          expect(response).to render_template("new")
```
barrucadu: Is it right that this successfully renders the `new` page, rather than redirecting?

bilbof: Yeah, now that CMA cases has a facet that requires a value (I changed the fixture in finder-frontend to reflect the change [here](https://github.com/alphagov/specialist-publisher/pull/1564/files#diff-63cd0912e8dee64396223c25801cccd4R30)), we want to ensure a user that doesn't set it will render `new` rather than creating a subscriber list.

PR #1910, opened 2020-02-03
Return T086 to notifications

huwd:
We removed a whole bunch of actions as part of the pre-brexit check work.

Finder frontend has some data integrity checks that rightly states there shouldn’t be a notification for an action that doesn’t exist.

In this case there are a bunch of notifications that exist for actions that temporarily do not. They are left commented rather than removed as the data is and will be useful in the future and should shortly be returned.

Periodically I am checking to see which of these actions have returned and am uncommenting the relevant notifications.

In this case, only one notification returns T086.


huwd:
Closed in favor of: https://github.com/alphagov/finder-frontend/pull/1917

PR #1911, opened 2020-02-03
Change show only brexit results to show only transition period results

huwd:
Trello: https://trello.com/c/Uanimli2/415-change-brexit-to-transition-period-on-search-checkboxes

News and communications finders have a special checkbox to limit all searches to brexit materials.
The brexit taxon has now been moved to transition. Whilst the check continues to work it is now mis-labelled.

This PR changes these expectations in fixtures so that they match the parallel work in [search api](https://github.com/alphagov/search-api/pull/1961), as well as fixing the cucumber tests.

It also takes the opportunity to rename the key from `related_to_brexit` to `related_to_transition_period`


features/fixtures/news_and_communications.json
```diff
@@ -89,11 +89,11 @@
         "preposition": "about"
       },
       {
-        "key": "related_to_brexit",
+        "key": "related_to_transition_period",
```
sihugh: Are we catering for redirects?  We know that people have RSS feeds set up for this, and changing the key will stop these from working unless we do something about it.

huwd: ah... humm.... that's a really good shout and a pain...
Ideally we don't want any user facing references to brexit anymore as all this content should now be about transition.

... can we redirect on a param?

huwd: Humm ok, rolling this back for now in an effort to separate out the content change from the key change.

PR #1924, opened 2020-02-06
Remove QA Content Feature

huwd:
Trello: https://trello.com/c/qBM5QLID/482-business-finder-tech-debt

We are now in a position to wind down the Business finder tech debt.
This was implemented with a QA content feature which is no longer being used. This PR removes all this code, to simplify finder-frontend.


koetsier:
- FacetExtractor can be removed / simplified

huwd:
@koetsier so looks like that is still used:
- [content_item.rb](https://github.com/alphagov/finder-frontend/blob/83d436901469b84fd9c42e5636019b39dae943a1/app/models/content_item.rb#L137)
- [facets_builder.rb](https://github.com/alphagov/finder-frontend/blob/83d436901469b84fd9c42e5636019b39dae943a1/lib/facets_builder.rb#L28)

Not really clear on how these work in other features... could we have a run through of what's going on here?

vanitabarrett:
@huwd I wasn't aware  https://www.gov.uk/uk-nationals-living-eu used the QA feature, but it looks like it does and is still live... so maybe we can't delete it quite yet. It's only one question though, so perhaps we can remove some code specific to the business finder, but it might be a more difficult job than just ripping it all out.

huwd:
@vanitabarrett I've removed the stuff that touched the `QaToContentController`.
The only bit I'm not sure about [is this tracking](https://github.com/alphagov/finder-frontend/pull/1924/files#diff-e1ff46625134b9f4a5a8424b263484a0), I'm not clear on if this should have also applied to the `/uk-nationals-living-eu` feature.

Would you be free to take a quick look over with me?

huwd:
>     * FacetExtractor

@koetsier does it make sense to do that one in a separate PR?

app/assets/stylesheets/application.scss
```diff
@@ -6,5 +6,5 @@ $govuk-use-legacy-palette: false;
 @import 'components/*';
 @import 'finder_frontend';
 @import 'views/search';
-@import 'views/qa';
+@import 'views/brexit_checker_question_page';
```
huwd: This is a rename - I removed all the qa related css, but there was one class that still seemed relevant to the brexit_checker, so i've renamed it and stuck it in it's own file below.

config/routes.rb
```diff
@@ -29,9 +29,6 @@
   get "/*slug/email-signup" => "email_alert_subscriptions#new", as: :new_email_alert_subscriptions
   post "/*slug/email-signup" => "email_alert_subscriptions#create", as: :email_alert_subscriptions
 
-  # Q&A frontend for "Find EU Exit guidance for your business" (www.gov.uk/find-eu-exit-guidance-business)
-  get "/prepare-business-uk-leaving-eu" => "qa#show"
-
   # Q&A frontend for "UK Nationals in the EU" (www.gov.uk/uk-nationals-in-the-eu)
   get "/uk-nationals-living-eu" => "qa_to_content#show"
```
huwd: Leaving this for now, it's still relevant.

app/assets/stylesheets/views/_brexit_checker_question_page.scss
```diff
@@ -0,0 +1,3 @@
+.govuk-form-footer {
```
vanitabarrett: In itself this feels like a bit of tech debt. This class looks like it comes from the design system (with the `.govuk-` prefix), but I can't find any record of it there. If it's not, we should rename it to make it clearer that it's only used in the checker.

Perhaps a separate card, but I think we need to look at:

1) If we actually need this any more
2) If we can replace this with something more specific

huwd: You happy for this to live orphaned up here for the moment and then we can drop in another card for planning to fix this for good?

vanitabarrett:
@huwd I think you're right about the tracking, but I'm just double checking with Dilwoar as I've found two similar things: track-qa-choices and track-brexit-qa-choices and neither look to be used (apart from in the business finder)

app/assets/stylesheets/views/_brexit_checker_question_page.scss
vanitabarrett: Yep! Perhaps create a card and add it as a comment here?

huwd: https://trello.com/c/i3nYBM2c/403-rename-brexit-checker-css-in-finder-frontend

PR #1928, opened 2020-02-10
Add retirement question to brexit checker

huwd:
Trello: https://trello.com/c/UUlLUAEj/461-add-option-for-retired-users-to-the-checker

Currently does not have any actions or grouping associated but serves a purpose by adding an option for users to create email subscriptions against or save their results URL


vanitabarrett:
@huwd Looks good! Is this ready to go? The Trello card you linked to refers to the SIC code prototype so I wasn't sure if this is ok to go live now.

PR #1953, opened 2020-02-21
Reduce compiled CSS file size

kr8n3r:
## What
Remove the disabled sass compression setting that was a workaround for failed sass-rails v6 compilation when we upgraded to rails 6.

As investigated in [this issue](https://github.com/alphagov/finder-frontend/issues/1124), sass will fail to compile with the error
```
SassC::SyntaxError: Error: "calc(0px)" is not a number for `max'
        on line 1181:20 of stdin, in function `max`
        from line 1181:20 of stdin
>> @supports (margin: max(calc(0px))) {
```
as sass-rails v6 moved the implementation to [sassc-rails](https://github.com/rails/sass-rails/releases/tag/v6.0.0)

We have now pinned sass-rails to version 5 which still uses Libsass

## Why
Reduce the weight of assets we ship to visitors, increase performance.

### Current production with GZIP
https://www.gov.uk/search/all
Screenshot 2020-02-21 at 13 24 55

### Tests on integration (with GZIP)

**Before**
before

**After**
after

- https://finder-front-pin-sass-r-b5hbtd.herokuapp.com/search/all
- https://finder-front-pin-sass-r-b5hbtd.herokuapp.com/search/research-and-statistics

[Trello ticket](https://trello.com/c/F872Wkxo/1407-reduce-the-size-of-css-bundle)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem "govuk_publishing_components", "~> 21.25.0"
 gem "rails", "~> 6.0.2"
 gem "slimmer", "~> 13.2.0"
 
-gem "sass-rails", "~> 6.0.0"
+gem "sass-rails", "~> 5.1.0"
 gem "uglifier", "~> 4.2"
 gem "whenever", "~> 1.0.0"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -345,8 +345,12 @@ GEM
     sass-listen (4.0.0)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
-    sass-rails (6.0.0)
-      sassc-rails (~> 2.1, >= 2.1.1)
+    sass-rails (5.1.0)
+      railties (>= 5.2.0)
+      sass (~> 3.1)
+      sprockets (>= 2.8, < 4.0)
+      sprockets-rails (>= 2.0, < 4.0)
+      tilt (>= 1.1, < 3)
     sassc (2.2.1)
       ffi (~> 1.9)
     sassc-rails (2.1.2)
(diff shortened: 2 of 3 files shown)


nickcolley:
What is the impact when taking into account GZIP? Since for CSS the main priority is bytes over the network it may be that the compression before GZIP does not make a big difference since GZIP is very good at compressing whitespace.

Gemfile
```diff
@@ -13,7 +13,7 @@ gem "govuk_publishing_components", "~> 21.24.0"
 gem "rails", "~> 6.0.2"
 gem "slimmer", "~> 13.2.0"
 
-gem "sass-rails", "~> 6.0.0"
+gem "sass-rails", "< 6"
```
bilbof: Those size improvements look impressive! 👍 

Could you provide a bit more context about this change? I am not clear on why we'd prefer to use a version of sass-rails below 6 here. I wan't able to reproduce the change when pulling this branch, but I may have missed a step

kr8n3r: i've update the description. for reasons it doesn't fail locally, but see this [commit](https://github.com/alphagov/finder-frontend/pull/1953/commits/675f6efa1fc45ec96683d583025d7f55726219fb) for failed build

kr8n3r:
> What is the impact when taking into account GZIP? Since for CSS the main priority is bytes over the network it may be that the compression before GZIP does not make a big difference since GZIP is very good at compressing whitespace.

this is with GZIP

kr8n3r:
Commit [675f6ef](https://github.com/alphagov/finder-frontend/pull/1953/commits/675f6efa1fc45ec96683d583025d7f55726219fb) is to show the compilation error and will be removed

Gemfile
kr8n3r: pinned to specific version

PR #1989, opened 2020-03-05
Make transition checker email clearer

laurentqro:
https://trello.com/c/WJlt625D/563-clarify-action-updated-email

### DIFF
--- a/app/mailers/brexit_checker_mailer.rb
+++ b/app/mailers/brexit_checker_mailer.rb
@@ -4,12 +4,12 @@ class BrexitCheckerMailer < ApplicationMailer
   def change_notification(notification)
     @notification = notification
     @action = @notification.action
-    mail(subject: subject)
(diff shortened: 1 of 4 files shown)


benthorner:
Just pasting this here from the card, since we have a PR...

> I think we need to be careful with the changes proposed here, because of digest emails. Looking > at the PR, we'd end up with digest emails like:
>
> ```
> Number of coronavirus (COVID-19) cases and risk in the UK
>
> [some other content title]
>
> A new action has been added <-- to what??
>
> A new action has been added
> ```
>
>The rolling-up of individual emails into digests adds constraints to the content we can put in different places.

vanitabarrett:
I'm going to close this PR as we've decided to stop work on this as the original content designers have moved on from the team. We will revisit this if it comes up again in our email template consolidation work.

PR #2052, opened 2020-04-24
Don't render non-finder content items as search pages

bilbof:
This will help if we forget to unpublish a finder.

We have some quite interesting behaviour at the moment. Finder frontend will try to render any content item as a search page, which leads to interesting search results.

This will make things less interesting unfortunately.

Links to check:

_There's no link for an unpublished finder unfortunately, but the tests should cover it_

404: http://finder-front-bilbof-404-ianvvm.herokuapp.com/blah
200: https://finder-front-bilbof-404-ianvvm.herokuapp.com/search/all
200: http://finder-front-bilbof-404-ianvvm.herokuapp.com/uk-nationals-living-eu

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -12,6 +12,8 @@ class FindersController < ApplicationController
   def show
     respond_to do |format|
       format.html do
+        raise UnsupportedContentItem unless content_item.is_finder?
+
         show_page_variables
       end
       format.json do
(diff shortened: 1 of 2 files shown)


app/controllers/finders_controller.rb
```diff
@@ -12,6 +12,8 @@ class FindersController < ApplicationController
   def show
     respond_to do |format|
       format.html do
+        raise UnsupportedContentItem unless content_item.is_search? || content_item.is_finder?
```
barrucadu, line 15: I think the only page on GOV.UK with a `search` content item is /search, which doesn't hit this controller.  So this condition can just be `unless content_item.is_finder?`

bilbof: @barrucadu thanks, I've changed that.

PR #2061, opened 2020-05-06
Update Transition Checker content

chao-xian:
https://trello.com/c/Cdcjn2tA

https://finder-front-transition-gemt9f.herokuapp.com/transition-check/questions

### DIFF
--- a/lib/brexit_checker/actions.yaml
+++ b/lib/brexit_checker/actions.yaml
@@ -212,7 +212,7 @@ actions:
     charged for NHS services.
   guidance_prompt: More information
   guidance_link_text: Living in Europe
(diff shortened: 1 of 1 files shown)


lib/brexit_checker/actions.yaml
```diff
@@ -424,12 +461,14 @@ actions:
...
-    - personal-eu-org-provide
-    - ip-copyright
+  - all_of:
+    - any_of:
+      - media
+      - digital
+      - telecoms
+    - any_of:
+      - personal-eu-org-provide
+      - ip-copyright
```
laurentqro, line 471: this is not on the trello card, so we should probably not commit this update here.

chao-xian: Removed

vanitabarrett:
@chao-xian This looks good to me. Just to double check, there are a few changes that aren't mentioned in the Trello card - is that intentional?

PR #2078, opened 2020-05-18
Expand taxon registry in finder-frontend to render email signup choices

laurentqro:
Changes the email signup link on [topic pages](https://www.gov.uk/money) to go to [/search/all/email-signup](https://www.gov.uk/search/all/email-signup) with the appropriate taxon pre-selected.

This works for any taxon, at any level in the taxon tree.

https://trello.com/c/r7XXLulF/303-expand-taxon-registry-in-finder-frontend-to-render-email-signup-choices

---

- https://finder-front-taxon-sign-7gfoam.herokuapp.com/search/all
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/search/research-and-statistics
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/get-ready-brexit-check/questions
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/drug-device-alerts
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/find-eu-exit-guidance-business
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/find-eu-exit-guidance-business?keywords=eori&order=relevance
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/find-eu-exit-guidance-business?sector_business_area%5B%5D=aerospace&order=topic
- https://finder-front-taxon-sign-7gfoam.herokuapp.com/uk-nationals-living-eu

### DIFF
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -30,11 +30,7 @@ private
   end
 
   def signup_presenter
-    @signup_presenter ||= SignupPresenter.new(content_item, taxon_facet, params)
-  end
-
-  def taxon_facet
-    TaxonFacet.new({}, filter_params.slice("level_one_taxon", "level_two_taxon", "topic"))
+    @signup_presenter ||= SignupPresenter.new(content_item, params)
   end
 
   def subscriber_list_params
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -34,7 +34,7 @@ private
   end
 
   def taxon_facet
-    TaxonFacet.new({}, filter_params.slice("level_one_taxon", "level_two_taxon"))
+    TaxonFacet.new({}, filter_params.slice("level_one_taxon", "level_two_taxon", "topic"))
   end
 
   def subscriber_list_params
--- a/app/controllers/email_alert_subscriptions_controller.rb
+++ b/app/controllers/email_alert_subscriptions_controller.rb
@@ -1,7 +1,7 @@
 class EmailAlertSubscriptionsController < ApplicationController
   layout "finder_layout"
   protect_from_forgery except: :create
-  before_action :signup_presenter
+  before_action :signup_presenter, :set_taxon
   helper_method :subscriber_list_params
 
   def create
@@ -25,20 +25,26 @@ private
     render action: :new
   end
 
-  def content
-    @content ||= ContentItem.from_content_store(request.path).as_hash
+  def content_item
+    @content_item ||= ContentItem.from_content_store(request.path).as_hash
   end
 
   def signup_presenter
-    @signup_presenter ||= SignupPresenter.new(content, params)
+    @signup_presenter ||= SignupPresenter.new(content_item, params)
+  end
+
+  def set_taxon
+    content_id = filter_params[:level_one_taxon]
+    @registry ||= Registries::BaseRegistries.new
+    @taxon ||= @registry.topic_taxon_with_content_id(content_id)
   end
 
   def subscriber_list_params
-    SubscriberListParamsPresenter.new(content, filter_params).subscriber_list_params
+    SubscriberListParamsPresenter.new(content_item, filter_params).subscriber_list_params
   end
 
   def email_alert_filter_params
-    @email_alert_filter_params ||= ParameterParser::EmailAlertParameterParser.new(content, filter_params, params)
+    @email_alert_filter_params ||= ParameterParser::EmailAlertParameterParser.new(content_item, filter_params, params)
   end
 
   def validate_choices!
@@ -53,8 +59,8 @@ private
   def email_alert_signup_api
     EmailAlertSignupAPI.new(
       applied_filters: applied_filters,
-      default_filters: content["details"].fetch("filter", {}),
-      facets: content["details"].fetch("email_filter_facets", []),
+      default_filters: content_item["details"].fetch("filter", {}),
+      facets: content_item["details"].fetch("email_filter_facets", []),
       subscriber_list_title: subscriber_list_title,
       email_filter_by: signup_presenter.email_filter_by,
     )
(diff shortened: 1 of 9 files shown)


app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -25,20 +25,26 @@ def render_error(error_message)
...
+  def content_item
+    @content_item ||= ContentItem.from_content_store(request.path).as_hash
   end
 
   def signup_presenter
-    @signup_presenter ||= SignupPresenter.new(content, params)
+    @signup_presenter ||= SignupPresenter.new(content_item, params)
+  end
+
+  def set_taxon
```
benthorner, line 36: Would it be possible to use the [taxon_facet](https://github.com/alphagov/finder-frontend/blob/master/app/models/taxon_facet.rb) instead of this?

(I think that's the direction we'll head in for the actual implementation, so worth exploring what re-using the existing code might look like.)

laurentqro: @benthorner I refactored to reuse as much of TaxonFacet as I could. It's indeed a bit nicer than before. TaxonFacet logic is still very much tailored to dropdown selects, so that's somewhat limiting.

The next thing I want to tackle is radio button inputs with different names (`level_one_taxon` and `level_two_taxon`). On email-alert-frontend, all radio buttons have `name = "topic"`.

benthorner: Thanks for looking into that @laurentqro 👍. I would have thought that, so long as the radio inputs use those `level_` names, then everything should just work. Maybe I'm missing something subtle - that's quite likely for a new codebase.

The other thing I'm wondering is: how could this fit into the existing SignupPresenter? We likely want that to be where the code goes eventually. Maybe that can be a stretch goal - it could be more of a hindrance to the spike, than an aid to it.

laurentqro: @benthorner I'll see how we could fit the logic into SignupPresenter 👌

app/models/taxon_facet.rb
```diff
@@ -118,7 +136,11 @@ def default_topic_value
     { text: "All topics", value: "" }
   end
 
-  def registry
-    @registry ||= Registries::BaseRegistries.new.all["part_of_taxonomy_tree"]
+  def full_topic_taxonomy
```
benthorner, line 139: Could we rename these (at the Registry level), so their usage is clearer? Maybe something for another PR, although we could use clearer names within this class. 

Example:

- `full_topic_taxonomy` -> `flat_topic_taxonomy_hash`
- `partial_topic_taxonomy` -> `upper_topic_taxonomy_tree`

app/models/taxon_facet.rb
```diff
@@ -43,8 +43,25 @@ def query_params
...
+  def selected_taxon_value
+    selected_level_one_value || selected_level_two_value || selected_topic_value
+  end
+
 private
 
+  def selected_topic_value
+    return unless @value_hash["topic"]
+
+    content_id = ContentItem.from_content_store(@value_hash["topic"]).content_id
```
benthorner, line 55: I see why this is necessary, because the incoming param is a base path, and we need a Content ID. There are a couple of downsides to this:

- We have to do the extra API call
- It's inconsistent with the other params, which are Content IDs

Since we're going to have to repoint all the existing links anyway, how about we assume the incoming param is a Content ID? Then we can avoid doing a lookup.

app/models/taxon_facet.rb
```diff
@@ -43,8 +43,25 @@ def query_params
     }
   end
 
+  def selected_taxon_value
+    selected_level_one_value || selected_level_two_value || selected_topic_value
```
benthorner, line 47: I think the `selected_level_two_value` needs to have precedence here, so that we show the most specific set of child taxons. Can we add a test for that?

app/controllers/email_alert_subscriptions_controller.rb
```diff
@@ -25,20 +25,24 @@ def render_error(error_message)
...
+  def content_item
+    @content_item ||= ContentItem.from_content_store(request.path).as_hash
   end
 
   def signup_presenter
-    @signup_presenter ||= SignupPresenter.new(content, params)
+    @signup_presenter ||= SignupPresenter.new(content_item, params)
+  end
+
+  def selected_taxon
```
benthorner, line 36: We should move this into the `SignupPresenter`, since it's a presentation concern.

app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -8,38 +8,85 @@
...
+
+    <%= form_tag email_alert_subscriptions_path(subscriber_list_params: subscriber_list_params), class: 'signup-choices' do %>
+      <p>Everything published to GOV.UK about</p>
+
+      <%= render "govuk_publishing_components/components/radio", {
+        name: "subscriber_list_params[all_part_of_taxonomy_tree]",
+        items: [{
+          value: @selected_taxon[:value],
+          text: @selected_taxon[:text],
+          checked: true,
```
benthorner, line 28: Nice one preserving this!

app/models/taxon_facet.rb
benthorner: (I think this was another one of those implicit assumptions in my head. Sorry!)

benthorner:
I thought of one other functional issue we need to address: if we merge this now, it will change the email signup interface for all search pages where a sub/topic has been selected.

This would be bad because the email signup page should primarily be focussed on the search page the user has come from, although there are definite design issues here too.

I can think of a few options that allow us to proceed:

- Put the new behaviour behind a feature flag (e.g. a query param). We should only do this once we have a clear idea of the next steps; otherwise, we could be merging redundant code.

- Make the new feature only work when the "topic" param is specified. This is basically the same as the above, but allows us to start repointing taxon subscriptions.

- Make the new feature only work with existing search pages i.e. remove "topic" support. This means we can start using the new feature, but requires design changes now.

We also seem to have a problem with the subscriber lists this is going to generate, which is unrelated to this PR, but blocks us from repointing taxon subscriptions. Steps to reproduce:

- Go to `gov.uk/search/all` (the general search page)
- Select "Business and Industry" (other topics seem to work OK)
- Append `/email-signup` to the search params
- Click through to the email entry screen
- See the title is "**Statistics** with topic of Business and industry"

Looks like we need to do an audit of the subscriber lists generated by search pages, as some appear to have incorrect titles (title should be "All content").

benthorner:

benthorner:
Closing as this work was paused long ago.

PR #2117, opened 2020-06-18
Add tracking on Subscribe button for Transition alerts

chao-xian:
We want to be able to track user journeys on email signups.

https://trello.com/c/mTcA2hFA

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,7 +21,7 @@ jQuery(function ($) {
 
   var $atomAutodiscoveryLink = $("link[type='application/atom+xml']").eq('0')
 
-  var $transitionSubscribeButton = document.querySelector('#checklist-email-signup button.js-subscribe-button')
+  var $transitionSubscribeButton = document.querySelector('.js-subscribe-button-tracker')
 
   $elementsRequiringJavascript.show()
 
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,14 +1,6 @@
-// This is a manifest file that'll be compiled into application.js, which will include all the files
-// listed below.
-//
-// Any JavaScript/Coffee file within this directory, lib/assets/javascripts, vendor/assets/javascripts,
-// or vendor/assets/javascripts of plugins, if any, can be referenced here using a relative path.
-//
-// It's not advisable to add code directly here, but if you do, it'll appear at the bottom of the
-// compiled file.
-//
-// Read Sprockets README (https://github.com/sstephenson/sprockets#sprockets-directives) for details
-// about supported directives.
+/* eslint-env browser, jquery */
+/* global GOVUK */
+
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
 //
@@ -20,8 +12,6 @@
 //= require_tree ./modules
 //= require_tree ./components
 
-/* eslint-env jquery */
-
 jQuery(function ($) {
   var $form = $('.js-live-search-form')
 
@@ -30,6 +20,8 @@ jQuery(function ($) {
   var $elementsRequiringJavascript = $('.js-required')
 
   var $atomAutodiscoveryLink = $("link[type='application/atom+xml']").eq('0')
+
+  var $transitionSubscribeButton = document.querySelector('#checklist-email-signup button.js-subscribe-button')
 
   $elementsRequiringJavascript.show()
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js
```diff
@@ -31,6 +32,8 @@ jQuery(function ($) {
 
   var $atomAutodiscoveryLink = $("link[type='application/atom+xml']").eq('0')
 
+  var $transitionSubscribeButton = document.querySelector("button[data-label='subscribe-button']")
```
injms, line 35: Is there only ever going to be one subscribe button on the page?

app/assets/javascripts/email-signup-analytics.js
```diff
@@ -0,0 +1,18 @@
...
+
+(function (Modules) {
+  function EmailSignupAnalytics () {}
+
+  EmailSignupAnalytics.prototype.trackSubscriptions = function (params) {
+    var $button = params.$button
+    var category = params.category
+    var action = params.action
+
+    $button.addEventListener('click', function (e) {
```
injms, line 12: Super picky - you don't need to pass `e` through if you're not going to use it:

```suggestion
    $button.addEventListener('click', function () {
```

app/assets/javascripts/application.js
injms: Instead of using data label to find the button you can pass through a `js-` class in the component to use as a JavaScript hook. https://components.publishing.service.gov.uk/component-guide/button/with_js_classes

chao-xian: Have yet to see a page that has any other. Unfortunately the button component doesn't accept an ID, although I could target the form?

spec/javascripts/modules/email-signup-analytics-spec.js
```diff
@@ -0,0 +1,39 @@
...
+describe('Email subscription button tracking', function () {
+  var GOVUK = window.GOVUK || {}
+  var $element
+
+  beforeEach(function () {
+    GOVUK.analytics = {
+      trackPageview: function () { },
+      trackEvent: function () { },
+      setDimension: function () {}
+    }
```
injms, line 12: :tada:

app/assets/javascripts/application.js
chao-xian: In fact that's what I've done.

spec/javascripts/modules/email-signup-analytics-spec.js
chao-xian: 😅

app/assets/javascripts/application.js
injms: If it's not likely then I wouldn't worry about multiple subscribe buttons on the page until it's requested.

I'd use a `js-subscribe-button-event` class (or something similarly named) to target the element, and pass the class into the component. Then that class can be used with `querySelector()`.

chao-xian:
I've had to resort to adding `// eslint-disable-next-line no-undef` as I'm not able to get `/* eslint-env browser */` to have an effect. It's not clear why this is the case.

app/assets/javascripts/modules/email-signup-analytics.js
```diff
@@ -0,0 +1,23 @@
...
+/* global GOVUK */
+
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function EmailSignupAnalytics () {}
+
+  EmailSignupAnalytics.prototype.trackSubscriptions = function (params) {
+    var $button = params.$button
```
injms, line 11: Oh, for the day when we can use destructuring!

app/assets/javascripts/application.js
```diff
@@ -31,6 +21,8 @@ jQuery(function ($) {
 
   var $atomAutodiscoveryLink = $("link[type='application/atom+xml']").eq('0')
 
+  var $transitionSubscribeButton = document.querySelector('#checklist-email-signup button.js-subscribe-button')
```
injms, line 24: Could we use a `js-` class as a hook for the JavaScript instead? This can be [passed into the button component](https://components.publishing.service.gov.uk/component-guide/button/with_js_classes) in the `class` parameter.

For example:
```suggestion
  var $transitionSubscribeButton = document.querySelector(".js-subscribe-button-tracker")
```

chao-xian:
Closing because I didn't know this was a thing: https://components.publishing.service.gov.uk/component-guide/button/with_data_attributes 😭

chao-xian:
Superseded by https://github.com/alphagov/finder-frontend/pull/2130

PR #2127, opened 2020-06-24
Retire feeds for the coronavirus topical event

sihugh:
The topical event has been removed, as content is now tagged to the coronavirus taxonomy topic (and sub-topics) instead. This allows for greater granularity of notifications, whether that's email or RSS feed.

We've moved email subscriptions over, but we don't want to do that with feeds because:

 - the feeds are probably different because publishers' tagging is imperfect (potentially confusing RSS readers)
 - we want people to sign up for the feed _they_ want, not the feed we give them
 - we also don't want to build up reams of redirect logic.  I intend to delete this _soon_

We should be able to delete the redirect logic for this feed soon, once all RSS readers have had a chance to collect it.

It validates: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Ffinder-front-redirect-a-5m5fgb.herokuapp.com%2Fsearch%2Fall.atom%3Ftopical_events%255B%255D%3Dcoronavirus-covid-19-uk-government-response

https://trello.com/c/ssPOXxya/347-retire-topical-event-rss-feeds-in-finder-frontend

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -26,10 +26,12 @@ class FindersController < ApplicationController
         end
       end
       format.atom do
-        @search_query = initialize_search_query(is_for_feed: true)
         if content_item.is_redirect?
           redirect_to_destination
+        elsif covid_topical_event_feed? # delete me after July 2020
+          show_retired_covid_feed
         else
+          @search_query = initialize_search_query(is_for_feed: true)
           expires_in(ATOM_FEED_MAX_AGE, public: true)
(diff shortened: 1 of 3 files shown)


app/controllers/finders_controller.rb
```diff
@@ -197,4 +199,12 @@ def organisation_registry
   def debug_score?
     params[:debug_score]
   end
+
+  def covid_topical_event_feed
```
reggieb, line 203: Should this method be `covid_topical_event_feed?`

PR #2133, opened 2020-06-26
Redirect coronavirus topical event searches

sihugh:
The coronavirus topical event has been deprecated for a number of weeks, and we have now deleted it. We're aware that there may be some remaining links to search pages using this topical event, and now there are no results appearing there.

So that we can be good citizens of the web, we should redirect those searches to the equivalent page using the coronavirus taxon.

This overwrites other taxons that may be present on the search page, but this approach is better than anything else we can think of.

Other parameters are preserved.

https://trello.com/c/0zBDoJBu/389-redirect-searches-for-the-topical-event-to-the-taxon

---

- https://finder-front-redirect-t-p6kitl.herokuapp.com/search/all?topical_events=coronavirus-covid-19-uk-government-response


app/controllers/redirection_controller.rb
```diff
@@ -2,6 +2,10 @@ class RedirectionController < ApplicationController
   include PublicationsRoutes
   DEFAULT_PUBLICATIONS_PATH = "search/all".freeze
 
+  def covid_topical_event
+    redirect_to(finder_path(params[:slug], params: covid_topic_and_other_params))
+  end
```
reggieb, line 7: Should this action be called `redirect_covid`?

config/routes.rb
```diff
@@ -43,5 +43,10 @@
 
   get "/redirect/statistics/announcements" => "redirection#upcoming_statistics"
 
+  get "/*slug" => "redirection#covid_topical_event", constraints: lambda { |request|
+    topical_events = request.params["topical_events"]
+    topical_events && topical_events.include?("coronavirus-covid-19-uk-government-response")
+  }, as: :redirect_covid
```
reggieb, line 49: If you reorder the elements, I think `do` can be used, which I think makes it a little cleaner:
```ruby
  get "/*slug" => "redirection#covid_topical_event", as: :redirect_covid, constraints: lambda do |request|
    topical_events = request.params["topical_events"]
    topical_events && topical_events.include?("coronavirus-covid-19-uk-government-response")
  end
```
Also if the action is called `redirect_covid` I don't think you need the `:as` declaration

sihugh: using the `lambda do |foo|; end` syntax weirdly gave me 93 errors like this:
```
An error occurred while loading ./spec/validators/param_validator_spec.rb.
Failure/Error: require File.expand_path("../config/environment", __dir__)

FrozenError:
  can't modify frozen Array
# ./config/environment.rb:5:in `<top (required)>'
# ./spec/spec_helper.rb:9:in `<top (required)>'
# ./spec/validators/param_validator_spec.rb:1:in `<top (required)>'
No examples found.
```
They stopped when I wrapped the lambda block in brackets - so I figure the routes parser struggles a bit with it.
I'll stick with the curly braces instead :-)

sihugh: I've taken your other suggestions though, thanks!

reggieb: OK - worth a try. I'm glad my other suggestions were helpful.

PR #2173, opened 2020-08-13
Bug fix for Brexit checker action groupings

hannako:
## What
Fix two bugs on the Brexit checker results page.

### Bug one
A user's selected criteria results in an action that has two grouping criteria. We always show both groups, regardless of the selected criteria. eg A user selects `living-uk`, `visiting-ie` and `visiting-bring-pet`.  Because the action for taking a pet abroad has both `visiting-eu` and `visiting-ie` as grouping criteria, the user is shown the `visiting-ie` group (makes sense) and also the `visiting-eu` group (whoops!). 

- [bug on production](https://www.gov.uk/transition-check/results?c%5B%5D=does-not-own-operate-business-organisation&c%5B%5D=visiting-bring-pet&c%5B%5D=visiting-ie&c%5B%5D=living-uk)

#### New logic:
- If an action has more than one grouping criteria, and a user selects
all of the ANY_OF criteria, we should show duplicate actions under both groups.
- If an action has more than one grouping criteria, and a user selects
some of the ANY_OF criteria, we should only show the action under the grouping that
has a key in common with the user's selected criteria.

- [fixed on review app](https://finder-front-remove-dup-ynsole.herokuapp.com/transition-check/results?c%5B%5D=does-not-own-operate-business-organisation&c%5B%5D=visiting-bring-pet&c%5B%5D=visiting-ie&c%5B%5D=living-uk)

### Bug two
When an action has two grouping criteria, and both criteria have been selected by the user, we display both criteria under each group. Eg a user selects `visiting-bring-pet` `visiting-ie` `visiting-eu` `living-uk`. This gives them the following grouped actions:

**Visiting the EU**
Because you said:

You live in the UK
You plan to travel to Ireland
You plan to travel to the EU
You are travelling with a pet

**Visiting Ireland**
Because you said:

You live in the UK
You plan to travel to Ireland
You plan to travel to the EU
You are travelling with a pet

-[bug on production](https://www.gov.uk/transition-check/results?c%5B%5D=does-not-own-operate-business-organisation&c%5B%5D=visiting-bring-pet&c%5B%5D=visiting-ie&c%5B%5D=visiting-eu&c%5B%5D=living-uk)

#### New logic:
Previously we were creating the list of criteria by combining all of the criteria for all of the actions displayed in a group. Now if an action has multiple grouping criteria, we check if the action's criteria shares a key with the current group. If it does we display the criteria, if it does not we don't.

-[fixed on review app](https://finder-front-remove-dup-ynsole.herokuapp.com/transition-check/results?c%5B%5D=does-not-own-operate-business-organisation&c%5B%5D=visiting-bring-pet&c%5B%5D=visiting-ie&c%5B%5D=visiting-eu&c%5B%5D=living-uk)

https://trello.com/c/MufPWtKB/376-investigate-how-we-can-remove-duplicate-actions-from-the-results-page

### DIFF
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -35,8 +35,8 @@ class BrexitCheckerController < ApplicationController
     @criteria = BrexitChecker::Criterion.load_by(criteria_keys)
     @actions = filter_actions(all_actions, criteria_keys)
     @audience_actions = @actions.group_by(&:audience)
-    @business_results = audience_results.populate_business_groups
-    @citizen_results_groups = audience_results.populate_citizen_groups
+    @business_results = grouped_results.populate_business_groups
+    @citizen_results_groups = grouped_results.populate_citizen_groups
   end
 
   def email_signup; end
@@ -50,8 +50,8 @@ class BrexitCheckerController < ApplicationController
 
 private
 
-  def audience_results
-    @audience_results ||= BrexitChecker::ResultsAudiences.new(@audience_actions, @criteria)
+  def grouped_results
+    @grouped_results ||= BrexitChecker::Results::GroupByAudience.new(@audience_actions, @criteria)
   end
 
   def subscriber_list_options
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -34,9 +34,9 @@ class BrexitCheckerController < ApplicationController
     all_actions = BrexitChecker::Action.load_all
     @criteria = BrexitChecker::Criterion.load_by(criteria_keys)
     @actions = filter_actions(all_actions, criteria_keys)
-    audience_actions = @actions.group_by(&:audience)
-    @business_results = BrexitChecker::ResultsAudiences.populate_business_groups(audience_actions["business"], @criteria)
-    @citizen_results_groups = BrexitChecker::ResultsAudiences.populate_citizen_groups(audience_actions["citizen"], @criteria)
+    @audience_actions = @actions.group_by(&:audience)
+    @business_results = audience_results.populate_business_groups
+    @citizen_results_groups = audience_results.populate_citizen_groups
   end
 
   def email_signup; end
@@ -49,6 +49,10 @@ class BrexitCheckerController < ApplicationController
   end
 
 private
+
+  def audience_results
+    @audience_results ||= BrexitChecker::ResultsAudiences.new(@audience_actions, @criteria)
+  end
 
   def subscriber_list_options
     path = transition_checker_results_path(c: criteria_keys)
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -33,7 +33,7 @@ class BrexitCheckerController < ApplicationController
   def results
     all_actions = BrexitChecker::Action.load_all
     @criteria = BrexitChecker::Criterion.load_by(criteria_keys)
-    @actions = filter_items(all_actions, criteria_keys)
+    @actions = filter_actions(all_actions, criteria_keys)
     audience_actions = @actions.group_by(&:audience)
     @business_results = BrexitChecker::ResultsAudiences.populate_business_groups(audience_actions["business"], @criteria)
     @citizen_results_groups = BrexitChecker::ResultsAudiences.populate_citizen_groups(audience_actions["citizen"], @criteria)
--- a/app/lib/brexit_checker/action.rb
+++ b/app/lib/brexit_checker/action.rb
@@ -62,6 +62,10 @@ class BrexitChecker::Action
     id == other.id
   end
 
+  def multiple_grouping_criteria?
+    grouping_criteria.count > 1
+  end
+
 private
 
   def has_criteria
--- a/app/lib/brexit_checker/group.rb
+++ b/app/lib/brexit_checker/group.rb
@@ -27,6 +27,10 @@ class BrexitChecker::Group
     @load_all ||= YAML.load_file(GROUPS_PATH)["groups"].map { |a| new(a) }
   end
 
+  def self.find_by(key)
+    load_all.detect { |group| group.key == key }
+  end
+
   delegate :hash, to: :key
 
   def eql?(other)
--- a/app/lib/brexit_checker/group.rb
+++ b/app/lib/brexit_checker/group.rb
@@ -15,7 +15,7 @@ class BrexitChecker::Group
                                 studying-uk
                                 common-travel-area]
 
-  attr_reader :key, :heading
+  attr_reader :key, :heading, :priority
 
   def initialize(attrs)
     attrs.each { |key, value| instance_variable_set("@#{key}", value) }
--- a/app/lib/brexit_checker/results_audiences.rb
+++ b/app/lib/brexit_checker/results_audiences.rb
@@ -12,17 +12,36 @@ class BrexitChecker::ResultsAudiences
     def populate_citizen_groups(audience_actions, selected_criteria)
       return [] if audience_actions.blank? || selected_criteria.blank?
 
-      BrexitChecker::Group.load_all.each_with_object([]) do |group, result|
-        selected_actions = group.actions & audience_actions
-        unless selected_actions.empty?
-          result << {
+      audience_actions.each_with_object([]) do |action, grouped_actions|
+        next if action.grouping_criteria.empty?
+
+        group_keys =
+          multiple_grouping_criteria?(action) ? selected_group(action, selected_criteria) : action.grouping_criteria
+        group_keys.each do |key|
+          group = BrexitChecker::Group.find_by(key)
+          next if grouped_actions.any? { |actions| actions[:group] == group }
+          selected_actions = group.actions & audience_actions
+          grouped_actions << {
             group: group,
             actions: selected_actions,
             criteria: selected_actions.flat_map(&:all_criteria).uniq & selected_criteria,
           }
         end
-        result
+        grouped_actions
       end
+    end
+
+    def selected_group(action, selected_criteria)
+      criteria_keys = selected_criteria.map(&:key)
+      action.grouping_criteria.select { |group_key| criteria_keys.include? group_key }
+    end
+
+    def all_groups
+      @all_groups ||= BrexitChecker::Group.load_all
+    end
+
+    def multiple_grouping_criteria?(action)
+      action.grouping_criteria.count > 1
     end
   end
 end
--- a/app/lib/brexit_checker/results_audiences.rb
+++ b/app/lib/brexit_checker/results_audiences.rb
@@ -1,79 +1,84 @@
 class BrexitChecker::ResultsAudiences
-  class << self
-    def populate_business_groups(audience_actions, selected_criteria)
-      return {} if audience_actions.blank? || selected_criteria.blank?
+  attr_reader :citizen_actions, :business_actions, :criteria
 
-      {
-        actions: audience_actions,
-        criteria: audience_actions.flat_map(&:all_criteria).uniq & selected_criteria,
-      }
-    end
+  def initialize(actions, criteria)
+    @citizen_actions = actions["citizen"]
+    @business_actions = actions["business"]
+    @criteria = criteria
+  end
 
-    def populate_citizen_groups(audience_actions, selected_criteria)
-      return [] if audience_actions.blank? || selected_criteria.blank?
+  def populate_business_groups
+    return {} if business_actions.blank? || criteria.blank?
 
-      audience_actions.each_with_object([]) do |action, grouped_actions|
-        next if action.grouping_criteria.empty?
+    {
+      actions: business_actions,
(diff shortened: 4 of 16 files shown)


app/lib/brexit_checker/group.rb
```diff
@@ -15,7 +15,7 @@ class BrexitChecker::Group
                                 studying-uk
                                 common-travel-area]
 
-  attr_reader :key, :heading
```
hannako, line 18: The priority of each group is set in the [spreadsheet](https://docs.google.com/spreadsheets/d/1wIeBTitJVfkWa7oKrGmusIo2r4TsvXVdlne_xG6YjYs/edit#gid=443740486) but I can't see anywhere in the app that the priority is pulled in and used to set the order of the groups. It looks like the order of the elements in the yaml file is all that was previously determining the group order.

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
-        unless selected_actions.empty?
-          result << {
+      audience_actions.each_with_object([]) do |action, grouped_actions|
+        next if action.grouping_criteria.empty?
+
+        group_keys =
+          multiple_grouping_criteria?(action) ? selected_group(action, selected_criteria) : action.grouping_criteria
+        group_keys.each do |key|
+          group = all_groups.select { |g| g.key == key }.first
+          next if grouped_actions.select { |actions| actions[:group] == group }.any?
```
sihugh: you can collapse this to be `next if grouped_actions.any? { |actions| actions[:group] == group }` I think, and it should have the same advantages as `detect` above :-)

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
+
+          grouped_actions << {
             group: group,
-            actions: selected_actions,
-            criteria: selected_actions.flat_map(&:all_criteria).uniq & selected_criteria,
+            actions: actions_for_group(audience_actions, group),
+            criteria: criteria_for_group(audience_actions, group, selected_criteria),
           }
+          grouped_actions.sort_by! { |actions| [(actions[:group].priority * -1), actions[:group].key] }
```
sihugh: Feels like the sorting by priority is significant enough to have a commit of its own? This is quite complicated :-)

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
+    def actions_for_group(audience_actions, group)
+      audience_actions & group.actions
+    end
+
+    def criteria_for_group(audience_actions, group, selected_criteria)
+      actions = audience_actions & group.actions
+      filtered_criteria =
+        actions.map do |action|
+          if multiple_grouping_criteria?(action)
+            rogue_criterion = (action.grouping_criteria - [group.key]).first
```
sihugh: This feels quite magic to me - could you walk me through it? I don't feel confident enough to know whether the `first` criterion is the one we need, or whether they may be others that we're ignoring, or I've missed something completely.

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
+            actions: actions_for_group(audience_actions, group),
+            criteria: criteria_for_group(audience_actions, group, selected_criteria),
           }
+          grouped_actions.sort_by! { |actions| [(actions[:group].priority * -1), actions[:group].key] }
         end
-        result
       end
     end
+
+    def selected_group(action, selected_criteria)
```
sihugh: Is this named correctly? It seems to be more `grouping_criteria` rather than `group`, or something?

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
+    def all_groups
+      @all_groups ||= BrexitChecker::Group.load_all
+    end
+
+    def actions_for_group(audience_actions, group)
+      audience_actions & group.actions
+    end
+
+    def criteria_for_group(audience_actions, group, selected_criteria)
+      actions = audience_actions & group.actions
```
sihugh: you could reuse the `actions_for_group` method you've created a couple of lines up? ;-)

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
+      filtered_criteria =
+        actions.map do |action|
+          if multiple_grouping_criteria?(action)
+            rogue_criterion = (action.grouping_criteria - [group.key]).first
+            (action.all_criteria & selected_criteria).reject { |criteria| criteria.key == rogue_criterion }
+          else
+            action.all_criteria & selected_criteria
```
sihugh: you've used this in a couple of places here. Maybe extract a method, so you could name your intent?

spec/lib/brexit_checker/results_audiences_spec.rb
```diff
@@ -2,81 +2,121 @@
...
-    let(:criteria7) { FactoryBot.build(:brexit_checker_criterion, key: "visiting-ie", text: "Visiting Ireland") }
-
-    let(:group_living_uk) { FactoryBot.build(:brexit_checker_group, key: "living-uk", heading: "Living in the UK") }
-    let(:group_visiting_eu) { FactoryBot.build(:brexit_checker_group, key: "visiting-eu", heading: "Visiting the EU") }
-    let(:group_studying_eu) { FactoryBot.build(:brexit_checker_group, key: "studying-eu", heading: "Studying in the EU") }
-    let(:group_visiting_ie) { FactoryBot.build(:brexit_checker_group, key: "visiting-ie", heading: "Visiting Ireland") }
-
-    let(:selected_criteria) { [criteria1, criteria2, criteria3, criteria4, criteria5] }
-    let(:actions) { [action1, action2, action3, action4, action5, action6] }
+    living_uk = FactoryBot.build(:brexit_checker_criterion, key: "living-uk")
```
sihugh, line 5: Was there a reason why these are no longer `let(:living_uk)` etc.?

app/controllers/brexit_checker_controller.rb
```diff
@@ -50,6 +50,10 @@ def confirm_email_signup
 
 private
 
+  def audience_results
+    BrexitChecker::ResultsAudiences.new(@audience_actions, @criteria)
```
sihugh, line 54: Would this benefit from memoization?

app/lib/brexit_checker/action.rb
```diff
@@ -62,6 +62,10 @@ def eql?(other)
     id == other.id
   end
 
+  def multiple_grouping_criteria?
+    grouping_criteria.count > 1
```
sihugh, line 66: 👍

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -1,65 +1,68 @@
 class BrexitChecker::ResultsAudiences
-  class << self
-    def populate_business_groups(audience_actions, selected_criteria)
-      return {} if audience_actions.blank? || selected_criteria.blank?
+  attr_reader :citizen_actions, :business_actions, :criteria
```
sihugh, line 2: This is a nice thing to do 👍

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -1,65 +1,68 @@
...
-          if multiple_grouping_criteria?(action)
-            rogue_criterion = (action.grouping_criteria - [group.key]).first
-            (action.all_criteria & selected_criteria).reject { |criteria| criteria.key == rogue_criterion }
-          else
-            action.all_criteria & selected_criteria
-          end
-        end
-      filtered_criteria.flatten.uniq
-    end
+  def criteria_keys
```
sihugh, line 51: 👍

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -1,65 +1,68 @@
...
+    @criteria_keys ||= criteria.map(&:key)
+  end
 
-    def multiple_grouping_criteria?(action)
-      action.grouping_criteria.count > 1
-    end
+  def criteria_for_group(group)
```
sihugh, line 55: wonder if it's worth naming this as being specific to citizens?

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -1,65 +1,68 @@
...
+      criteria: business_actions.flat_map(&:all_criteria).uniq & criteria,
+    }
+  end
 
-        group_keys =
-          multiple_grouping_criteria?(action) ? selected_group(action, selected_criteria) : action.grouping_criteria
-        group_keys.each do |key|
-          group = all_groups.select { |g| g.key == key }.first
-          next if grouped_actions.select { |actions| actions[:group] == group }.any?
+  def populate_citizen_groups
```
sihugh, line 19: This method is still pretty complicated. Any chance we could break it down a little?

sihugh (changes requested):
I'm pretty sure you've got the right outcome, but I'm afraid I found the 3rd commit too big for my brain to cope with right now. Could we have a go at splitting it out a bit? Happy to pair on it.

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
 
-      BrexitChecker::Group.load_all.each_with_object([]) do |group, result|
-        selected_actions = group.actions & audience_actions
-        unless selected_actions.empty?
-          result << {
+      audience_actions.each_with_object([]) do |action, grouped_actions|
+        next if action.grouping_criteria.empty?
+
+        group_keys =
+          multiple_grouping_criteria?(action) ? selected_group(action, selected_criteria) : action.grouping_criteria
```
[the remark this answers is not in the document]
hannako: The grouping criteria key on the groups.yaml file stores an array. So even if it only has one grouping_criteria, it's still an array of one element. [eg](https://github.com/alphagov/finder-frontend/blob/314aa1112d0333cf44f947e52b47b98dc712349d/app/lib/brexit_checker/actions.yaml#L21)

app/lib/brexit_checker/results_audiences.rb
```diff
@@ -12,17 +12,54 @@ def populate_business_groups(audience_actions, selected_criteria)
...
+        group_keys.each do |key|
+          group = all_groups.select { |g| g.key == key }.first
```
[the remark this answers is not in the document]
hannako: brilliant thank you!

app/lib/brexit_checker/results_audiences.rb
hannako: thank you!

app/lib/brexit_checker/results_audiences.rb
hannako: thanks 👍

app/lib/brexit_checker/results_audiences.rb
hannako: Currently the highest number of grouping criteria an action can have is 2, so calling `first` was just a way to access the element from the array. But your comment made me think that if in the future an action is given more than 2, this code would have introduced a bug. So I've reworked in a later commit:

```
def filtered_criteria(action, group_key)
  unfiltered = criteria_for_action(action)
  rogue_criteria_keys = (action.grouping_criteria - [group_key])
  unfiltered.reject { |criterion| rogue_criteria_keys.include? criterion.key }
end
```

app/lib/brexit_checker/results_audiences.rb
hannako: I've refactored so that the actions are passed in.  If we add groupings for business actions we can reuse the method, in theory!

app/lib/brexit_checker/results_audiences.rb
hannako: I've renamed `selected_grouping_criteria`, as it's the grouping criteria which have a key in common with the user's selected criteria, rather than all of that action's `grouping_criteria`.

spec/lib/brexit_checker/results_audiences_spec.rb
hannako: Every time factorybot builds the criteria or action via the let, it gets a new id. Because these models aren't active record objects, they are just hashes, if the value of the id changed the test failed. 

Because the criteria and actions are loaded from the yaml more than once, and passed around, even though the data is consistent the ID is changing. Which is maybe bad? Makes it hard to test at the very least. So in a later commit I refactored so that the criteria and actions are passed in at the initialise, so they remain unchanged - I reinstate the `let` blocks at that point.

hannako:
Thank you for the review! I've split the third commits into smaller pieces, and I've extracted the criteria filtering to its own class. Hopefully that makes it a bit clearer.

sihugh (approved):
Looks grand 👍 🎉 🚀 🦸 

Do we need to get Marcus Arborius to re-review?

PR #2184, opened 2020-09-03
Fix facet tag layout when styles disabled

andysellick:
- changing the p and div elements to spans makes the button to remove the facet tag line up better with the text of the facet tag if styles are disabled
- this makes the layout much clearer

Trello card: https://trello.com/c/sMBiJpNG/360-search-results-facet-tags-disjointed

## Visual changes

No changes other than when styles are disabled.

Before | After
------ | ------
Screenshot 2020-09-03 at 12 32 33 | Screenshot 2020-09-03 at 12 30 50

### DIFF
--- a/app/views/finders/_facet_tags.html.erb
+++ b/app/views/finders/_facet_tags.html.erb
@@ -11,9 +11,9 @@
       <div class="facet-tags__group">
         <% applied_filter.each do |filter| %>
           <div class="facet-tags__wrapper">
-            <p class="facet-tags__preposition"><%= filter[:preposition] %></p>
-            <div class="facet-tag">
-              <p class="facet-tag__text"><%= filter[:text] %></p>
+            <span class="facet-tags__preposition"><%= filter[:preposition] %></span>
+            <span class="facet-tag">
+              <span class="facet-tag__text"><%= filter[:text] %></span>
               <button type="button"
                 class="facet-tag__remove"
                 aria-label="Remove filter <%= filter[:text] %>"
@@ -22,7 +22,7 @@
                 data-facet="<%= filter[:data_facet] %>"
                 data-value="<%= filter[:data_value] %>"
                 data-name="<%= filter[:data_name] %>">&#x2715;</button>
-            </div>
+            </span>
           </div>
         <% end %>
       </div>


maxgds (dismissed):

maxgds:
This PR is fine, but doesn't go far enough in solving the accessibility problem. At a glance as a sighted user I see a complete sentence "About <subject x> and <subject y>" but it's still broken up into paragraphs and divs which will make the screenreader experience awkward. Instead of being able to consume the sentence as one block they'll have to navigate through 4:

- About
- subject x
- and
- subject y

It's the same for all the sentences at the top of the search. They need bringing together semantically.

app/views/finders/_facet_tags.html.erb
```diff
@@ -11,9 +11,9 @@
       <div class="facet-tags__group">
         <% applied_filter.each do |filter| %>
           <div class="facet-tags__wrapper">
-            <p class="facet-tags__preposition"><%= filter[:preposition] %></p>
-            <div class="facet-tag">
-              <p class="facet-tag__text"><%= filter[:text] %></p>
+            <span class="facet-tags__preposition"><%= filter[:preposition] %></span>
+            <span class="facet-tag">
+              <span class="facet-tag__text"><%= filter[:text] %></span>
```
maxgds, line 16: This has brought together the first half of the sentence but it still breaks up at the "and" where a new div is introduced:
 About -subject x-
 and -subject y-

maxgds (changes requested):

app/views/finders/_facet_tags.html.erb
andysellick: I decided to do it this way because there's no chance the 'remove' button could be confused with the wrong bit of text. I think it also makes it more readable. Each option is clearly separated by a new line and we don't end up with a very long block of text where line wrapping could lead to confusion.

maxgds: I'd suggest taking it back to Anika in this case, my understanding is that it should be grouped as a whole sentence in the same way for assistive tech that it is visually but I understand your point too.

andysellick: Anika seems happy with this approach. Are you okay to approve?

maxgds: Sure thing.

maxgds (approved):

PR #2192, opened 2020-09-09
Change search region roles

andysellick:

---

## What

Make changes to the `role` attributes on the main search page and other finders to correct an accessibility issue.

Previously, on `/search/all`, the search filters (topic, content type, date etc.) had a role of `search`, which was correct but did not include the search box itself. The page was structured as follows:

- filters had `role="search" aria-label="Search"`
- search results had `role="region" aria-label="Search search results"`

(the repetition in that aria-label is because on other finders the first 'Search' is replaced with the name of the finder, e.g. `News and communications search results`).

The structure of the markup means that it's not possible to wrap all of the search controls in the `role="search"` without also wrapping the results. 

According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role you can have multiple search roles on a page but they need an aria-label to distinguish them. I've therefore changed the roles and labels to be as follows.

### `/search/all`

Screenshot 2020-09-09 at 10 39 41

### Other finders e.g. `/search/news-and-communications`

Screenshot 2020-09-09 at 10 48 57

Note that I've marked the site search in the header as having an updated `aria-label`. This isn't part of this PR but from reading the guidance I believe this is necessary so will do in a [separate PR](https://github.com/alphagov/static/pull/2265).

## Why

`role` attribute was set only around the search filters on the main search and did not include the search box.

Trello card: https://trello.com/c/pv1P0xgu/362-search-results-wrong-use-of-search-aria-role

### DIFF
--- a/app/views/finders/_facet_collection.html.erb
+++ b/app/views/finders/_facet_collection.html.erb
@@ -1,6 +1,6 @@
 <div class="filter-form">
   <% if !content_item.all_content_finder? %>
-    <div id="keywords">
+    <div id="keywords" role="search" aria-label="<%= content_item.title %>">
       <% label_text = capture do %>
         Search <span class="govuk-visually-hidden"><%= content_item.title %></span>
       <% end %>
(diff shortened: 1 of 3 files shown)


injms (approved):
👍

PR #2196, opened 2020-09-10
Use an h2 for the email subscription header

sihugh:
This is presented as a header (bold, uses the govuk-heading-m class), but uses a p tag. It seems semantically better to use an h2. It appears below the page h1, so h3 would be inappropriate. 

This also matches the h2 used on the email callout at the bottom of the page.

An alternative would be to not present this as a heading, but this may require a change to the design, which we don't have time for right now.

There is no visible change.

Screenshot 2020-09-10 at 16 18 18

[Example URL on GOV.UK](https://www.gov.uk/transition-check/results?c%5B%5D=owns-operates-business-organisation&c%5B%5D=visiting-driving&c%5B%5D=visiting-ie&c%5B%5D=visiting-eu&c%5B%5D=visiting-row&c%5B%5D=travel-eu-business-no&c%5B%5D=working-uk&c%5B%5D=living-uk&c%5B%5D=nationality-uk)

https://trello.com/c/EE5x4wxA/444-transition-check-results-heading-looks-like-heading-but-isnt-one

### DIFF
--- a/app/views/components/_email_link.html.erb
+++ b/app/views/components/_email_link.html.erb
@@ -4,7 +4,7 @@
 %>
 <div class="app-c-email-link">
   <% if link_title %>
-    <p class="govuk-heading-m"><%= link_title %></p>
+    <h2 class="govuk-heading-m"><%= link_title %></h2>
   <% end %>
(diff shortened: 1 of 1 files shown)


andysellick (approved):
My only concern would be where else is this used and what effect will this change have? But this seems sensible.

hannako:
@andysellick This component is only used on the brexit checker results page so I think we're safe to merge.

PR #2209, opened 2020-09-24
Transition accessibility fixes

sihugh:
## Make heading-like things actual headings

Each action in the results page starts with a line of text / link that is styled like a heading, but is a `p` tag. The content is designed in a heading style, and it would be useful for them to be listed as headings so that they can be recounted by a voice assistant etc, so let's make them so.

I've used h3s for actions appearing in the business section and h4s for citizen results. It's suboptimal to have a mixture, but pending a redesign (which is actually happening next week), I'm not sure what the alternative is, while honouring the page structure.

## Remove aria-label from Brexit results asides

The labels are not sensible at the moment - they are supposed to be readable language, not kebab-case slugs.

This also rejigs the markup a little to extract the section heading from the aside. The heading refers to both the aside, and any actions that are presented for this group. It makes sense if it is not encapsulated within the aside.

[Example results page](https://finder-front-a11y-label-mjeoku.herokuapp.com/transition-check/results?c%5B%5D=employ-eu-citizens&c%5B%5D=owns-operates-business-organisation-uk&c%5B%5D=owns-operates-business-organisation&c%5B%5D=visiting-ie&c%5B%5D=visiting-eu&c%5B%5D=travel-eu-business&c%5B%5D=working-uk&c%5B%5D=working-ie&c%5B%5D=living-uk&c%5B%5D=nationality-uk)

https://trello.com/c/DtragFBC/473-transition-check-results-unintelligible-labels

### DIFF
--- a/app/views/brexit_checker/_action_list.html.erb
+++ b/app/views/brexit_checker/_action_list.html.erb
@@ -1,7 +1,7 @@
 <% actions.each.with_index do | action, index | %>
   <div class="govuk-grid-row brexit-checker__action">
     <div class="brexit-checker__content-wrapper">
-      <p class="govuk-body brexit-checker__action_title">
+      <h4 class="govuk-body brexit-checker__action_title">
         <% if action.title_url.present? %>
           <a
             class="govuk-link"
@@ -16,7 +16,7 @@
         <% else %>
           <%= action.title %>
         <% end %>
-      </p>
+      </h4>
       <% if action.lead_time.present? %>
         <div class="govuk-body">
           <%= render "govuk_publishing_components/components/inset_text", text: action.lead_time %>
--- a/app/views/brexit_checker/_action_list.html.erb
+++ b/app/views/brexit_checker/_action_list.html.erb
@@ -1,7 +1,10 @@
+<% heading_level ||= 4 %>
+<% heading_tag = "h#{heading_level}"%>
+
 <% actions.each.with_index do | action, index | %>
   <div class="govuk-grid-row brexit-checker__action">
     <div class="brexit-checker__content-wrapper">
-      <h4 class="govuk-body brexit-checker__action_title">
+      <%= content_tag(heading_tag, class: "govuk-body brexit-checker__action_title") do %>
         <% if action.title_url.present? %>
           <a
             class="govuk-link"
@@ -16,7 +19,7 @@
         <% else %>
           <%= action.title %>
         <% end %>
-      </h4>
+      <% end %>
       <% if action.lead_time.present? %>
         <div class="govuk-body">
           <%= render "govuk_publishing_components/components/inset_text", text: action.lead_time %>
--- a/app/views/brexit_checker/_results_criteria.html.erb
+++ b/app/views/brexit_checker/_results_criteria.html.erb
@@ -1,22 +1,21 @@
 <% if criteria.present? %>
-  <aside
-    class="govuk-grid-column-one-third"
-    aria-label=<%="#{criteria_aside_label(local_assigns[:heading])}"%>
-  >
-    <% if local_assigns[:heading] %>
-      <h3 class="govuk-heading-m"><%= heading %></h3>
-    <% end %>
-    <p class="govuk-body brexit-checker__critera_explaination">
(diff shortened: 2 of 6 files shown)


app/views/brexit_checker/_results_criteria.html.erb
```diff
@@ -1,22 +1,21 @@
...
-          <%= criterion.text %>
-        </li>
-      <% end %>
-    </ul>
-  </aside>
+  <div class="govuk-grid-column-one-third">
+  <% if local_assigns[:heading] %>
+    <h3 class="govuk-heading-m"><%= heading %></h3>
+  <% end %>
+    <aside>
```
andysellick, line 6: If this is an `aside`, shouldn't it include the heading?

sihugh: I think the heading belongs outside the aside, because it refers to the actions in the section as well as the aside bit:

Screenshot 2020-09-24 at 15 35 30

So "Visiting Ireland" is the h3, the "Because you said" bit is the aside, and the "Check whether your mobile phone" bit is subordinate to the h3.

So "Visiting Ireland" is key to the page and the aside is assistive. I thought my way would be more appropriate. Though I may be wrong :-D

andysellick: That provides a bit more context, thanks. Genuinely I don't know, now.

app/views/brexit_checker/_action_list.html.erb
```diff
@@ -1,7 +1,7 @@
 <% actions.each.with_index do | action, index | %>
   <div class="govuk-grid-row brexit-checker__action">
     <div class="brexit-checker__content-wrapper">
-      <p class="govuk-body brexit-checker__action_title">
+      <h4 class="govuk-body brexit-checker__action_title">
```
andysellick, line 4: Looks like this should be a H3? (I think it's this one, there's definitely a H4 somewhere in this page that should be a H3).

Screenshot 2020-09-24 at 16 05 33

andysellick: (ignore the H2/cookies error at the top, that's actually okay)

sihugh: Ah, that's a snag. The business section doesn't get the h3 title that the other bits get 😞 . I'll have to have a wrestle with that. I don't really want to promote all the business actions to be h3s, though cos I reckon all the actions should remain at the same level. 
Humph

sihugh: And boo

andysellick: Sorry...

app/views/brexit_checker/_results_criteria.html.erb
hannako: I think moving the heading out of the aside makes sense :+1:

Q: Is it fine that we are removing the aria label, as opposed to amending it?

sihugh: I don't know what it was supposed to add, really?

app/views/brexit_checker/_action_list.html.erb
sihugh: OK, I've raised this with product/content to hopefully get an appropriate h3 for the business section.

sihugh: I've added a commit to use h3s on business results and h4s on citizen results. The commit has more details about the longer term options which would be better.

app/views/brexit_checker/_action_list.html.erb
```diff
@@ -16,7 +19,7 @@
         <% else %>
           <%= action.title %>
         <% end %>
-      </h4>
+      </<%= heading_tag %>>
```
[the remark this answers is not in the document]
sihugh: I agree this is neater. I've noticed that [this is referred to as legacy](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag) though, so at some point we might have to find another way. 
> "Note: this is legacy syntax, see tag method description for details."
We've got this in a few places, all over, so I guess we'll have to tackle them then.

andysellick (approved):

PR #2211, opened 2020-09-25
Move Brexit breadcrumbs so skiplinks skip them

sihugh:
I've moved the breadcrumbs/back links on the Brexit checker so that they live outside the main content section. This means that using skiplinks navigates you to the start of the content, and not just to a confusing breadcrumb link.

Check the skip links on:
- [a question page](https://finder-front-a11y-fix-f-d1ljfw.herokuapp.com/transition-check/questions)
- [a results page](https://finder-front-a11y-fix-f-d1ljfw.herokuapp.com/transition-check/results?c%5B%5D=nationality-uk&page=1)

They should now link to the main content section, and skip over the breadcrumbs.

I've also updated a link to /get-ready-brexit-check so that it points to /transition. The old page has been retired, redirecting to the transition landing page, so this change eliminates that redirected request, while not changing the outcome.

Describe the problem
--------------------
The main element of the transition check flow includes the back link and the skip link leads to the back link.
The same is happening on the transition check results page but with the breadcrumbs.

Problem category
----------------
Fail of WCAG SC 2.4.1

Impact of the problem
---------------------
Skip link: Keyboard only users cannot skip directly to the main content.
Main element: Screen reader users cannot skip directly to the main content.

https://trello.com/c/meGNv7bk/474-search-results-main-element-and-skip-link-before-main-content
https://trello.com/c/MR7bfTid/475-transition-check-main-element-and-skip-link-before-main-content

### DIFF
--- a/app/views/brexit_checker/results.html.erb
+++ b/app/views/brexit_checker/results.html.erb
@@ -19,21 +19,21 @@
 <% end %>
 
 <% content_for :breadcrumbs do %>
-<div class="govuk-width-container">
-  <%= render 'govuk_publishing_components/components/breadcrumbs', {
-    collapse_on_mobile: true,
-    breadcrumbs: [
-      {
-        title: t('brexit_checker.breadcrumbs.home'),
-        url: "/"
-      },
-      {
-        title: t('brexit_checker.breadcrumbs.brexit-home'),
-        url: "/transition"
-      }
-    ]
-  } %>
-</div>
+  <div class="govuk-width-container">
+    <%= render 'govuk_publishing_components/components/breadcrumbs', {
+      collapse_on_mobile: true,
+      breadcrumbs: [
+        {
+          title: t('brexit_checker.breadcrumbs.home'),
+          url: "/"
+        },
+        {
+          title: t('brexit_checker.breadcrumbs.brexit-home'),
+          url: "/transition"
+        }
+      ]
+    } %>
+  </div>
 <% end %>
 
 <div class="govuk-width-container brexit-checker-results-page">
--- a/app/views/brexit_checker/results.html.erb
+++ b/app/views/brexit_checker/results.html.erb
@@ -18,7 +18,8 @@
   <link rel="canonical" href="<%= page_url %>">
 <% end %>
 
-<div class="govuk-width-container brexit-checker-results-page">
+<% content_for :breadcrumbs do %>
+<div class="govuk-width-container">
   <%= render 'govuk_publishing_components/components/breadcrumbs', {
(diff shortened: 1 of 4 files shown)


app/views/brexit_checker/results.html.erb
```diff
@@ -18,7 +18,8 @@
   <link rel="canonical" href="<%= page_url %>">
 <% end %>
 
-<div class="govuk-width-container brexit-checker-results-page">
+<% content_for :breadcrumbs do %>
+<div class="govuk-width-container">
```
danacotoran, line 22: Nitpick: we would normally indent the markup inside a `content_for` block.

app/views/brexit_checker/show.html.erb
```diff
@@ -12,17 +12,21 @@
   end
 %>
 
+<% content_for :breadcrumbs do %>
 <div class="govuk-width-container">
```
danacotoran, line 16: Same comment regarding indentation inside the `content_for` block.

app/views/brexit_checker/show.html.erb
```diff
@@ -12,17 +12,21 @@
...
   <% if @previous_page %>
     <%= render "govuk_publishing_components/components/back_link", {
       href: transition_checker_questions_path(page: @previous_page, c: criteria_keys),
     } %>
   <% else %>
     <%= render "govuk_publishing_components/components/back_link", {
-      href: "/get-ready-brexit-check",
+      href: "/transition",
```
danacotoran, line 23: Besides fixing the skip link issue, this PR also seems to change the back link href from `/get-ready-brexit-check` to `/transition`.
Having tested I understand why it's been done. But could we please also update the PR description and commit message to mention this change and explain why it is happening? 🙏

danacotoran (approved):
Thank you for fixing this WCAG fail 🙏 
Just left a few non-blocking comments. Otherwise LGTM 👍

app/views/brexit_checker/results.html.erb
sihugh: Thanks, have added that in the next commit, so that it's easier to see what has changed here

app/views/brexit_checker/show.html.erb
sihugh: Thanks, have added that in the next commit, so that it's easier to see what has changed here

app/views/brexit_checker/show.html.erb
sihugh: Have updated the commit message and PR - thanks for the spot :-)

PR #2214, opened 2020-09-28
Allow logging into the Transition Checker using a GOV.UK Account

brucebolt:
Allows logging into the Transition Checker using a GOV.UK Account and retrieval of previous results.

All the functionality here is currently hidden from users behind a feature flag, therefore there will be no public changes until we enable this flag.

This adds a number of routes:
- `/transition-check/login`
- `/transition-check/login/callback`
- `/transition-check/logout`
- `/transition-check/saved-results`
- `/transition-check/edit-saved-results`

All the routes should be self explanatory, apart from the final two.

The `saved-results` route conducts some logic:
- User logged in and has checker results: redirects to the results page with their responses pre-populated as part of the query string
- User logged in and has no checker results: redirects to first question in checker
- User logged out: redirects to login route, then one of the above two scenarios after login

The `edit-saved-results` route is similar:
- User logged in and has checker results: redirects to the first question with their responses pre-populated as part of the query string
- User logged in and has no checker results: redirects to first question in checker
- User logged out: redirects to login route, then one of the above two scenarios after login

Trello card: https://trello.com/c/F2pCU73B

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -34,6 +34,12 @@ private
     end
   end
 
+  def logout!
+    session.delete(:sub)
+    session.delete(:access_token)
+    session.delete(:refresh_token)
+  end
+
   def error_503(exception)
     error(503, exception)
   end
@@ -84,5 +90,13 @@ private
 
       ParamsCleaner.new(permitted_params).cleaned
     end
+  end
+
+  def logged_in?
+    current_user.present?
+  end
+
+  def current_user
+    session[:sub]
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,6 +8,8 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   helper :application
 
+  after_action :skip_session_cookie
+
   # rescue_from precedence is bottom up - https://stackoverflow.com/a/9121054/170864
   unless Rails.env.development?
     rescue_from GdsApi::BaseError, with: :error_503
@@ -25,6 +27,12 @@ class ApplicationController < ActionController::Base
   end
 
 private
+
+  def skip_session_cookie
+    unless current_user
+      request.session_options[:drop] = true
+    end
+  end
 
   def error_503(exception)
     error(503, exception)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -92,6 +92,14 @@ private
     end
   end
 
+  def oidc
+    @oidc ||= OidcClient.new(
+      Services.accounts_api,
+      ENV.fetch("GOVUK_ACCOUNT_OAUTH_CLIENT_ID"),
+      ENV.fetch("GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET"),
+    )
+  end
+
   def logged_in?
     current_user.present?
   end
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -13,7 +13,7 @@ class BrexitCheckerController < ApplicationController
     expires_in(30.minutes, public: true) unless Rails.env.development?
   end
 
-  before_action :check_accounts_enabled, only: %i[save_results saved_results]
+  before_action :check_accounts_enabled, only: %i[save_results saved_results edit_saved_results]
 
   def show
     all_questions = BrexitChecker::Question.load_all
@@ -69,6 +69,16 @@ class BrexitCheckerController < ApplicationController
       redirect_to transition_checker_questions_path
     else
       redirect_to transition_checker_results_path(c: results_from_account)
+    end
+  end
+
+  def edit_saved_results
+    redirect_to transition_checker_new_session_path(redirect_path: transition_checker_edit_saved_results_path) and return unless logged_in?
+
+    if results_from_account.empty?
+      redirect_to transition_checker_questions_path
+    else
+      redirect_to transition_checker_questions_path(c: results_from_account, page: 0)
     end
   end
 
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -60,6 +60,35 @@ class BrexitCheckerController < ApplicationController
 
   def save_results; end
 
+  def saved_results
+    redirect_to transition_checker_new_session_path and return unless logged_in?
+
+    @saved_results =
+      if logged_in?
+        begin
+          Array(
+            update_session_tokens(
+              oidc.get_checker_attribute(
+                access_token: session[:access_token],
+                refresh_token: session[:refresh_token],
+              ),
+            ),
+          )
+        rescue OidcClient::OAuthFailure
+          # this means the refresh token has been revoked or the
+          # accounts services are down
+          logout!
+          []
+        end
+      end
+
+    if @saved_results.empty?
+      redirect_to transition_checker_questions_path
+    else
+      redirect_to transition_checker_results_path(c: @saved_results["criteria_keys"])
+    end
+  end
+
 private
 
   def grouped_results
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -13,13 +13,15 @@ class BrexitCheckerController < ApplicationController
     expires_in(30.minutes, public: true) unless Rails.env.development?
   end
 
-  before_action :check_accounts_enabled, only: [:save_results]
+  before_action :check_accounts_enabled, only: %i[save_results saved_results]
 
   def show
-    @account_information = if logged_in?
+    @account_information = if logged_in? && accounts_enabled?
                              "Logged in. <a class=\"govuk-link\" href=\"#{transition_checker_end_session_path}\">Log out.</a>"
-                           else
+                           elsif accounts_enabled?
                              "Not logged in. <a class=\"govuk-link\" href=\"#{transition_checker_new_session_path}\">Login.</a>"
+                           else
+                             ""
                            end
 
     all_questions = BrexitChecker::Question.load_all
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -1,3 +1,5 @@
+require_relative "../lib/oidc_client.rb"
+
 class BrexitCheckerController < ApplicationController
   include BrexitCheckerHelper
 
@@ -14,6 +16,12 @@ class BrexitCheckerController < ApplicationController
   before_action :check_accounts_enabled, only: [:save_results]
 
   def show
+    @account_information = if logged_in?
+                             "Logged in. <a class=\"govuk-link\" href=\"#{transition_checker_end_session_path}\">Log out.</a>"
+                           else
+                             "Not logged in. <a class=\"govuk-link\" href=\"#{transition_checker_new_session_path}\">Login.</a>"
+                           end
+
     all_questions = BrexitChecker::Question.load_all
     @question_index = next_question_index(
       all_questions: all_questions,
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -65,31 +65,32 @@ class BrexitCheckerController < ApplicationController
   def saved_results
     redirect_to transition_checker_new_session_path(redirect_path: transition_checker_saved_results_path) and return unless logged_in?
 
-    @saved_results =
-      begin
-        Array(
-          update_session_tokens(
-            oidc.get_checker_attribute(
-              access_token: session[:access_token],
-              refresh_token: session[:refresh_token],
-            ),
-          ),
-        )
-      rescue OidcClient::OAuthFailure
-        # this means the refresh token has been revoked or the
-        # accounts services are down
-        logout!
-        []
-      end
-
-    if @saved_results.empty?
+    if results_from_account.empty?
       redirect_to transition_checker_questions_path
     else
-      redirect_to transition_checker_results_path(c: @saved_results)
+      redirect_to transition_checker_results_path(c: results_from_account)
     end
   end
 
 private
+
+  def results_from_account
+    @results_from_account ||= begin
+      Array(
+        update_session_tokens(
+          oidc.get_checker_attribute(
+            access_token: session[:access_token],
+            refresh_token: session[:refresh_token],
+          ),
+        ),
+      )
+                              rescue OidcClient::OAuthFailure
+                                # this means the refresh token has been revoked or the
+                                # accounts services are down
+                                logout!
+                                []
+    end
+  end
 
   def grouped_results
     @grouped_results ||= BrexitChecker::Results::GroupByAudience.new(@audience_actions, @criteria)
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -61,31 +61,29 @@ class BrexitCheckerController < ApplicationController
   def save_results; end
 
   def saved_results
-    redirect_to transition_checker_new_session_path and return unless logged_in?
+    redirect_to transition_checker_new_session_path(redirect_path: transition_checker_saved_results_path) and return unless logged_in?
 
     @saved_results =
-      if logged_in?
-        begin
-          Array(
-            update_session_tokens(
-              oidc.get_checker_attribute(
-                access_token: session[:access_token],
-                refresh_token: session[:refresh_token],
-              ),
+      begin
+        Array(
+          update_session_tokens(
+            oidc.get_checker_attribute(
+              access_token: session[:access_token],
+              refresh_token: session[:refresh_token],
             ),
-          )
-        rescue OidcClient::OAuthFailure
-          # this means the refresh token has been revoked or the
-          # accounts services are down
-          logout!
-          []
-        end
+          ),
+        )
+      rescue OidcClient::OAuthFailure
+        # this means the refresh token has been revoked or the
+        # accounts services are down
+        logout!
+        []
       end
 
     if @saved_results.empty?
       redirect_to transition_checker_questions_path
     else
-      redirect_to transition_checker_results_path(c: @saved_results["criteria_keys"])
+      redirect_to transition_checker_results_path(c: @saved_results)
     end
   end
 
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -16,14 +16,6 @@ class BrexitCheckerController < ApplicationController
   before_action :check_accounts_enabled, only: %i[save_results saved_results]
 
   def show
-    @account_information = if logged_in? && accounts_enabled?
-                             "Logged in. <a class=\"govuk-link\" href=\"#{transition_checker_end_session_path}\">Log out.</a>"
-                           elsif accounts_enabled?
-                             "Not logged in. <a class=\"govuk-link\" href=\"#{transition_checker_new_session_path}\">Login.</a>"
-                           else
-                             ""
-                           end
-
     all_questions = BrexitChecker::Question.load_all
     @question_index = next_question_index(
       all_questions: all_questions,
@@ -43,6 +35,14 @@ class BrexitCheckerController < ApplicationController
   end
 
   def results
+    @account_information = if logged_in? && accounts_enabled?
+                             "Logged in. <a class=\"govuk-link\" href=\"#{transition_checker_end_session_path}\">Log out.</a>"
+                           elsif accounts_enabled?
+                             "Not logged in. <a class=\"govuk-link\" href=\"#{transition_checker_new_session_path}\">Login.</a>"
+                           else
+                             ""
+                           end
+
     all_actions = BrexitChecker::Action.load_all
     @criteria = BrexitChecker::Criterion.load_by(criteria_keys)
     @actions = filter_actions(all_actions, criteria_keys)
--- a/app/controllers/brexit_checker_controller.rb
+++ b/app/controllers/brexit_checker_controller.rb
@@ -95,6 +95,12 @@ private
     @grouped_results ||= BrexitChecker::Results::GroupByAudience.new(@audience_actions, @criteria)
   end
 
+  def update_session_tokens(result)
+    session[:access_token] = result[:access_token]
+    session[:refresh_token] = result[:refresh_token]
+    result[:result]
+  end
+
   def subscriber_list_options
     path = transition_checker_results_path(c: criteria_keys)
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,0 +1,51 @@
+require_relative "../lib/oidc_client.rb"
+
+class SessionsController < ApplicationController
+  def create
+    if logged_in?
+      redirect_to redirect_path
+      return
+    end
+
+    redirect_to oidc.auth_uri[:uri]
+  end
+
+  def callback
+    unless params[:code]
+      redirect_to redirect_path
+      return
+    end
+
+    state = params.require(:state)
+
+    user_info = oidc.callback(
+      params.require(:code),
+      state,
+    )
+
+    session[:sub] = user_info.sub
+    session[:email] = user_info.email
+
+    redirect_to redirect_path
+  end
+
+  def delete
+    session.delete(:sub)
+    session.delete(:email)
+    redirect_to redirect_path
+  end
+
+private
+
+  def redirect_path
+    transition_checker_questions_path
+  end
+
+  def oidc
+    @oidc ||= OidcClient.new(
+      Services.accounts_api,
+      ENV.fetch("GOVUK_ACCOUNT_OAUTH_CLIENT_ID"),
+      ENV.fetch("GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET"),
+    )
+  end
+end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -3,16 +3,20 @@ require_relative "../lib/oidc_client.rb"
 class SessionsController < ApplicationController
   def create
     if logged_in?
-      redirect_to redirect_path
+      redirect_to default_redirect_path
       return
     end
 
-    redirect_to oidc.auth_uri[:uri]
+    if Rails.env.test?
+      render plain: "Redirecting to login"
+    else
+      redirect_to oidc.auth_uri(redirect_path: params["redirect_path"])[:uri]
+    end
   end
 
   def callback
     unless params[:code]
-      redirect_to redirect_path
+      redirect_to default_redirect_path
       return
     end
 
@@ -25,6 +29,7 @@ class SessionsController < ApplicationController
 
     access_token = callback[:access_token]
     sub = callback[:sub]
+    redirect_path = callback[:redirect_path] || default_redirect_path
 
     session[:sub] = sub
     session[:access_token] = access_token.token_response[:access_token]
@@ -35,12 +40,12 @@ class SessionsController < ApplicationController
 
   def delete
     logout!
-    redirect_to redirect_path
+    redirect_to default_redirect_path
   end
 
 private
 
-  def redirect_path
+  def default_redirect_path
     transition_checker_questions_path
   end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -18,13 +18,16 @@ class SessionsController < ApplicationController
 
     state = params.require(:state)
 
-    user_info = oidc.callback(
+    callback = oidc.callback(
       params.require(:code),
       state,
     )
 
-    session[:sub] = user_info.sub
-    session[:email] = user_info.email
+    access_token = callback[:access_token]
+    sub = callback[:sub]
+
+    session[:sub] = sub
+    session[:access_token] = access_token.token_response[:access_token]
 
     redirect_to redirect_path
   end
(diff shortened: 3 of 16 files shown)


barrucadu (commented):
Looks ok to me, but I'm not convinced merging this as-is will be more helpful than implementing from scratch based on the designs we have (linked from the trello card), as we know this PR isn't implementing what we actually want to see.  What do you think?

brucebolt:
My thought was probably to merge this (so at least we get the logging in behaviour available), then iterate based on the designs once they are more finalised.  Hopefully it should just be moving around the login banner, updating the user flow and improving the call to actions.

app/controllers/brexit_checker_controller.rb
```diff
@@ -33,6 +43,39 @@ def show
...
+            ),
+          )
+        end
+      rescue OidcClient::OAuthFailure
+        # this means the refresh token has been revoked or the
+        # accounts services are down
+        logout!
+      end
+    end
+
```
barrucadu, line 78: This can be removed, it's been superseded by the JWT flow in #2207

app/controllers/brexit_checker_controller.rb
```diff
@@ -52,12 +95,47 @@ def confirm_email_signup
...
+        end
+      end
+
+    if @saved_results.empty?
+      redirect_to transition_checker_questions_path
+    else
+      redirect_to transition_checker_results_path(c: @saved_results.last["criteria_keys"])
+    end
+  end
+
```
barrucadu, line 126: We're only storing one result for now, see https://github.com/alphagov/finder-frontend/pull/2207/files#diff-e4257d6c3bb3ff33ad32620a2d04b871R28

app/controllers/brexit_checker_controller.rb
```diff
@@ -11,9 +13,17 @@ class BrexitCheckerController < ApplicationController
...
 
   def show
+    @account_information = if logged_in? && accounts_enabled?
+                             "Logged in. <a class=\"govuk-link\" href=\"#{transition_checker_end_session_path}\">Log out.</a><br><a class=\"govuk-link\" href=\"#{transition_checker_saved_results_path}\">See previous results.</a>"
+                           elsif accounts_enabled?
+                             "Not logged in. <a class=\"govuk-link\" href=\"#{transition_checker_new_session_path}\">Login.</a>"
+                           else
+                             ""
+                           end
+
```
barrucadu, line 26: Rather than something on the question pages for this, we'll want to change the call-to-action on the results page... I'm not sure if we've agreed what that should look like yet, so maybe for now just say "You are already logged in" above the green button?

app/controllers/brexit_checker_controller.rb
brucebolt: Ah, yes, of course.  I've now removed this code and the results no longer get saved to the account automatically.

app/controllers/brexit_checker_controller.rb
brucebolt: Done.  Kept the plain text for now, which can get updated to match the designs as part of a frontend card.

app/controllers/brexit_checker_controller.rb
brucebolt: Great, I've updated to remove the expectation this will be a hash of arrays.

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,131 @@
...
+      if ENV["VCAP_APPLICATION"].present?
+        "https://#{JSON.parse(ENV['VCAP_APPLICATION'])['uris'].first}"
+      else
+        "http://#{ENV['VIRTUAL_HOST']}"
+      end
+    host + Rails.application.routes.url_helpers.transition_checker_new_session_callback_path
+  end
+
+  def scopes
+    %i[email transition_checker]
```
barrucadu, line 44: I don't think we need the `email` scope here (and we don't need it in the session either)

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,131 @@
...
+    }
+  end
+
+  def redirect_uri
+    host =
+      if ENV["VCAP_APPLICATION"].present?
+        "https://#{JSON.parse(ENV['VCAP_APPLICATION'])['uris'].first}"
+      else
+        "http://#{ENV['VIRTUAL_HOST']}"
+      end
```
barrucadu, line 39: How does this fit into our normal app runtime environment?  Is the `VIRTUAL_HOST` env var set in AWS?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    if logged_in?
+      redirect_to redirect_path
+      return
+    end
+
+    if Rails.env.test?
+      render plain: "Redirecting to login"
+    else
+      redirect_to oidc.auth_uri[:uri]
+    end
```
barrucadu, line 14: We just do the redirect and check that happens in the tests

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def callback(code, state)
+    client.authorization_code = code
+    access_token = client.access_token!
+    id_token = OpenIDConnect::ResponseObject::IdToken.decode access_token.id_token, discover.jwks
+    id_token.verify! client_id: client_id, issuer: discover.issuer, nonce: state
+    {
+      access_token: access_token,
+      user_info: access_token.userinfo!,
```
barrucadu, line 54: I think the `id_token` has the `sub`, so if we remove the `email` from the session I think we can remove this call to the userinfo endpoint, which would be nice

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def update_checker_attribute(access_token:, refresh_token:, data:)
+    oauth_request(
+      access_token: access_token,
+      refresh_token: refresh_token,
+      method: :put,
+      arg: { "value" => data.to_json },
+    ).except(:result)
+  end
```
barrucadu, line 79: This can be removed, so the `method` and `arg` parameters of `oauth_request` can be inlined

config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-FinderFrontend::Application.config.session_store :disabled
+FinderFrontend::Application.config.session_store :cookie_store
```
barrucadu, line 3: Just as a sanity check, Rails isn't going to start setting any cookies unless people start visiting the login URL, right?

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -0,0 +1,263 @@
...
+  end
+
+  context "with accounts enabled" do
+    let(:attribute_service_url) { "http://attribute-service" }
+
+    before do
+      ENV["OIDC_CLIENT_ID"] = "id"
+      ENV["OIDC_CLIENT_SECRET"] = "secret"
+
+      allow(Services).to receive(:accounts_api).and_return(Plek.find("account-manager"))
```
barrucadu, line 50: Isn't this just returning what `Services.acounts_api` will already produce?

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -0,0 +1,263 @@
...
+    let(:mock_results) { %w[nationality-eu] }
+
+    context "the user is not logged in" do
+      context "/transition-check/questions" do
+        it "does not show the login state" do
+          given_im_on_a_question_page
+          expect(page).to_not have_content("Logged in.")
+          expect(page).to_not have_content("Not logged in.")
+        end
+      end
```
barrucadu, line 66: I don't think we need this test

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -0,0 +1,263 @@
...
+
+      let(:transition_checker_state) { %w[nationality-uk] }
+
+      context "/transition-check/questions" do
+        it "does not show the login state" do
+          given_im_on_a_question_page
+          expect(page).to_not have_content("Logged in.")
+          expect(page).to_not have_content("Not logged in.")
+        end
+      end
```
barrucadu, line 106: Or this test

app/controllers/brexit_checker_controller.rb
```diff
@@ -52,12 +62,47 @@ def confirm_email_signup
 
   def save_results; end
 
+  def saved_results
+    redirect_to transition_checker_new_session_path and return unless logged_in?
```
barrucadu, line 66: Should visiting this page when not logged in take you through the login flow and then redirect back to the results page, rather than the questions page?

barrucadu: If it did the account manager could just link to `www.gov.uk/transition-check/saved-results` and not need to worry about constructing a querystring.

app/lib/oidc_client.rb
brucebolt: Yes, looks like we don't need it since email signups will be handled in the account manager flow.

app/controllers/sessions_controller.rb
brucebolt: I was having some trouble getting this to run in the tests, possibly around the redirect to an external URL.  I'll see if I can reproduce this to come up with a more elegant solution.

config/initializers/session_store.rb
brucebolt: It looks like we set a `_session_id` cookie.  We probably don't want this set unless the user logs in.

Adding this to the `ApplicationController` seems to do the trick of only enabling this cookie when the user is logged in:
```
after_action :skip_session_cookie

...

  def skip_session_cookie
    unless current_user
      request.session_options[:drop] = true
    end
  end
```

spec/features/brexit_checker/accounts_spec.rb
brucebolt: Good spot!  I had reordered and rebased some commits but didn't spot this wouldn't be needed anymore.

spec/features/brexit_checker/accounts_spec.rb
brucebolt: I'll remove this test

spec/features/brexit_checker/accounts_spec.rb
brucebolt: And this one too

app/lib/oidc_client.rb
brucebolt: Looks like `VIRTUAL_HOST` is something we set in the `docker-compose.yml` for `finder-frontend` in govuk-docker, but not in any of the AWS environments.  I stole this code from an app deployed to the PaaS (barking permit, I think), so I'll update this to use something we have available in the AWS environment.

brucebolt: I think `GOVUK_WEBSITE_ROOT` is the env var I'm looking for.

barrucadu: :+1: I don't think we need `VCAP_APPLICATION` as that was just useful for testing things on the PaaS, but we're now getting closer to running it for real.  So `ENV["VIRTUAL_HOST"] || ENV["GOVUK_WEBSITE_ROOT"]` is probably what we want.

app/controllers/sessions_controller.rb
brucebolt: I remember now.  Capybara ignores the root of the redirect, so a request to (say) `http://www.login.service.dev.gov.uk/login` will actually get redirected to `http://finder-frontend.dev.gov.uk/login`, which then results in a query being made to `http://content-store.dev.gov.uk/content/login` (because of the "catch-all" wildcard in routes).  I found a little bit of explanation in https://makandracards.com/makandra/15217-test-redirects-to-an-external-url-with-cucumber-capybara, but could't come up with a more elegant solution.

app/lib/oidc_client.rb
brucebolt: Yeah, that's a good idea.  I'll remove the extra `user_info` stuff.

app/controllers/brexit_checker_controller.rb
brucebolt: I've condensed this into the `/login` flow, so after logging in, the user will be taken to either the questions or results (depending on whether they have anything saved).  In the account manager's "service card", we can just link to `/login`.

barrucadu: Screenshot 2020-10-07 at 13 48 24

I think we do want to be able to link to both independently

brucebolt: I wonder if we'll need to add another `redirect_uri` for the client in the account manager then?  Since we've only got the `/login/callback` route at the moment, and this doesn't have any knowledge of the task you're trying to complete.

barrucadu: I think we can use the `state` parameter we pass as part of the OAuth call for this.  The server passes it back unchanged.  Currently we're just using it to store the nonce, but it could be something like `{nonce}:q` or `{nonce}:r`

brucebolt: Thanks for the pointer @barrucadu.  I've used the `state` to determine where to send the user after the callback.  Also added a `/edit-saved-results` route which can be used on the "Update answers" link.

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,121 @@
...
+        scope: scopes,
+        state: state,
+        nonce: nonce,
+      ),
+      state: state,
+    }
+  end
+
+  def redirect_uri
+    host = ENV["GOVUK_WEBSITE_ROOT"] || "http://#{ENV['VIRTUAL_HOST']}"
```
barrucadu, line 36: Should `VIRTUAL_HOST` come first?  In govuk-docker both env vars are set but it won't work as a redirect URI because there's no router.

app/controllers/brexit_checker_controller.rb
```diff
@@ -52,12 +62,56 @@ def confirm_email_signup
...
+    @results_from_account ||= begin
+      Array(
+        update_session_tokens(
+          oidc.get_checker_attribute(
+            access_token: session[:access_token],
+            refresh_token: session[:refresh_token],
+          ),
+        ),
+      )
+                              rescue OidcClient::OAuthFailure
```
barrucadu, line 97: This is some weird indentation

app/lib/oidc_client.rb
brucebolt: It probably should.  I don't have the `GOVUK_WEBSITE_ROOT` set, which is probably why I haven't seen an issue.

app/controllers/brexit_checker_controller.rb
brucebolt: Yikes, looks like `rubocop -a` did this in an attempt to keep the `rescue` in alignment with `begin`.

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  def create
+    if logged_in?
+      redirect_to default_redirect_path
+      return
+    end
+
+    if Rails.env.test?
+      render plain: "Redirecting to login"
+    else
+      redirect_to oidc.auth_uri(return_path: params["redirect_path"])[:uri]
```
barrucadu, line 13: Could you also call this parameter `redirect_path`, to be consistent?

brucebolt: Done

config/routes.rb
```diff
@@ -18,21 +18,20 @@
     get "/test-search/search/opensearch" => "search#opensearch"
   end
 
-  # Routes for the for Transition/Brexit Checker
```
huwd, line 21: Did you mean to get rid of this comment?
I think it's there because this is all wierd tech debt (an app nested in an app), might be helpful to leave it highlighed?

config/routes.rb
```diff
@@ -18,21 +18,20 @@
...
+    get "/save-your-results" => "brexit_checker#save_results", as: :transition_checker_save_results
+  end
 
   # Transition/Brexit checker email signup routes
   get "/email/subscriptions/new", to: proc { [200, {}, [""]] }, as: :email_alert_frontend_signup
 
   get "/*slug/email-signup" => "email_alert_subscriptions#new", as: :new_email_alert_subscriptions
   post "/*slug/email-signup" => "email_alert_subscriptions#create", as: :email_alert_subscriptions
 
-  # GOV.UK Accounts discovery routes
```
huwd, line 33: Potentially.... ditto with this one...
As if the checker might outlive the account discovery... this could be worth highlighting to anyone that might want to unpick the code later?

config/routes.rb
```diff
@@ -18,21 +18,20 @@
     get "/test-search/search/opensearch" => "search#opensearch"
   end
 
-  # Routes for the for Transition/Brexit Checker
-  get "/transition-check/results" => "brexit_checker#results", as: :transition_checker_results
-  get "/transition-check/questions" => "brexit_checker#show", as: :transition_checker_questions
-  get "/transition-check/email-signup" => "brexit_checker#email_signup", as: :transition_checker_email_signup
-  post "/transition-check/email-signup" => "brexit_checker#confirm_email_signup", as: :transition_checker_confirm_email_signup
+  scope "/transition-check" do
```
huwd, line 21: Much cleaner though, great improvement!

config/environments/development.rb
```diff
@@ -54,3 +54,34 @@
...
   config.hosts.clear
 end
+
+# make discovery work over HTTP
+module OpenIDConnect
+  module Discovery
+    module Provider
+      class Config
+        class Resource
+          def initialize(uri)
```
huwd, line 64: So i get this has to be called explcitly. And that a sensible dev should only ever do that in our dev environmnet...

But I'm wondering if it's extra paranoid to add some sort of dump out check here... Like if it's initialized in the wrong environment then raise an error instead of executing?

Take it or leave it!

config/environments/development.rb
```diff
@@ -54,3 +54,34 @@
...
+            @host = uri.host
+            @port = uri.port unless [80, 443].include?(uri.port)
+            @path = File.join uri.path, ".well-known/openid-configuration"
+            @scheme = uri.scheme
+            attr_missing!
```
huwd, line 69: This is interesting, not come across this one before.
Couldn't find it in either the rails docs (found [attribute_missing](https://api.rubyonrails.org/classes/ActiveModel/AttributeMethods.html#method-i-attribute_missing) instead!), nor [ruby](https://ruby-doc.com/search.html?cx=011815814100681837392%3Awnccv6st5qk&q=attr_mising%21&sa=Search&cof=FORID%3A9), nor [this app](https://github.com/alphagov/finder-frontend/search?q=attr_missing)!

I must be missing something obvious, where does this come from and what does it do? (I have my suspicions it might raise an error if one of those uri attributes are missing, but would like to be sure!)

config/environments/development.rb
```diff
@@ -54,3 +54,34 @@
...
+          end
+
+          def endpoint
```
huwd, line 72: Not really :100:  clear what's going on here. What's the `SWD` where does it come from?

The commit message is nice and clear about what we're trying to achieve, but i'm not really sure how that's working here?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+private
+
+  def redirect_path
+    transition_checker_questions_path
+  end
+
+  def oidc
+    @oidc ||= OidcClient.new(
+      ENV["ACCOUNT_MANAGER_URL"],
```
huwd: Could we get this out of Plek instead?
```suggestion
      Plek.find("account-manager"),
```

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      ENV["OIDC_CLIENT_ID"],
+      ENV["OIDC_CLIENT_SECRET"],
```
huwd: I'm wondering if there's anything we can do to bring these together with [these Variables](https://github.com/alphagov/govuk-docker/pull/417/files#diff-81716fac3ebcda04e6914c20d3333c82R38-R42):

... would the client_id always be the same between the JWT and the OIDC? If so could we merge them and just use `GOVUK_ACCOUNT_OAUTH_CLIENT_ID` perhaps?

The client secret will almost certainly be unique as we don't want to confuse the JWT permissions for these... :thinking: 

What's actually in there, is it a key?
I'm leaning towards OTT descriptive names for these, ones that flag this is for GOV.UK accounts and if it's a key say if it's public or not, or perhaps just secret if it's something else?

Those are all things I've scratched my head over a bit when digging through the env...

perhaps `GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET`?

app/controllers/application_controller.rb
```diff
@@ -93,6 +93,14 @@ def filter_params
     end
   end
 
+  def oidc
+    @oidc ||= OidcClient.new(
+      ENV["ACCOUNT_MANAGER_URL"],
```
huwd: Could we get this out of Plek instead?
```suggestion
      Plek.find("account-manager"),
```

app/controllers/application_controller.rb
```diff
@@ -26,6 +28,19 @@ class ApplicationController < ActionController::Base
...
 
+  def skip_session_cookie
+    unless current_user
+      request.session_options[:drop] = true
+    end
+  end
+
+  def logout!
+    session.delete(:sub)
+    session.delete(:email)
```
barrucadu, line 39: This `session.delete(:email)` is no longer needed

app/controllers/application_controller.rb
```diff
@@ -93,6 +93,14 @@ def filter_params
     end
   end
 
+  def oidc
+    @oidc ||= OidcClient.new(
+      ENV["ACCOUNT_MANAGER_URL"],
+      ENV["OIDC_CLIENT_ID"],
+      ENV["OIDC_CLIENT_SECRET"],
```
huwd: See comments on [these lines](https://github.com/alphagov/finder-frontend/pull/2214/commits/b5c32978b7f5e897802e8c985333df0645ace55e#diff-d5241d488259f32ecbe2f636133e5ddaR47-R48)

app/controllers/application_controller.rb
```diff
@@ -95,9 +95,9 @@ def filter_params
 
   def oidc
     @oidc ||= OidcClient.new(
-      ENV["ACCOUNT_MANAGER_URL"],
-      ENV["OIDC_CLIENT_ID"],
-      ENV["OIDC_CLIENT_SECRET"],
+      ENV.fetch("ACCOUNT_MANAGER_URL"),
```
huwd: woop!
Though still with the Plek thing

app/controllers/application_controller.rb
```diff
@@ -95,9 +95,9 @@ def filter_params
...
+      ENV.fetch("OIDC_CLIENT_ID"),
```
huwd, line 99: could we squash these back at the end, or are you cool with these being separate?
I'm not sure what I prefer... In some ways this is nicely highlighted and the commit message makes the advantage clear.

See earlier comments [about hte Envar names though](https://github.com/alphagov/finder-frontend/pull/2214/commits/b5c32978b7f5e897802e8c985333df0645ace55e#diff-d5241d488259f32ecbe2f636133e5ddaR47-R48)

app/controllers/application_controller.rb
```diff
@@ -95,7 +95,7 @@ def filter_params
 
   def oidc
     @oidc ||= OidcClient.new(
-      ENV.fetch("ACCOUNT_MANAGER_URL"),
+      Services.accounts_api,
```
huwd, line 98: Doh! Ignore all my plek stuff then... this does exactly that :+1: 
Would you consider squashing this back after review?

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -0,0 +1,243 @@
...
+    def given_i_am_on_the_saved_results_page
+      visit transition_checker_saved_results_path
+    end
+  end
+
+  context "with accounts enabled" do
+    let(:attribute_service_url) { "http://attribute-service" }
+
+    before do
+      ENV["OIDC_CLIENT_ID"] = "id"
```
huwd, line 50: See earlier stuff about [these names](https://github.com/alphagov/finder-frontend/pull/2214/commits/b5c32978b7f5e897802e8c985333df0645ace55e#diff-d5241d488259f32ecbe2f636133e5ddaR47-R48)

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -0,0 +1,243 @@
+require "spec_helper"
+
+RSpec.feature "Brexit Checker accounts", type: :feature do
```
huwd, line 3: These tests look really clear and easy to follow, good job!

app/lib/services.rb
```diff
@@ -31,9 +31,7 @@ def self.registries
     Registries::BaseRegistries.new
   end
 
-  if Rails.configuration.feature_flag_govuk_accounts
-    def self.accounts_api
-      Plek.find("account-manager")
-    end
+  def self.accounts_api
```
huwd, line 34: Ah interesting, fair!

huwd: Relic from the Env.fetch days, woops - good catch

huwd (commented):
Looks good,
- A couple of questions about how bits are working where i've not seen similar code before.
- Generally really clear and nice to follow.
- Would be good to have a chat about those environment variable names, see if we can slim down the docker-compose environment sections at all?
- I'd 2nd your bit [here](https://github.com/alphagov/finder-frontend/pull/2214#issuecomment-703557981) this looks like a really good base to make further tweaks off?
- I realise you're still filling in from some comments, after you're done were you intendeing to do a bit of `git squash`? especially towards the end there some commit start to re-tread similar ground?

app/controllers/application_controller.rb
brucebolt: Missed that one!   Have removed.

config/routes.rb
brucebolt: Yes, it was deliberate.  I think now the routes are in a scope, their purpose are probably self-explanatory from the route `/transition-check`.

config/environments/development.rb
brucebolt: Since this is in the `development.rb` config file, I can't see any way this would be executed outside of development?

config/environments/development.rb
brucebolt: It comes from the [`attr_required` gem](https://github.com/nov/attr_required) (which is a dependency of `openid_connect`.  You're right, it raises an exception when an attribute is missing.

config/environments/development.rb
brucebolt: I have to confess I just plagiarised this code from the [barking permit app](https://github.com/alphagov/govuk-accounts-demo-app/blob/a1585ef01bf804307b0c93bc0be2035946367c97/config/environments/development.rb#L67-L96).  I'd have to defer to @barrucadu for the thinking about how this works.

app/controllers/sessions_controller.rb
brucebolt: I had changed this in a later commit (because the `accounts_api` service wasn't there when I started writing this code).  I've rebased so we use Plek from the start, which has removed one of the commits.

app/controllers/application_controller.rb
brucebolt: Yeah, I've fixed this up into the earlier commits so the "change to use Plek" commit is now gone.

app/lib/services.rb
brucebolt: Yeah, I decided to remove because it was causing some issues with testing and didn't seem like it mattered whether the method were available or not.

config/routes.rb
brucebolt: I'm hoping we should be able to unpick accounts through the pull request and commit history, should we need to remove them from the checker

config/environments/development.rb
barrucadu: I also plagiarised it

app/controllers/sessions_controller.rb
brucebolt: You're absolutely right.  I've just realised `GOVUK_ACCOUNT_OAUTH_CLIENT_ID` and `OIDC_CLIENT_ID` are the same.  I'll update to use the more descriptive name that already exists, so `OIDC_CLIENT_ID` can be removed from https://github.com/alphagov/govuk-docker/pull/417.

brucebolt: I'll also rename `OIDC_CLIENT_SECRET` to `GOVUK_ACCOUNT_OIDC_CLIENT_SECRET` as that is much clearer and consistent

app/controllers/application_controller.rb
brucebolt: I'll squash the commit down, since the fetch would be valid right from the start

config/environments/development.rb
brucebolt: Found [this comment](https://github.com/nov/openid_connect/issues/47#issuecomment-573701929) in my browser history.  Looks like that's where we got the code form, just with a bit of linting applied.

barrucadu (approved):
LGTM, have you tried running this end-to-end locally?

brucebolt:
Yes, everything appears to work as expected (with respect to logging in/out and retrieving the previous responses).  I'm not able to save results, but I think that's because there's an issue with my environment variables, but I can sort that out separately from this PR.

config/routes.rb
huwd: Fair enough, I guess it's less likely we'll rebrand the entire enterprise again this year.

config/routes.rb
huwd: fair, commits over comments, you're right!

config/environments/development.rb
huwd: ah very cool, thanks for the link!

config/environments/development.rb
huwd: cool cool

config/environments/development.rb
huwd: :D cool, no prob looks like it's a fairly generic piece of code anyway so that's likely fitting.
I'll spin up a rails console and have a look at the classes for SWD to find out a bit more about it.

Thanks for the trail!

huwd (approved):

PR #2218, opened 2020-10-01
Add level two restrictions

JonathanHallam:
Adds a yaml file that describes all of the gss codes that are currently under level two restrictions.

I've chosen app/lib as the place for this. This matches what's been done with the brexit checker. We should easily be able to add other levels to this and change things around as needed.

### DIFF
--- a/app/lib/local_restrictions/local-restrictions.yaml
+++ b/app/lib/local_restrictions/local-restrictions.yaml
@@ -1,0 +1,433 @@
+---
+E08000001:
+  alert_level: 2
+  name: Bolton
+  start_date:
+  end_date:
+  guidance:
+    label: The guidance
+    link: "guidance/bolton-local-restrictions"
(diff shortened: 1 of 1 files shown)


app/lib/local_restrictions/local-restrictions.yaml
```diff
@@ -0,0 +1,442 @@
+---
+E06000003:
+  alert_level: 2
```
Rosa-Fox, line 3: Redcar and Cleveland are level 1 so we need to take this chunk out!

Rosa-Fox (changes requested):

app/lib/local_restrictions/local-restrictions.yaml
JonathanHallam: Good spot, Sorted!!

Rosa-Fox (approved):

PR #2222, opened 2020-10-02
Model local restriction data

JonathanHallam:
Adds a service that takes a GSS code (from Mapit), and matches against the local lockdown data held in finder-frontend, and returns data for that area.

We have chosen to stub out the fixture path in order to separate test data from that which is live.

## Testing locally

The test suite can be run with `bundle exec rake`. You can also open a rails console and check that the LocalRestriction class is responding to the expected methods. 

```
irb(main):004:0> restriction = LocalRestriction.new("E08000001")
=> #<LocalRestriction:0x00007fc98d9b7c20 @gss_code="E08000001">
irb(main):005:0> restriction.area_name
=> "Bolton"
```

## Trello 
https://trello.com/c/40EHq00z/829-add-mapping-service-to-get-lockdown-data-from-gss-code

### DIFF
--- a/app/models/local_restriction.rb
+++ b/app/models/local_restriction.rb
@@ -1,0 +1,43 @@
+class LocalRestriction
+  FILE_PATH = "app/lib/local_restrictions/local-restrictions.yaml".freeze
+
+  attr_reader :gss_code
+
+  def initialize(gss_code)
+    @gss_code = gss_code
+  end
+
+  def area_name
+    restriction["name"]
+  end
+
+  def alert_level
+    restriction["alert_level"]
+  end
+
+  def guidance
+    restriction["guidance"]
+  end
+
+  def extra_restrictions
+    restriction["extra_restrictions"]
+  end
+
+  def start_date
+    restriction["start_date"]&.to_date
+  end
+
+  def end_date
+    restriction["end_date"]&.to_date
+  end
+
+private
+
+  def all_restrictions
+    @all_restrictions ||= YAML.load_file(FILE_PATH)
+  end
+
+  def restriction
+    all_restrictions[gss_code] || {}
+  end
+end
--- a/spec/fixtures/local-restrictions.yaml
+++ b/spec/fixtures/local-restrictions.yaml
@@ -1,0 +1,10 @@
+---
+E08000001:
+  alert_level: 4
+  name: Tatooine
+  start_date: "01 October 2020"
+  end_date: "02 October 2020"
+  guidance:
+    label: These are not the restrictions you are looking for
+    link: "guidance/tatooine-local-restrictions"
+  extra_restrictions:
--- a/spec/models/local_restriction_spec.rb
+++ b/spec/models/local_restriction_spec.rb
@@ -1,0 +1,57 @@
+require "spec_helper"
+
+RSpec.describe LocalRestriction do
+  before do
+    stub_const("LocalRestriction::FILE_PATH", "spec/fixtures/local-restrictions.yaml")
+  end
+
+  let(:restriction) { described_class.new("E08000001") }
+
+  it "returns the area name" do
+    expect(restriction.area_name).to eq("Tatooine")
+  end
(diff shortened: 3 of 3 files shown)


app/models/local_restriction.rb
```diff
@@ -0,0 +1,43 @@
...
+    area_data["end_date"].to_date
+  end
+
+private
+
+  def restriction_data
+    @restriction_data ||= YAML.load_file(FILE_PATH)
+  end
+
+  def area_data
```
leenagupte, line 40: I wonder if we might need to make this public later, so that we can check if any restrictions exist for a gss code. i.e. `LocalRestriction.new("E000000001").area_data.present?` 🤔

spec/fixtures/local-restrictions.yaml
```diff
@@ -0,0 +1,10 @@
+---
+E08000001:
+  alert_level: 4
+  name: Tatooine
```
leenagupte, line 4: ❤️

app/models/local_restriction.rb
```diff
@@ -0,0 +1,43 @@
...
+    area_data["start_date"].to_date
+  end
+
+  def end_date
+    area_data["end_date"].to_date
+  end
+
+private
+
+  def restriction_data
```
leenagupte, line 36:
```suggestion
  def all_restrictions
```

app/models/local_restriction.rb
leenagupte, line 40:
```suggestion
  def restriction
```

spec/models/local_restriction_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "spec_helper"
+
+RSpec.describe LocalRestriction do
+  before do
+    stub_const("LocalRestriction::FILE_PATH", "spec/fixtures/local-restrictions.yaml")
```
leenagupte, line 5: Nice!

leenagupte (commented):
Looking good!
A couple of naming thoughts, but nothing major.

app/models/local_restriction.rb
JonathanHallam: Yeah, good point. I guess we could use a proxy like `LocalRestriction.new("E000000001").area_name.present?` but it's not as clear.

app/models/local_restriction.rb
JonathanHallam: Good shout 👍

app/models/local_restriction.rb
JonathanHallam: done! 🥇

app/models/local_restriction.rb
```diff
@@ -0,0 +1,43 @@
...
+  def guidance
+    restriction["guidance"]
+  end
+
+  def extra_restrictions
+    restriction["extra_restrictions"]
+  end
+
+  def start_date
+    restriction["start_date"].to_date
```
leenagupte, line 27: I think we need to add tests for when start_date and end_date is nil:

```
irb(main):005:0> rst = LocalRestriction.new("E06000008")
=> #<LocalRestriction:0x0000562becd45eb8 @gss_code="E06000008">
irb(main):006:0> rst.start_date
Traceback (most recent call last):
        2: from (irb):6
        1: from app/models/local_restriction.rb:27:in `start_date'
NoMethodError (undefined method `to_date' for nil:NilClass)
Did you mean?  to_d
```

It should be possible to fix this with a safe navigation operator, e.g.:

```
restriction["start_date"]&.to_date
```

leenagupte (changes requested):
`start_date` and `end_date` throw an error if they're nil. Once that's fixed this is good to go!

app/models/local_restriction.rb
JonathanHallam: ooh good spot, sorry! Been fixed now and tests added :)

leenagupte (approved):

PR #2231, opened 2020-10-08
Check accounts are enabled before sessions controller

barrucadu:
I realised we probably want this here as well.

---

[Trello card](https://trello.com/c/F2pCU73B/298-implement-signing-in-to-the-checker)


brucebolt (approved):
Good idea

PR #2233, opened 2020-10-08
Add timestamp to transition_checker_state attribute

barrucadu:
This PR goes with https://github.com/alphagov/govuk-account-manager-prototype/pull/170

### DIFF
--- a/app/lib/brexit_checker/account_jwt.rb
+++ b/app/lib/brexit_checker/account_jwt.rb
@@ -25,7 +25,12 @@ private
   end
 
   def attributes
-    { transition_checker_state: criteria_keys }
+    {
+      transition_checker_state: {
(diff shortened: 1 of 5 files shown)


app/lib/brexit_checker/account_jwt.rb
```diff
@@ -25,7 +25,12 @@ def scopes
   end
 
   def attributes
-    { transition_checker_state: criteria_keys }
+    {
```
brucebolt, line 28: Do we have any tests around the saving of attributes?  It feels like something important that should have test coverage.

barrucadu: I don't think we do, I'll add some

huwd (approved):

PR #2242, opened 2020-10-12
Email topic slug

huwd:
Relates to: [GOV.UK accounts | Send email topic slug in the JWT](https://trello.com/c/zZ7uOGy4/326-send-email-topic-slug-in-the-jwt)

## What
- Make the query to email-alert-api for checker results subscription slugs reusable
- Query it (or get it as a memoised value) when we're building a JWT to send over to accounts
- Put it in the payload to be encoded into the Account JWT.

## Why

On account creation we want to offer users the chance to be notified of changes to their checker results. So we'll send this slug across to the account manager so we can set up a subscription if the user chooses to.


huwd:
Can squash these all down into a single commit if you want, just thought it'd be easier to reivew in these four bits.

barrucadu (approved):
:+1: I think squash these commits, so we have one commit which passes its tests.

huwd:
Done

PR #2249, opened 2020-10-15
Update email alerts when updating answers

barrucadu:
If the user doesn't already have an active subscription, prompt them
to create one.  This isn't quite ideal, as we now have the email
signup frontend and content in both the account manager (for the
registration journey) and here, but we want to rethink this whole
email integration later anyway.

If the user already has an active subscription just update it without
explicitly asking, as we tell them on the confirmation page that this
will happen.

---

[Trello card](https://trello.com/c/9UxDwIhI/340-in-the-update-flow-create-or-update-email-subscriptions)

### DIFF
--- a/app/lib/oidc_client.rb
+++ b/app/lib/oidc_client.rb
@@ -59,12 +59,14 @@ class OidcClient
       access_token: access_token,
       refresh_token: refresh_token,
       method: :get,
+      uri: attribute_uri,
     )
 
-    if response[:result].empty?
+    body = response[:result].body
+    if body.empty?
       response.merge(result: {})
     else
-      response.merge(result: JSON.parse(response[:result])["claim_value"])
+      response.merge(result: JSON.parse(body)["claim_value"])
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/lib/oidc_client.rb
```diff
@@ -73,26 +75,50 @@ def set_checker_attribute(value:, access_token:, refresh_token:)
...
+      method: :post,
+      uri: email_subscription_uri,
+      arg: { topic_slug: slug },
+    )
+  end
+
 private
 
-  def oauth_request(access_token:, refresh_token:, method:, arg: nil)
+  OK_STATUSES = [200, 204, 404, 410].freeze
```
huwd, line 106: How come these are OK?

barrucadu: These are the ones which we might legitimately return from the API, anything else would be unexpected, so it should retry the request.

app/lib/oidc_client.rb
```diff
@@ -73,26 +75,50 @@ def set_checker_attribute(value:, access_token:, refresh_token:)
...
     access_token_str = access_token
     refresh_token_str = refresh_token
 
-    args = [attribute_uri, arg].compact
+    args = [uri, arg].compact
 
     response = Rack::OAuth2::AccessToken::Bearer.new(access_token: access_token_str).public_send(method, *args)
 
-    unless [200, 404].include? response.status
+    unless OK_STATUSES.include? response.status
```
huwd, line 116: Not sure i understand this bit.
So if our response isn't 200, 204, 404, 410 then we refresh...

app/lib/oidc_client.rb
```diff
@@ -73,26 +75,50 @@ def set_checker_attribute(value:, access_token:, refresh_token:)
...
       client.refresh_token = refresh_token
       access_token = client.access_token!
 
       response = access_token.public_send(method, *args)
-      raise OAuthFailure unless [200, 404].include? response.status
+      raise OAuthFailure unless OK_STATUSES.include? response.status
```
huwd, line 121: then if the refreshed token isn't 200, 204, 404, 410 then we raise an OAuthFailure.

Wouldn't we always want this to be 200?
When is 204, 404, 410 relevant?

huwd (approved):

app/lib/oidc_client.rb
barrucadu: Would it help if I changed this to something like

```
if [401, 403].include? response.status
    # refresh tokens and retry
elsif response.status >= 500
    # retry
elsif OK_STATUSES.include? response.status
    # this is ok
else
    # this is an unrecoverable error
```

app/lib/oidc_client.rb
huwd: ah ha: https://github.com/alphagov/govuk-account-manager-prototype/pull/194#pullrequestreview-509327925

barrucadu: The email endpoint can return a 204, 404, or 410 legitimately.  It could instead always return a 200 and say what the response really means in the body, but that feels like reinventing status codes.

PR #2253, opened 2020-10-19
Rename JWT environment variables

brucebolt:
This makes it clear the environment variables are related to JWT, rather than the OAuth flow.

This corresponds with the environment variables set in https://github.com/alphagov/govuk-puppet/pull/10777.

Trello card: https://trello.com/c/CyOeHwIY


barrucadu (approved):

spec/features/brexit_checker/email_signup_spec.rb
```diff
@@ -40,8 +40,8 @@
     before do
```
huwd, line 40: We should stick a note in these for the testing discussion,
Turns out [we should likely be using Climate Control here](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#testing-utilities)

huwd (approved):

PR #2260, opened 2020-10-22
Validate grouping criteria of actions when we import them.

hannako:
## What
Start the process of validating data before we store it as yaml.

## Context

The Brexit Checker relies on the content of yaml files rather than a database to persist its data: [Actions](https://github.com/alphagov/finder-frontend/blob/master/app/lib/brexit_checker/actions.yaml), Criteria, Groups and Questions.

When we import the master spreadsheet of actions and [write that to yaml](https://github.com/alphagov/finder-frontend/blob/3878014c4f971bb4d188b9dc97824a63b1732515/lib/tasks/brexit_checker/convert_csv_to_yaml.rake#L15), we should be checking that the actions pass their validations, so that bad data is not added to the yaml. 

Currently there are validations set on the model, but by the time these are run the data is already persisted so all they are doing is providing a way to raise an error that would cause the page to blow up. 

## How

I suggest we start to move the validations out of the models and instead run them:
- When new data is imported, in the case of actions.
- And as part of the CI, as it's possible to update the yaml files without running the import task.

## Why

- Yaml files can only be updated by a developer following a PR review,  and it's obvious if something is deleted or totally missing. But it would be less easy to spot a minor typo in a criteria key when a new action is added. 
- It will mean we can simplify the models, and test the validations. Currently there are no test for model validations because they are run as part of the initialiser.

## First step
This PR adds a validation to actions, checking that each action has a [valid citizen grouping criteria](https://github.com/alphagov/finder-frontend/blob/3878014c4f971bb4d188b9dc97824a63b1732515/app/lib/brexit_checker/group.rb#L6). If this looks like a good plan, I will move the [remaining action validations](https://github.com/alphagov/finder-frontend/blob/master/app/lib/brexit_checker/action.rb#L8-L13) in to the action validator in a subsequent PR.

---


hannako:
Superceded by https://github.com/alphagov/finder-frontend/pull/2291

PR #2263, opened 2020-10-26
Manage our own session cookie for accounts

barrucadu:
Using the rails session management has a couple of problems: we had to
implement the "has_session" hack to avoid setting the cookie on people
who haven't opted in to accounts; and we can't delete the cookie, only
reset it to a blank session.  It would be nice if the cookie only
existed while a user was logged in, because then we could use "does
the cookie exist?" in the collections app to determine whether to
render the accounts header on /transition.

By managing our own cookies we solve both of those problems: we can
set the cookie on log in and delete it on log out.  We do need to bump
the expiration time ourselves, which is a bit of extra work, but I
think the benefits outweigh that.

---

[Trello card](https://trello.com/c/vdZrgBRy/363-make-changes-to-transition-for-trial-launch)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,8 +8,6 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   helper :application
 
-  after_action :skip_session_cookie
-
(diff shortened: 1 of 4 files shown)


huwd (approved):
Looks great

PR #2271, opened 2020-10-27
Implement rough frontend for the "save changes" screens

barrucadu:
This isn't quite right, but it's much closer than it was, and now has the right content.

## Results are different

I've gone for a ~~success alert~~ notice here because it's the closest component to what we want, but we actually want something more bespoke:

Screenshot 2020-10-28 at 09 05 28

## Results comparison

I think we want some highlighting or shading on the two answer columns:

Screenshot 2020-10-27 at 18 11 07

## Saved

Again, we want something a bit different to the success alert:

Screenshot 2020-10-27 at 18 11 40

---

[Trello card](https://trello.com/c/uTPaVhv1/344-implement-frontend-for-change-your-answers-journey)

### DIFF
--- a/config/locales/en/brexit_checker/results.yml
+++ b/config/locales/en/brexit_checker/results.yml
@@ -34,3 +34,10 @@ en:
         business:
           heading: Your business or organisation
       print_link: Print your results
+      accounts:
+        results_differ:
+          message: These results are different to the ones saved in your account.
+          description: You can <a class="govuk-link" href="%{link}">see what answers you’ve changed and save these results to your account instead.</a>
+        results_saved:
+          message: Success
+          description: Your results have been saved.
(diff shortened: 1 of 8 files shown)


config/locales/en/brexit_checker/results.yml
```diff
@@ -34,3 +34,10 @@ en:
         business:
           heading: Your business or organisation
       print_link: Print your results
+      accounts:
+        results_differ:
+          message: These results are different to the ones saves in your account.
```
brucebolt, line 39:
```suggestion
          message: These results are different to the ones saved in your account.
```

brucebolt (approved):
Looks good to me, but you'll probably want to get a front end review too

alex-ju (approved):

PR #2289, opened 2020-11-03
Persist cross domain tracking through form leading to accounts

andysellick:
[Trello](https://trello.com/c/4kIhJHvE/390-analytics-snags)

## What

The button leading to the accounts system is inside a form, so cross domain tracking doesn't work through it (unlike links, which are automatically decorated). So we're going to need to manually decorate the target of the form with the ga parameter.

This is a proof of concept using an approach of clicking on the button, but I think it'll be easier if we have the decoration happening on page load, so we don't have to intercept the button click.

References:

- https://www.knewledge.com/en/blog/2013/11/cross-domain-tracking-for-links-with-gtm/#manual-tagging

### DIFF
--- a/app/assets/javascripts/ga_cross_domain_form_tracking.js
+++ b/app/assets/javascripts/ga_cross_domain_form_tracking.js
@@ -1,0 +1,25 @@
+/* eslint-env jquery */
+/* global ga */
+
+function attachCrossDomainTrackerToInput (form, trackers) {
+  if (trackers.length) {
+    var existingAction = form.attr('action')
+    var linker = new window.gaplugins.Linker(trackers[0])
+    var trackedAction = linker.decorate(existingAction)
+    form.attr('action', trackedAction)
+  }
+}
+
+window.GOVUK = window.GOVUK || {}
+var GOVUK = window.GOVUK
+GOVUK.attachCrossDomainTrackerToInput = attachCrossDomainTrackerToInput
+
+$(window).on('load', function () {
+  if (window.ga !== undefined && typeof window.ga.getAll === 'function') {
+    var trackers = ga.getAll()
+    var form = $('form#account-signup')
+    if (form.length) {
+      window.GOVUK.attachCrossDomainTrackerToInput(form, trackers)
+    }
+  }
+})
--- a/spec/javascripts/ga_cross_domain_form_tracking_spec.js
+++ b/spec/javascripts/ga_cross_domain_form_tracking_spec.js
@@ -1,0 +1,53 @@
+/* eslint-env jasmine, jquery */
+
+describe('attachCrossDomainTrackerToInput', function () {
+  var form
+  var tracker
+  var linker
+  var GOVUK = window.GOVUK || {}
+
+  beforeEach(function () {
+    form = $('<form method="POST" action="/somewhere" class="account-signup">' +
+               '<input type="hidden" name="key" value="value" />' +
+               '<button type="submit">Create a GOV.UK account</button>' +
+             '</form>')
+
+    window.ga = {
+      getAll: function () {}
+    }
+
+    window.gaplugins = {
+      Linker: function () {}
+    }
+
+    linker = {
+      decorate: function () {}
+    }
+
+    spyOn(window.ga, 'getAll').and.returnValue([])
+    spyOn(window.gaplugins, 'Linker').and.returnValue(linker)
+    spyOn(linker, 'decorate').and.returnValue('/somewhere?_ga=abc123')
+  })
+
+  afterEach(function () {
+    form.remove()
+  })
+
+  it('leaves the form action unchanged if ga is not present', function () {
+    tracker = []
+    GOVUK.attachCrossDomainTrackerToInput(form, tracker)
+    expect(form.attr('action')).toEqual('/somewhere')
+  })
+
+  it('leaves the form action if unchanged there are no trackers in ga', function () {
+    tracker = []
+    GOVUK.attachCrossDomainTrackerToInput(form, tracker)
+    expect(form.attr('action')).toEqual('/somewhere')
+  })
+
+  it('modifies the form action to append ids from ga to the destination url', function () {
+    tracker = [{ ga_mock: 'foobar' }]
+    GOVUK.attachCrossDomainTrackerToInput(form, tracker)
+    expect(form.attr('action')).toEqual('/somewhere?_ga=abc123')
+  })
+})
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -23,6 +23,7 @@
 //= require live_search
 //= require search-analytics
 //= require taxonomy-select
+//= require ga_cross_domain_form_tracking
 //= require_tree ./modules
 //= require_tree ./components
 
--- a/app/views/brexit_checker/save_results.html.erb
+++ b/app/views/brexit_checker/save_results.html.erb
@@ -66,7 +66,7 @@
           <% end %>
         </ul>
 
-        <%= form_tag Services.accounts_api, id: "account-signup", class: "account-signup" do %>
+        <%= form_tag Services.accounts_api, id: "account-signup" do %>
           <input type="hidden" name="jwt" value="<%= @account_jwt %>">
           <%= render "govuk_publishing_components/components/button", {
             text: t('brexit_checker.account_signup.cta_button'),


huwd:
if you're happy @andysellick I'll rebase those commits?

app/assets/javascripts/ga_cross_domain_form_tracking.js
```diff
@@ -0,0 +1,17 @@
...
+    var existingAction = form.attr("action");
+    var linker = new window.gaplugins.Linker(trackers[0])
+    var trackedAction = linker.decorate(existingAction)
+    form.attr("action", trackedAction);
+  }
+}
+
+$(window).on('load', function() {
+  if (ga !== undefined && typeof ga.getAll === 'function') {
+    var form = $("form#account-signup");
```
andysellick, line 12: We should probably check that `form` exists before proceeding here. I think `if (form.length)` should do it?

app/assets/javascripts/ga_cross_domain_form_tracking.js
```diff
@@ -0,0 +1,17 @@
+function attachCrossDomainTrackerToInput(form, trackers) {
```
andysellick, line 1: We should probably have a spec file for this code. Should be a case of having a chunk of form markup, running the code on it and checking the action attribute has been updated. Might need some cleverness to test the right ga parameter has been appended.

app/assets/javascripts/ga_cross_domain_form_tracking.js
```diff
@@ -0,0 +1,17 @@
...
+  if (trackers.length) {
+    var existingAction = form.attr("action");
```
andysellick, line 3: Nitpick: we're moving away from semicolons at the end of lines. Not sure why that hasn't been flagged by the linter.

app/assets/javascripts/ga_cross_domain_form_tracking.js
huwd: humm, I've not done a lot of mocking and stubbing in Jasmine, would need to go read up.

Would be happy to pair through this if you've got an idea of how this would work?

huwd:
I've added those tests @andysellick, feels a bit more like i'm testing the mocks and the code though... perhaps that's OK given how much we trust the GA code is well tested elsewhere though?

Let me know any thoughts / suggestions?

huwd:
Looks like this:
ga_x_domain

app/assets/javascripts/ga_cross_domain_form_tracking.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-env jquery */
+/* global ga */
```
huwd, line 2: Stole this from [here](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/analytics/ecommerce.js#L2)

Is this the right way to do it?

spec/javascripts/ga_cross_domain_form_tracking_spec.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-env jasmine, jquery */
+/* global attachCrossDomainTrackerToInput */
```
huwd, line 2: Seems like there should be a better way to quiet the liner here...
Am i supposed to be importing / destrcturing these into my tests... they are globally available?

Fees like fighting the linter?

spec/javascripts/ga_cross_domain_form_tracking_spec.js
```diff
@@ -0,0 +1,53 @@
...
+    form.remove()
+  })
+
+  it('leaves the form action unchanged if ga is not present', function () {
+    tracker = []
+    attachCrossDomainTrackerToInput(form, tracker)
+    expect(form.attr('action')).toEqual('/somewhere')
+  })
+
+  it('leaves the form action unchanged there are no trackers in ga', function () {
```
andysellick, line 42: Nitpick, missing `if` between `unchanged` and `there`.

andysellick, line 42: Am I missing something or is this identical to the previous test?

spec/javascripts/ga_cross_domain_form_tracking_spec.js
```diff
@@ -0,0 +1,53 @@
...
+    tracker = []
+    attachCrossDomainTrackerToInput(form, tracker)
+    expect(form.attr('action')).toEqual('/somewhere')
+  })
+
+  it('modifies the form action to append ids from ga to the destination url', function () {
+    tracker = [{ ga_mock: 'foobar' }]
```
andysellick, line 49: This might not need any changes, but I don't really understand this. Is it the case that if `tracker` contains anything, the code runs? Or does it specifically have to contain a `ga_mock` attribute? If so, what is `ga_mock`?

spec/javascripts/ga_cross_domain_form_tracking_spec.js
andysellick: What we should probably be doing is adding `attachCrossDomainTrackerToInput` to the `window.GOVUK` object. Which would involve more work. Sorry. But then you wouldn't have to do this.

app/assets/javascripts/ga_cross_domain_form_tracking.js
andysellick: Not sure this is really needed - does the linter really throw an error?

andysellick (commented):
This all looks good apart from my comments, but I think the big one is that we need to include this as part of the window.GOVUK object. It's not only best practice and our practice, but it should help with a few things. Sorry, should have realised this sooner.

app/assets/javascripts/ga_cross_domain_form_tracking.js
huwd: Yeah if I remove it, it gets picked up on CI

spec/javascripts/ga_cross_domain_form_tracking_spec.js
huwd: ah, humm yeah initially we were initalising GA within the test... and we were going to have a test to show that the code doesn't fire if GA isn't around.

But this might be an issue here...

app/assets/javascripts/ga_cross_domain_form_tracking.js
huwd: Yeah reconfirmed if you run `govuk-docker-run yarn lint" you can see the error if you remove that line

spec/javascripts/ga_cross_domain_form_tracking_spec.js
huwd: So i think i've done that here: https://github.com/alphagov/finder-frontend/pull/2289/commits/8eede1974f3e8127018d7f5725df1013a63f1d70

spec/javascripts/ga_cross_domain_form_tracking_spec.js
huwd: So this code will only be fired if this array has any length.
In the wild we get two fairly complex ga tracking objects, with the first being cross domain.

For these tests, as we're mocking the results (again starting to feel a bit more like testing the mocks than the logic, but I suppose we're intercepting known steps, so perhaps ok?).

So i just need something here and i figured `[{}]` unclear and not descriptive enough.

So the keys and values don't do anything except describe that this is an object. Happy to take any suggestions on preferred ways of doing things

alex-ju (changes requested):
For this to follow the GOV.UK modules structure we'd need further refactoring

Attaching the script would be done using data attributes
```erb
<%= form_tag Services.accounts_api, id: "account-signup", data: { "module": "attach-cross-domain-tracker-to-input" } do %>
```

The JavaScript will use the existing Module system and will fire `onready`, which is quicker than `onload`.
```js
(function(Modules) {
  'use strict'

  function AttachCrossDomainTrackerToInput () {}
  AttachCrossDomainTrackerToInput.prototype.start = function($form) {
    if (window.ga !== undefined && typeof window.ga.getAll === 'function') {
      var trackers = ga.getAll()
      if (trackers.length) {
        var existingAction = $form.attr('action')
        var linker = new window.gaplugins.Linker(trackers[0])
        var trackedAction = linker.decorate(existingAction)
        form.attr('action', trackedAction)
      }
  }
  Modules.AttachCrossDomainTrackerToInput = AttachCrossDomainTrackerToInput
})(window.GOVUK.Modules)
```

Note the code above was written in the browser 😅 – happy to pair on this if you'd like to talk things through.

alex-ju (approved):
The module pattern doesn't work for this use-case, unfortunately, so we'll get this in for pragmatic reasons until we figure out the issue with the module-based approach.

huwd:
Alex and I paired on this, moved the last comment across to own PR, where there's some outstanding tech debt challenges:
https://github.com/alphagov/finder-frontend/pull/2309

PR #2291, opened 2020-11-04
Brexit checker: Test model validations

hannako:
## What

- Adds and extends unit tests for the Brexit checker models.

- Move the validations for the group model into a custom validator to make it easier to extend this model.

[trello](https://trello.com/c/VN5HJYMS/573-tidy-up-checker-in-prep-for-business-groupings)

----

### DIFF
--- a/app/lib/brexit_checker/action.rb
+++ b/app/lib/brexit_checker/action.rb
@@ -5,7 +5,7 @@ class BrexitChecker::Action
 
   CONFIG_PATH = Rails.root.join("app/lib/brexit_checker/actions.yaml")
 
-  validates_presence_of :id, :title, :consequence, :criteria
+  validates_presence_of :id, :title, :consequence
   validates_inclusion_of :audience, in: %w[business citizen]
   validates_presence_of :guidance_link_text, if: :guidance_url
   validates_numericality_of :priority, only_integer: true
@@ -75,7 +75,7 @@ private
   def has_criteria
     return unless all_criteria_keys.none?
 
-    errors.add "Action must have at least one criterion"
+    errors.add(:criteria, "can't be blank")
   end
 
   def path_from_url(full_url)
@@ -88,8 +88,8 @@ private
   end
 
   def citizen_action_has_grouping_criteria
-    if audience == "citizen" && grouping_criteria.nil?
-      errors.add(:grouping_criteria, "Can't be empty for citizen actions")
+    if audience == "citizen" && grouping_criteria.blank?
+      errors.add(:grouping_criteria, "can't be empty for citizen actions")
     end
   end
 end
--- a/app/lib/brexit_checker/validators/group_validator.rb
+++ b/app/lib/brexit_checker/validators/group_validator.rb
@@ -1,0 +1,24 @@
+class BrexitChecker::Validators::GroupValidator < ActiveModel::Validator
+  CITIZEN_KEYS = %w[ visiting-eu
+                     visiting-uk
+                     visiting-ie
+                     living-eu
+                     living-ie
+                     living-uk
+                     working-uk
+                     studying-eu
+                     studying-uk
+                     common-travel-area ].freeze
+
+  def validate(record)
+    validate_citizen_group(record)
+  end
+
+private
+
+  def validate_citizen_group(record)
+    unless CITIZEN_KEYS.include? record.key
+      record.errors[:key] << "is not included in the list"
+    end
+  end
+end
--- a/app/lib/brexit_checker/validators/group_validator.rb
+++ b/app/lib/brexit_checker/validators/group_validator.rb
@@ -17,7 +17,7 @@ class BrexitChecker::Validators::GroupValidator < ActiveModel::Validator
 private
 
   def validate_citizen_group(record)
-    unless CITIZEN_KEYS.include? record.key
+    unless CITIZEN_KEYS.include?(record.key)
       record.errors[:key] << "is not included in the list"
     end
   end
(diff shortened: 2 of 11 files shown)


app/lib/brexit_checker/validators/group_validator.rb
```diff
@@ -0,0 +1,24 @@
+class BrexitChecker::Validators::GroupValidator < ActiveModel::Validator
```
huwd, line 1: This is a _much_ better idea, not sure why we didn't do this last year already! Nice one!

app/lib/brexit_checker/validators/group_validator.rb
```diff
@@ -0,0 +1,24 @@
...
+                     common-travel-area ].freeze
+
+  def validate(record)
+    validate_citizen_group(record)
+  end
+
+private
+
+  def validate_citizen_group(record)
+    unless CITIZEN_KEYS.include? record.key
```
huwd, line 20: Interesting how does this line get past rubocop?
I thought rubocop enforces parenthesis...
so would be:
```
    unless CITIZEN_KEYS.include?(record.key)
```

Surprised the linter doesn't get stroppy?

spec/lib/brexit_checker/action_spec.rb
```diff
@@ -1,6 +1,51 @@
 require "spec_helper"
 
 RSpec.describe BrexitChecker::Action do
+  describe "validations" do
```
huwd, line 4: :tada:

app/lib/brexit_checker/action.rb
```diff
@@ -5,7 +5,7 @@ class BrexitChecker::Action
 
   CONFIG_PATH = Rails.root.join("app/lib/brexit_checker/actions.yaml")
 
-  validates_presence_of :id, :title, :consequence, :criteria
+  validates_presence_of :id, :title, :consequence
```
huwd, line 8: Good catch

app/lib/brexit_checker/action.rb
```diff
@@ -88,8 +88,8 @@ def path_from_url(full_url)
   end
 
   def citizen_action_has_grouping_criteria
-    if audience == "citizen" && grouping_criteria.nil?
-      errors.add(:grouping_criteria, "Can't be empty for citizen actions")
+    if audience == "citizen" && grouping_criteria.blank?
```
huwd, line 91: so what's the deal with this one? Was `grouping_criteira` an array all along?

spec/lib/brexit_checker/notification_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    before :each do
+      allow(BrexitChecker::Action).to receive(:load_all).and_return([action, action2])
+    end
+
+    it "returns the action relating to that notification" do
+      expect(notification.action).to eq action
+    end
+  end
+
+  describe ".load" do
```
huwd, line 34: Great tests,
minor question/quibble on the naming.

On line 20 we're testing a method with `#action` here we're testing `.load`.
Is there something I'm missing signified by the difference in name?

Not sure if we have a GOV.UK standard for these, I've seen a lot of `#action` style descriptions around I think, but if there's not someting different between them, it'd be good to make these consistent.

spec/lib/brexit_checker/question_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    it "#options returns an array of option objects" do
+      expect(question.options.first).to be_an_instance_of(BrexitChecker::Question::Option)
+    end
+
+    it "#options removes an option if the criteria matches the option's exclude_if value " do
+      expect(question.options("living-uk")).to eq([])
+    end
+  end
+
+  describe ".find_by_key" do
```
huwd, line 75: So again dots or #, I guess so long as we're consistent or there's a reason either are cool?

app/lib/brexit_checker/validators/group_validator.rb
hannako: Yes that is weird based on (the docs)[https://rubystyle.guide/#method-invocation-parens]. I grepped finder frontend and we seems to do it both ways. I'll parenthesise it though, thanks.

hannako: Ah found an unrelated [ruby style guide example](https://rubystyle.guide/#ranges-or-between) where `.include?` is used, and they parenthesise 👍

spec/lib/brexit_checker/notification_spec.rb
hannako: The [convention](https://rspec.rubystyle.guide/#describe-the-methods) for describing class methods is "." and for instance methods it's "#".

spec/lib/brexit_checker/question_spec.rb
hannako: See [docs](https://rspec.rubystyle.guide/#describe-the-methods) here

app/lib/brexit_checker/action.rb
hannako: It's an array of criteria keys, ie strings. If the grouping criteria is left off the action on the spreadsheet, when we import the action, `grouping_criteria.nil?` would return true so it works fine. But I changed it to `.blank?` so we would catch `nil`, `[]`, and `{}` so if a dev manually adds new grouping criteria and gets the structure wrong something fails!

huwd: much better, great improvement!

spec/lib/brexit_checker/notification_spec.rb
huwd: ahh i see, so action is an instance.
Well TIL.. time to go back and rewrite some tests in my current project :blush: 
Thanks for the links

huwd (approved):

PR #2305, opened 2020-11-16
Use name for date filter title

thomasleese:
At the moment the title of the date filter is hard coded to the string "Date". This means that if we want to display two date filters, then we end up with two filters both with the word "Date".

I'm not really clear on the consequence for this and whether it was intentional, so it might need some product/frontend approval first.

## [Flood and Coastal Erosion Risk Management Research Reports Finder](https://www-origin.integration.publishing.service.gov.uk/flood-and-coastal-erosion-risk-management-research-reports)

This is the new finder which triggered this work.

### Collapsed

Screenshot 2020-11-16 at 14 48 13

### Before

Screenshot 2020-11-16 at 14 39 35

### After

Screenshot 2020-11-16 at 14 43 57

### Before

Screenshot 2020-11-16 at 14 42 56

### After

Screenshot 2020-11-16 at 14 44 08

[Trello Card](https://trello.com/c/atDp5GQO/2199-3-add-flood-and-coastal-erosion-specialist-document)

### DIFF
--- a/app/views/components/_date-filter.html.erb
+++ b/app/views/components/_date-filter.html.erb
@@ -8,9 +8,7 @@
   date_errors_to ||= nil
 %>
 <% if key && name %>
-  <%= render "components/expander", {
-      title: "Date"
-    } do %>
+  <%= render "components/expander", { title: name } do %>
     <div class="app-c-date-filter govuk-!-margin-bottom-0" id="<%= key %>" <% if aria_controls_id %> data-module="enable-aria-controls"<% end %>>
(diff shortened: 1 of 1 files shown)


nicholsj:
Looks good to me - the name of the filter appears to be more descriptive for the user anyway, as opposed to a simple `date`

maxgds (approved):
Seems very sensible to me. From an accessibility perspective disambiguation is good - reduces cognitive load and makes things targetable for voice control more easily.

PR #2309, opened 2020-11-18
[WIP] Put GA tracking in the GOVUK Module

huwd:
@alex-ju and I worked on this one.
It takes the existing tracking code written in this PR, and updates it to proper GOV.UK module standards.

However, in doing so we discovered a strage bug where there `ga` object is being overwritten and not passing the tests.
We suspect this is a race condition with some of the module code not being loaded by the time our x-domain analytics code attempts to fire here.

here's an example output, with console.logs in at various stages of the code initialising.
We then re-run the code after the page has loaded and get a different result... which is suspcious.

I'm leaving this here as a note towards tech debt.

@alex-ju has spotted race condition in other parts of the GA setup, perhaps we can add this as a specific instance when we're unravelling that tech debt.

image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -23,6 +23,7 @@
 //= require live_search
 //= require search-analytics
 //= require taxonomy-select
+//= require ga_cross_domain_form_tracking
 //= require_tree ./modules
 //= require_tree ./components
 
--- a/app/assets/javascripts/ga_cross_domain_form_tracking.js
+++ b/app/assets/javascripts/ga_cross_domain_form_tracking.js
@@ -1,0 +1,25 @@
+/* eslint-env jquery */
+/* global ga */
+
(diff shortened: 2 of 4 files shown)


andysellick:
I've found some interesting behaviour in testing. My understanding is that if cross domain tracking has worked, the pageviews fired on both sides should have the same value in them for `clientId`. Based on some initial testing of the live site, this was not the case, but when I cleared out my cookies (for both GOV.UK and the accounts system) then accepted them again, all of the `clientId` values on both sides were the same.

I think what this means is that this code is working correctly for new users, but not necessarily for existing users who already have cookie preferences set on accounts. This is somewhat out of the scope of this PR, but I thought it worth recording.

huwd:
@andysellick I think we can kill this PR right?
This was back in our days of x-domain tracking which we now have working?

Shout if I should re-open this?

PR #2315, opened 2020-11-24
Update email signup with new design

benthorner:
https://trello.com/c/mk5M2aKD/619-general-design-and-stylistic-updates-to-signup-workflow

This is to be consistent with the page for content item signups [1],
when there are no modifiable choices. In particular, while some signup
pages may have a description, we have decided these are too inconsistent
to provide any value, so we have removed them altogether. The design also
calls for the removal of the contextual title, and "alert" wording.

## Before

Screenshot 2020-11-24 at 16 10 29

Screenshot 2020-11-24 at 16 10 40

## After

Screenshot 2020-11-24 at 17 21 20

Screenshot 2020-11-24 at 16 12 03

### DIFF
--- a/app/views/email_alert_subscriptions/new.html.erb
+++ b/app/views/email_alert_subscriptions/new.html.erb
@@ -1,6 +1,7 @@
-<% content_for :title, @signup_presenter.page_title %>
-<% content_for :head do %>
-  <%= render 'email_alert_subscription_meta', email_alert_subscription: @signup_presenter %>
+<% if @signup_presenter.can_modify_choices? %>
+  <% content_for :title, @signup_presenter.page_title %>
+<% else %>
+  <% content_for :title, t("email_alert_subscriptions.new.title") %>
 <% end %>
 
 <% if @signup_presenter.beta? %>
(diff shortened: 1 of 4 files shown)


app/views/email_alert_subscriptions/new.html.erb
```diff
@@ -9,38 +10,43 @@
...
+      <%= render partial: 'govuk_publishing_components/components/title', locals: {
+        title: t("email_alert_subscriptions.new.title"),
+      } %>
+
+      <p class="govuk-body">
+        <%= t("email_alert_subscriptions.new.description") %>
+      </p>
+
+      <p class="govuk-body govuk-!-margin-bottom-8">
+        <strong style="font-weight: bold"><%= @signup_presenter.name %></strong>
```
kevindew, line 40: Oh dear - is there an overly zealous CSS reset? 

There seems to be a govuk-frontend override that can be used for it (not sure if we expose it or not) of `govuk-!-font-weight-bold` which might look a bit more informed than an inline style.

Whatever decided, it'd be good to include some information about it in the commit message.

kevindew (commented):
Looks good - minor nit that you seem to have missed the link in your commit message.

app/views/email_alert_subscriptions/new.html.erb
benthorner: Oh yes, thanks for spotting. It was irritating to find `<strong>` isn't so in this app. Your suggestion does work and I agree it's a better indicator of intent, so I've used that instead, and fixed up the commit message (with links).

benthorner:
Thanks for taking a look @kevindew! I think this should be ready for another review.

Regarding the resets, I'm wondering if there's a good forum to raise this in. It's not clear that we can just remove it, as I don't think there's an easy way to tell what it's propping up, without some kind of visual testing.

Some options: raise an issue, ask about it in #govuk-frontenders, propose it as tech debt.

What do you think?

kevindew:
> Thanks for taking a look @kevindew! I think this should be ready for another review.
> 
> Regarding the resets, I'm wondering if there's a good forum to raise this in. It's not clear that we can just remove it, as I don't think there's an easy way to tell what it's propping up, without some kind of visual testing.
> 
> Some options: raise an issue, ask about it in #govuk-frontenders, propose it as tech debt.
> 
> What do you think?

I think asking #govuk-frontenders seems the best option there to get a gauge for how they consider this. Essentially this is just the use of a legacy technique. Showing my age I recall it being the vogue to nuke all browser styles and rebuild upon them as a way for cross browser compatibility, then that seemed to be replaced with a [normalize.css trend](https://necolas.github.io/normalize.css/), whereas now it seems to be the much more atomic class based approach that the design system uses. Which leaves me a little unclear as to what is seen as the best way to do something as simple as apply bold to a bit of text (feels a bit overkill to need a class, but seems to conflict with the atomic class based approach of the design system).

kevindew (approved):

app/views/email_alert_subscriptions/new.html.erb
gavinwye: There is an overide class in the [design system](https://design-system.service.gov.uk/styles/typography/#font-weight) that you can apply to the `<p>` to make text bold. 

It's done like this:
```
<p class="govuk-body govuk-!-font-weight-bold">govuk-!-font-weight-bold</p>
```

gavinwye:
Am I right in thinking that this is the second time I've seen this, and the change is the removal of hint text I can't see the previous commits so just want to be sure I'm giving the 👍 to the right thing?

app/views/email_alert_subscriptions/new.html.erb
benthorner: I agree we could just use a `<p>` element. However, I think `<strong>` is semantically clearer, and should be what we use for important content - it's just a shame our apps are working against us here.

benthorner:
@gavinwye yes, that's right - this is the latest version after making that change. I suspect there's nothing more to say.

app/views/email_alert_subscriptions/new.html.erb
gavinwye: Okay yea, but, can we apply the `govuk-!-font-weight-bold` override class to the `<strong>` element so we're consistent and get any changes in the future if we change the appearance of bold on GOV.UK.

benthorner: Oh yes, I already did that. Sorry, it's not very clear in the GitHub UI - you can see the top of the panel here has an "outdated" flag on it, which means you'll need to check in "Files changed" to see the update.

gavinwye: Perfect! GitHub's UI 🙄

PR #2332, opened 2020-12-04
Implement cross-domain tracking on accounts redirects

barrucadu:
Depends on: https://github.com/alphagov/govuk_publishing_components/pull/1811 & https://github.com/alphagov/govuk-account-manager-prototype/pull/478

---

There are some links to www.gov.uk which may redirect to the accounts
domain.  We want cross-domain tracking to work in the case where there
is a redirect, which means we need to pass along the cross-domain _ga
parameter.

Since these are links to www.gov.uk, they don't get decorated by the
normal cross-domain logic.  So instead, instantiate a module from
govuk_publishing_components which does it explicitly.

As well as decorating the links, we need to pass along the parameter
with any redirects.

---

[Trello card](https://trello.com/c/ISYymNsg/426-fix-cross-domain-tracking-on-redirects)


andysellick (commented):
Frontend looks good to me 👍  but can't comment on the backend, best to get another pair of eyes.

barrucadu:
I've now tested this and the account manager branch end-to-end on integration, and I ended up back on the checker with a `_ga` param after logging in

app/controllers/sessions_controller.rb
```diff
@@ -2,9 +2,9 @@ class SessionsController < ApplicationController
   before_action :check_accounts_enabled
 
   def create
-    redirect_to account_manager_url and return if logged_in?
+    redirect_with_ga account_manager_url and return if logged_in?
```
huwd, line 5: nice, I like the method reads really well

huwd (approved):
Looks great

PR #2353, opened 2020-12-17
Fix using subscription_url from email API

benthorner:
This was removed in [1], without updating this app to use the list
slug to form the subscription URL. While it's not ideal to have
shared knowledge of URLs, it's only the case for this app and Email
Alert Frontend, which provides the rest of the signup flow anyway.

Note that there is definitely scope for refactoring here to DRY-up
the repetitious stubbing in the unit and controller specs. However,
the priority for now is to fix the issue we're seeing.

[1]: https://github.com/alphagov/email-alert-api/pull/1517

### DIFF
--- a/app/lib/email_alert_signup_api.rb
+++ b/app/lib/email_alert_signup_api.rb
@@ -11,7 +11,7 @@ class EmailAlertSignupAPI
   end
 
   def signup_url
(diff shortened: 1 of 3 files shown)


benthorner:
Tested and works on Integration ✅

sihugh (approved):

PR #2354, opened 2020-12-17
Update email link component

alex-ju:
Update the email link component and use it consistently on the Brexit checker results page (with a variation on design).

[Trello card](https://trello.com/c/Q3jZKhC0)
[Design reference](https://trello.com/c/T233wdBp)

<table>
<tr><th>Before (production)</th><th>After (this branch)</th></tr>
<tr><td valign="top">

before

</td><td valign="top">

after

</td></tr>
</table>


config/locales/en/brexit_checker/results.yml
```diff
@@ -12,11 +12,12 @@ en:
       description_no_answers: |
         <p class="govuk-body-l">To change your answers, follow the link below.</p>
         <p class="govuk-body-l">This is a safe and secure service. We do not store any of your personal information.</p>
-      email_sign_up_title: What you need to do may change
+      email_sign_up_title: Stay up to date
+      email_sign_up_description: Sign up to get emails about Brexit changes that may affect you.
       email_sign_up_link: |
-        Subscribe to email updates about changes that may affect you and get a link to your results
+        Get email updates
       email_sign_up_link_no_actions: |
```
benthorner, line 19: We could remove this and [the corresponding helper](https://github.com/alphagov/finder-frontend/blob/27a74325d47f17a95f77126ab5817ee21a713201/app/helpers/brexit_checker_helper.rb#L51-L57), since the text is now invariant between the two scenarios. I think it would be confusing to leave it in, if someone came across this in future.

app/views/brexit_checker/results.html.erb
```diff
@@ -96,7 +96,20 @@
...
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+      <%= render 'components/email_link', {
+        data_attributes: {
+          "module": "track-click",
+          "track-action": action_based_email_link_label,
+          "track-category": "StayUpdated",
+          "track-label": transition_checker_email_signup_path
+        },
+        link_text: action_based_email_link_label,
```
benthorner, line 110: Note to self: this is consistent with the [other instance](https://github.com/alphagov/finder-frontend/pull/2354/commits/f0fc3568fa82e475ccf415918dc81779cb350bbb#diff-2ea979876c23d52fa27a5aa788b7a5dec0d25ff553e9751c11dd7d110627ca56R80) of the link.

benthorner (commented):
Looks good - I've tried it out locally too and it's all working for me. The commits made the changes easy to follow too 💯. I've just got one more major comment about the link text. Since you're not around, I'll check with @sihugh and @hannako whether to add that in myself. Otherwise, I think this is good to go 🚀.

config/locales/en/brexit_checker/results.yml
sihugh: Yep, I think this is OK to remove. It would be simple enough to re-add if we needed to 👍

benthorner: I've done that in a new commit: https://github.com/alphagov/finder-frontend/pull/2354/commits/845889464103e67f9234c44ff985943be35786d0.

benthorner (approved):

sihugh (approved):

benthorner:
Converting this to draft: on asking for a wider review of the UI change, it became clear the original intention was to apply this change in an A/B test, due to a concern that the more prominent CTA for email signup may deter users from actually looking at their results. While this initially made me wonder why we're making this change at all (therefore), I believe it's because a previous UI change to mark "urgent" actions was in conflict with email signups.

chris-gds:
👋🏼     So, I was looking at this UI element recently on these pages in order to swap this element to [align it with the gem that is used for email notifications](https://components.publishing.service.gov.uk/component-guide/subscription-links).

As the Design was significantly different this work was paused.   It seems as though this element is undergoing AB testing - is the goal for this Design change to be absorbed into the gem at some point?

image

PR #2357, opened 2020-12-18
A11y adjustments for the Brexit checker views

danacotoran:
The Brexit checker templates use the `finder_layout` layout file.
This layout file already adds a `<main>` wrapper around the content. 
However, some of the brexit checker views also contain a `<main>` wrapper, which means that on those pages there would be two `main` landmarks, which goes against the rule that the `<main>` element must be unique to the document.

In addition to this, these views also contain breadcrumbs. 
Since breadcrumbs are a navigation element, they should live outside of `<main>`. I updated these views to store the breadcrumbs using `content_for`. This way the breadcrumbs will be placed outside of `<main>` in the layout file.

### Before: two `main` landmarks, breadcrumbs sit inside of `main`
Screenshot 2020-12-18 at 14 28 17

### After: only one `main` landmark, breadcrumbs sit outside of `main`
Screenshot 2020-12-18 at 14 31 24

### DIFF
--- a/app/views/brexit_checker/email_signup.html.erb
+++ b/app/views/brexit_checker/email_signup.html.erb
@@ -6,26 +6,29 @@
   <%= account_variant.analytics_meta_tag.html_safe %>
 <% end %>
 
+<% content_for :breadcrumbs do %>
+  <div class="govuk-width-container">
+    <%= render 'govuk_publishing_components/components/breadcrumbs', {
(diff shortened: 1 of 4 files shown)


injms (approved):
Good find - thanks for fixing 👍

danacotoran:
Thank you for approving @injms, if you're happy with this would you mind merging it in for me as well?

PR #2358, opened 2020-12-18
Replace show results button with button component on mobile search page

owenatgov:
## What
Replaces the button to apply filters for mobile on [the search page](https://www.gov.uk/search/all) with a govuk publishing components [button component](https://components.publishing.service.gov.uk/component-guide/button).

## Why
This is part of a wider piece of work by the govuk accessibility team to ensure that as many of our apps as possible are using up to date, consistent markup. This can be achieved in part by ensuring we make use of our components tooling where possible.

No visual changes.

[Card](https://trello.com/c/9pRq4aIl/567-update-finder-frontend-to-use-form-components)

### DIFF
--- a/app/views/finders/_facet_collection.html.erb
+++ b/app/views/finders/_facet_collection.html.erb
@@ -57,9 +57,10 @@
           </button>
         </div>
         <div class="facets__footer">
-          <button class="gem-c-button govuk-button js-close-filters js-show-results" type="button">
-            Show <span class="js-result-count"> <%= result_set_presenter.displayed_total %></span>
-          </button>
+          <%= render "govuk_publishing_components/components/button", {
+            text: sanitize("Show <span class=\"js-result-count\">#{result_set_presenter.displayed_total}</span>"),
+            classes: "js-close-filters js-show-results"
+          } %>
         </div>
       </div>
     </div>


app/views/finders/_facet_collection.html.erb
```diff
@@ -56,10 +56,10 @@
             Clear all filters
           </button>
         </div>
-        <div class="facets__footer">
-          <button class="gem-c-button govuk-button js-close-filters js-show-results" type="button">
-            Show <span class="js-result-count"> <%= result_set_presenter.displayed_total %></span>
-          </button>
+        <div class="facets__footer js-close-filters js-show-results">
```
injms, line 59: Just wondering why the `js-` JavaScript hook class has moved from the `button` to the `div`?

owenatgov: I've moved this because there's no way to get the js class onto the button component on line 60. The js for mobile filters is reliant on that class or at least some sort of hook into the page so I've amended it slightly so that a div containing the generated button markup interacts with the js.

andysellick: Is this relevant? https://components.publishing.service.gov.uk/component-guide/button/with_js_classes

owenatgov: I missed that! Whoops, will sort out.

injms (approved):
Super!

PR #2360, opened 2020-12-21
Improve tests for EmailAlertSignupAPI class

benthorner:
Previously these were hard to maintain [1], due to the amount of
repetition and confusing structure. The commits in this PR are an
attempt to incrementally improve the tests.

[1]: https://github.com/alphagov/finder-frontend/pull/2353

### DIFF
--- a/spec/lib/email_alert_signup_api_spec.rb
+++ b/spec/lib/email_alert_signup_api_spec.rb
@@ -21,8 +21,6 @@ describe EmailAlertSignupAPI do
   let(:applied_filters) { {} }
   let(:facets) { [] }
   let(:subscriber_list_title) { "Subscriber list title" }
-  let(:slug) { "slug" }
-  let(:subscription_url) { "/email/subscriptions/new?topic_id=#{slug}" }
 
   describe "default_attributes" do
     context "no default_attributes or attributes" do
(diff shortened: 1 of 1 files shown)


hannako (approved):

spec/lib/email_alert_signup_api_spec.rb
```diff
@@ -13,103 +13,78 @@
...
+        req = stub_email_alert_api_has_subscriber_list(
+          "tags" => {},
+          "slug" => "slug",
+        )
 
-          expect(subject.signup_url).to eql subscription_url
-          assert_requested(req)
-        end
+        subscription_url = "/email/subscriptions/new?topic_id=slug"
+        expect(subject.signup_url).to eql subscription_url
```
hannako, line 33: does `signup_api_wrapper` == `subject`?

hannako (approved):

spec/lib/email_alert_signup_api_spec.rb
benthorner: Ooh, good spot - I've added another commit to remove those inconsistencies. Do you want to check it over?

hannako: Looks good!

benthorner:
Looking through again, I removed one of the commits around default values, as it wasn't actually redundant: the presence of default values was being tested implicitly, and the test pointed to in the commit was only focussing on a variant of that behaviour. Default value behaviour is still only tested in specific contexts, but it's no less tested than previously.

PR #2368, opened 2020-12-30
Update template for Brexit checker emails

benthorner:
Depends on: https://github.com/alphagov/email-alert-api/pull/1538

Please see the commits for more details.

## Before

Screenshot 2020-12-30 at 17 06 25

## After

Screenshot 2020-12-31 at 10 33 36

### DIFF
--- a/app/views/brexit_checker_mailer/change_notification.text.erb
+++ b/app/views/brexit_checker_mailer/change_notification.text.erb
@@ -1,4 +1,4 @@
-<%= t("brexit_checker_mailer.change_notification.#{@notification.type}") %>
+# <%= t("brexit_checker_mailer.change_notification.#{@notification.type}") %>
 
 <% if @action.title_url.present? %>
 [<%= @action.title %>](<%= notification_email_link(@action.title_url, @notification) %>)
--- a/app/views/brexit_checker_mailer/change_notification.text.erb
+++ b/app/views/brexit_checker_mailer/change_notification.text.erb
@@ -11,14 +11,14 @@
 <% if @action.exception.present? %><%= @action.exception %><% end %>
 
 <% if @action.guidance_url.present? %>
-<%= @action.guidance_prompt || t("brexit_checker_mailer.change_notification.guidance_prompt") %>
+<%= @action.guidance_prompt || t("brexit_checker_mailer.change_notification.guidance_prompt") %>:
 [<%= @action.guidance_link_text %>](<%= notification_email_link(@action.guidance_url, @notification)  %>)
 <% end %>
 
-Updated
+Updated:
 <%= Date.parse(@notification.date).strftime("%-d %B %Y") %>
 
 <% if @notification.note.present? %>
-Changes made
+Changes made:
 <%= @notification.note %>
 <% end %>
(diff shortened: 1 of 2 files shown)


app/views/brexit_checker_mailer/change_notification.text.erb
```diff
@@ -11,14 +11,14 @@
 <% if @action.exception.present? %><%= @action.exception %><% end %>
 
 <% if @action.guidance_url.present? %>
-<%= @action.guidance_prompt || t("brexit_checker_mailer.change_notification.guidance_prompt") %>
```
huwd, line 14: What's the thinking behind doing this in the template instead of [here in the yml](https://github.com/alphagov/finder-frontend/blob/bc6dbbc959a9f0c18401354d0f8614a6650ac34c/config/locales/en/brexit_checker_mailer/change_notification.yml#L4)

huwd: :thinking: we do seem to do the same thing here:
https://github.com/alphagov/finder-frontend/blob/d881befeb68af24720ec35af92484c9dc089313c/app/views/brexit_checker/_action_list.html.erb#L52

Is there anywhere we're using this without the `:`?

app/views/brexit_checker_mailer/change_notification.text.erb
```diff
@@ -1,4 +1,4 @@
-<%= t("brexit_checker_mailer.change_notification.#{@notification.type}") %>
+## <%= t("brexit_checker_mailer.change_notification.#{@notification.type}") %>
```
huwd, line 1: This is markdown in an erb right?
The notify api will take this and turn it into an `h2`...
... and it's not an ERB comment?

(just thought i'd check, confusing with the file names!)

huwd (approved):

huwd:
Two questions, neither blocking so have approved!

app/views/brexit_checker_mailer/change_notification.text.erb
benthorner: Yes e.g. [here](https://github.com/alphagov/finder-frontend/blob/f55dfc4bcb5e29926629e503ebd4c42b720f937e/app/lib/brexit_checker/actions.yaml#L12). It seems to be pretty unnecessary, but unfortunately we don't have full control of the content here. Looking over the file, it always makes sense to add the colon after, but would be impractical to try and add it consistently in `actions.yaml` / the-mighty-spreadsheet.

app/views/brexit_checker_mailer/change_notification.text.erb
benthorner: Yes, that's right. Apparently an ERB comment is like `<%#`, although I've never used them myself.

(I need to force-generate an email to double-check the new formatting.)

benthorner:
Adding "DNM" until I can force-generate an email with the new template, to put in the description and prove it's working as expected.

app/views/brexit_checker_mailer/change_notification.text.erb
huwd: Ace thanks for confirming!

benthorner:
Code to generate test email (after subscribing email-alert-api-staging@ to checker results for `nationality-uk`):

```
notification = BrexitChecker::Notification.new(id: SecureRandom.uuid, action_id: "S039", type: "addition", date: "2020-12-03")
payload = BrexitChecker::Notifications::Payload.new(notification).to_h
GdsApi.email_alert_api.create_message(payload)
```

PR #2369, opened 2020-12-31
Help the checker fail gracefully if accounts returns 503

huwd:

Trello: https://trello.com/c/LaSdrZyI/459-feature-turn-off-accounts

## What

Expands the definition of `account_enabled?` to be BOTH that we have the feature flag in the env and that accounts returns a status code that is not 503

## Why

Whilst accounts being "down" could see it return a number of codes, we're specifically interested in 503 here. We have an emergency brake feature that would allow an accounts dev or on call / 2ndline dev to suspend all of accounts in a hurry and see all traffic remapped to a service unavailable page where we can give more information to the user.

When this was tested on staging we saw that the checker became very sad with this new state of affairs.

In the case of an intentional but temporary withdrawal of the accounts service, we want the checker to continue to work but revert back to old behaviour just offering email subscriptions and obscuring all accounts related features.


brucebolt:
I like the idea of checking the accounts application is available prior to showing the user anything about accounts, however I think this can probably be done without extra dependencies.

We already have a `/healthcheck` endpoint in the account manager which returns JSON with a `status` field.  We could utilise this, which would cover other scenarios, not just a 503 being returned.

To avoid including a dependency, we could extend `gds-api-adapters` within this application (e.g. as we did for Content Publisher to communicate with Whitehall: https://github.com/alphagov/content-publisher/blob/master/lib/gds_api/whitehall_export.rb) using the `get_json` method to get the content of the account manager healthcheck instead of using HTTParty.

huwd:
My slack is down so I'm not sure direct messages are getting through,
But yearh fair point - the caveat is how the healthcheck endpoints would work with the status error page.

In the circumsatnce where we need this the most, it is likely that we will have remapped the account to the error pages.
I've got a WIP branch here to switch from a cf staticfile buildpack to nginx, but am having some fun with getting cf to build:
https://github.com/alphagov/govuk-account-static-errors/pull/6

Are you thinking I should add a `/healthcheck` end point to there, instead of relying on root?

barrucadu:
I'm unsure about this approach.  If the accounts apps are returning a high rate of 5xx errors, but not for *every* request, we could get into a situation where this new check passes, but subsequent requests fail.  So we'll still need to be resilient to 5xx errors even if `accounts_enabled? == true`.

app/controllers/application_controller.rb
```diff
@@ -90,7 +90,7 @@ def check_accounts_enabled
 
   helper_method :accounts_enabled?
   def accounts_enabled?
-    Rails.configuration.feature_flag_govuk_accounts
+    Rails.configuration.feature_flag_govuk_accounts && RestClient.get(Services.accounts_api).code != 503
```
barrucadu, line 93: Does this need some exception handling?

https://rubydoc.info/gems/rest-client/

> - for result codes between 200 and 207, a RestClient::Response will be returned
> - for result codes 301, 302 or 307, the redirection will be followed if the request is a GET or a HEAD
> - for result code 303, the redirection will be followed and the request transformed into a GET
> - for other cases, a RestClient::ExceptionWithResponse holding the Response will be raised; a specific exception class will be thrown for known error codes

huwd:
Worked out the last of these comments on slack with Bruce / Micheal. We decided:
- The healthcheck endpoint isn't useful for us here, healthchecks return 200 status codes instead of our desired 503. We also don't want to use the status, because whilst it is `critical` there might be other kinds of `critical` status that necessitate a different response.
- For now this is _just_ about responding to 503s when if we ever purposfully take down the account making the service temporarily but conciously unavailable. There may be other resiliance features we want to look at in future for other status codes.
- Hopes that we could use GDS API adaptors were dashed by it's inistance on only using json_get queries. We don't hvae an easy JSON endpoint here, and just want a status code.
- However HTTParty is new and doesn't seem widely used on GOV.UK whereas RestClient is used in a few of the team's other apps, and is already a sub depndency of "GDS API" Adaptors. So if we use this instead no additional deps get installed 🎉

huwd:
Ah alas... RestClient handles HTTP 5xx [as errors](https://github.com/rest-client/rest-client#exceptions-see-httpwwww3orgprotocolsrfc2616rfc2616-sec10html)...

Error handling increases the verbosity of the code quite a bit.
It's also not really an error for us here, here in some ways 503 sucessfully confirms to finder-frontend that the account isn't available and it should adopt a plan B...

I wonder if this ends up being an argument for just using HTTParty again...
🤔 Back to the drawing board a bit

app/controllers/application_controller.rb
```diff
@@ -90,7 +90,17 @@ def check_accounts_enabled
...
-    Rails.configuration.feature_flag_govuk_accounts
+    return false unless Rails.configuration.feature_flag_govuk_accounts
+
+    if @check_accounts_enabled.nil?
+      @check_accounts_enabled = true
+      begin
+        RestClient.get(Services.accounts_api)
+      rescue RestClient::ServiceUnavailable
+        @check_accounts_enabled = false
+      end
```
barrucadu, line 101: Hmm, I guess this will also throw other errors too, I think we'll also need a `rescue StandardError` clause which doesn't do anything, and it'd be good to have a test that some other 5xx error (like just a plain 500) results in `accounts_enabled? == true`

huwd: Righto that makes sense, I'll do that now!

huwd: https://github.com/alphagov/finder-frontend/pull/2369/commits/a94264ac25243f699520efe6b4931e43e0f9c604

barrucadu (approved):

barrucadu:
It'd be good to monitor the performance impact of this change for a bit

PR #2389, opened 2021-01-12
Fix checker/accounts paths redirecting for suspended account

huwd:

[Trello](https://trello.com/c/LaSdrZyI/459-feature-turn-off-accounts)

https://github.com/alphagov/finder-frontend/pull/2369 focused on what the user could see on the results page, so effectively which journies the user could start after a point where we suspended accounts.

Turned out there were a number of account specific paths that would still be in place if we suspended the account.

We want to be in a position where:
- if we remove the feature flag those paths all correctly 404
- if a user is already on a journey which features one of those paths (so they probably started before we suspended accounts) their journey is halted and we redirect them to the accounts service URL which will, by then be showing a 503 error message explaining why accounts has gone away.

This is part of a wider piece of work to allow us to enable a planned and temporary suspension of accounts should we ever need to for maintainace or in response to an incident.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -97,6 +97,7 @@ private
     accounts_available?
   end
 
+  helper_method :account_feature_flag_enabled?
   def account_feature_flag_enabled?
     Rails.configuration.feature_flag_govuk_accounts
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -83,15 +83,25 @@ private
   end
 
   def check_accounts_enabled
-    unless accounts_enabled?
+    if account_feature_flag_enabled?
+      redirect_to Services.accounts_api unless accounts_available?
+    else
       render file: Rails.root.join(Rails.root, "public/404.html"), status: :not_found
     end
   end
 
   helper_method :accounts_enabled?
   def accounts_enabled?
-    return false unless Rails.configuration.feature_flag_govuk_accounts
+    return false unless account_feature_flag_enabled?
 
+    accounts_available?
+  end
+
+  def account_feature_flag_enabled?
+    Rails.configuration.feature_flag_govuk_accounts
+  end
+
+  def accounts_available?
     if @check_accounts_available.nil?
       @check_accounts_available = true
       begin
--- a/spec/features/brexit_checker/accounts_spec.rb
+++ b/spec/features/brexit_checker/accounts_spec.rb
@@ -8,40 +8,6 @@ RSpec.feature "Brexit Checker accounts", type: :feature do
     stub_request(:get, Services.accounts_api).to_return(status: 200)
   end
 
-  context "without accounts enabled" do
-    let(:mock_results) { %w[nationality-eu] }
-
-    context "/transition-check/saved-results" do
-      it "returns a 404" do
-        given_i_am_on_the_saved_results_page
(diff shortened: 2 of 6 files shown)


spec/requests/account_suspended_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+RSpec.describe "When account is enabled", type: :request do
+  before do
+    allow(Rails.configuration).to receive(:feature_flag_govuk_accounts).and_return(true)
+  end
+
+  context "and accounts is returning a 503 error code" do
+    before { stub_request(:get, Services.accounts_api).to_return(status: 503) }
+
+    describe "/login" do
+      it "redirects a user to the accounts 503 error page" do
```
huwd, line 12: I was considering just metaprogramming these as nearly all the tests are the same, but I wasn't sure where we stood on metaprogramming tests as a team?

I could just shove all these routes in an array and iterate over them if folks prefer that?

spec/requests/account_suspended_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+        expect(response).to redirect_to(Services.accounts_api)
+      end
+    end
+
+    describe "/logout" do
+      it "redirects a user to the accounts 503 error page" do
+        get transition_checker_end_session_path
+
+        expect(response.body).to eq("Redirecting to #{Services.accounts_api}/sign-out?continue=1")
```
huwd, line 31: i suspect the parameters here are much a muchness, 
but i've made a special case of `/logout`, I figure in that scenario we _do_ want to delete the session cookie.

But then vary based on where the user ends up. If it's accounts, they end up back on the 503 error page, and the params are ignored.

if it's transition, then fine they can just end up back there.

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -8,40 +8,6 @@
     stub_request(:get, Services.accounts_api).to_return(status: 200)
   end
 
-  context "without accounts enabled" do
-    let(:mock_results) { %w[nationality-eu] }
-
-    context "/transition-check/saved-results" do
```
huwd, line 14: I figured these might work better as request specs now.
We can drop the mock results and just focus on what happens if you send an HTTP request and the expected response?

app/controllers/application_controller.rb
```diff
@@ -83,15 +83,25 @@ def filter_params
   end
 
   def check_accounts_enabled
-    unless accounts_enabled?
+    if Rails.configuration.feature_flag_govuk_accounts
```
barrucadu, line 86: Use `account_feature_flag_enabled?` here too for consistency

barrucadu:
Has this been tested (on staging or integration) with accounts suspended?

app/controllers/application_controller.rb
huwd: quite right, that should have been there!

huwd: amended here: https://github.com/alphagov/finder-frontend/pull/2389/commits/4631afc5bb629ccd7149055b8a387fcf7d9a1785

huwd:
Not yet, but I'm happy to do that before merging.
You happy with the code otherwise?

barrucadu:
Yes, I'm happy with it if it does gracefully fail

huwd:
Cool tested on staging, looks good!

barrucadu (approved):

PR #2404, opened 2021-01-20
Add a rake task to import actions by batch id

hannako:
- If we add a column to the dynamic spreadsheet for batch id, we could import
by batch id.
- This would write to a new yaml file in /batched_actions.
- We would then manually copy those batched_actions into actions.yaml.
- Would just be a temporary thing whilst there are so many updates flying around.


hannako:
Closing as feature no longer required

PR #2407, opened 2021-01-21
Move account logic out of the brexit controller & into concerns

barrucadu:
We've got a lot of logic in the ApplicationController (ew) and the BrexitCheckerController.  We can make those controllers simpler and keep the shared logic out of ApplicationController by moving stuff into concerns.

The controller methods are left behind, but most of the control flow and error handling is gone, making the controller itself easier to read.


huwd (approved):
Looks good!

PR #2435, opened 2021-02-19
Remove caching / refreshing of access token in JWT flow 

barrucadu:
If we end up with a bad refresh token cached currently we will not
regenerate the tokens at all, and just return an error.  Since
generating tokens is fast, just don't cache.


AlanGabbianelli (approved):
Sensible 👍

PR #2458, opened 2021-03-12
Switch to the account-api email subscription endpoints

barrucadu:
These endpoints don't have any use outside of the transition checker,
but by migrating the functionality to account-api we're one step
closer to being able to treat the GOVUK-Account-Session as a black box
in this app.

---

[Trello card](https://trello.com/c/N49KKdOW/666-move-transition-checker-email-management-from-finder-frontend-to-the-new-app)


huwd (approved):
I love a commit where tests pass and there's more removed than added :D Nice one.

PR #2459, opened 2021-03-12
Switch to the account-api attribute endpoints & remove OAuth / OIDC code

barrucadu:
All the token wrangling is gone, the Transition Checker is now doing personalisation the way we want all GOV.UK apps to!

---

[Trello card](https://trello.com/c/uRMyiEHZ/654-move-oauth-attribute-logic-from-the-transition-checker-to-the-new-app)


app/controllers/concerns/account_brexit_checker_concern.rb
```diff
@@ -5,6 +5,8 @@ module AccountBrexitCheckerConcern
 
   ACCOUNT_ACTIONS = %i[save_results save_results_sign_up save_results_confirm save_results_email_signup save_results_apply saved_results edit_saved_results].freeze
 
+  ATTRIBUTE_NAME = "transition_checker_state"
```
huwd, line 8: This is interesting, why put this in a constant if it's just a string?
Are you thinking this is a concern we might copy paste around, or is this us shooting for better labelling in the code?

huwd (approved):

app/controllers/concerns/account_brexit_checker_concern.rb
barrucadu: It was to avoid specifying it twice... though I've just noticed I only actually used the constant in one of the places.

PR #2472, opened 2021-03-24
Delegate sessions to frontend

huwd:
[Trello](https://trello.com/c/43FM4pwv/673-switch-to-the-new-sign-in-out-endpoints)

We can now simplify down and remove a good deal of code from finder-frontend.


huwd:
🤔 what's going on with e2e...
Will kick those off again

app/controllers/brexit_checker_controller.rb
```diff
@@ -12,7 +13,7 @@ class BrexitCheckerController < ApplicationController
   ]
 
   before_action :enable_caching, only: %i[show email_signup confirm_email_signup]
-  before_action :enable_caching_unless_accounts, only: %i[results]
+  before_action :enable_caching_unless_logged_in, only: %i[results]
```
barrucadu, line 16: We can actually just enable caching now, as we're setting the `Vary` response header in the `set_account_variant` action.  So one user's results page won't be cached for another.

config/routes.rb
```diff
@@ -1,6 +1,9 @@
 FinderFrontend::Application.routes.draw do
   mount GovukPublishingComponents::Engine, at: "/component-guide"
 
+  get "/sign-in", to: proc { [200, {}, %w[OK]] }
+  get "/sign-out", to: proc { [200, {}, %w[OK]] }
```
barrucadu, line 5: What are these for?

app/lib/services.rb
```diff
@@ -4,6 +4,10 @@
 require "gds_api/email_alert_api"
 
 module Services
+  def self.frontend
+    Plek.find("frontend")
+  end
```
barrucadu, line 9: I think just using `Plek.find("frontend")` in the concern would be clearer.

app/controllers/concerns/session_path_concern.rb
```diff
@@ -0,0 +1,29 @@
...
+    "#{Services.frontend}/#{transition_checker_new_session_path(**params)}"
+  end
+
+  def transition_checker_new_session_callback_url(**params)
+    "#{Services.frontend}/#{transition_checker_new_session_path(**params)}"
+  end
+
+  def transition_checker_end_session_url(**params)
+    "#{Services.frontend}/#{transition_checker_new_session_path(**params)}"
+  end
```
barrucadu, line 28: These `_path`s aren't used outside of this concern, so I think they should be inlined into the `_url`s.  Also you've used the `transition_checker_new_session_path` in all of the `_url`s.

app/controllers/concerns/session_path_concern.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module SessionPathConcern
+  extend ActiveSupport::Concern
```
barrucadu, line 4: This doesn't need to extend `ActiveSupport::Concern` because it's not implementing any concern functionality.

Should these methods actually just be in the `AccountConcern`?

app/controllers/brexit_checker_controller.rb
huwd: ah ace, I did wonder about this bit.
Cool will add this to the pile with the others

app/lib/services.rb
huwd: That seemed inconsistent with what we've done before.
BUT! looking again I see that's because with the other services we're either doing initalisation calls or using require to bring other stuff in.

This does neither so makes sense, will amend

app/controllers/concerns/session_path_concern.rb
huwd: yep fair, I'll remove those, I was thinking I might need those, but guess I didn't!

app/controllers/concerns/session_path_concern.rb
huwd: The reason I didn't put them in the account concern was so that they can sit apart and be included as a separate module.
The account concern has before filters and so freaks out if you put it in tests for instance.

Perhaps this means it's not really a concern and should be a Plain Old Ruby Module somewhere, but i wasn't sure where that might be...

And I figured for controllers a concern was the place you'd put something that hijacked routes like this?

config/routes.rb
huwd: From what I found if you remove or comment these out then we cannot run feature tests,
The routes at the bottom of this file seem to do a very generous wildcard match (I'm not sure why I guess there's an open ended number of finder slugs?).

But without this proc it things `/sign-in` is a finder and treats it as a content item finder needs to process.
These stub those out, but if you typed the real URL in on prod, the router takes over and throws you to `frontend`.

I couldn't figure out a better way to test these and still get routes helpers.
I tried `direct` too and none, but no joy.

Open to better ideas?

barrucadu: Can we make the tests not follow the redirects?  I don't like the idea of adding dummy `/sign-in` and `/sign-out` routes to every frontend app which uses accounts.

app/controllers/brexit_checker_controller.rb
```diff
@@ -2,6 +2,7 @@ class BrexitCheckerController < ApplicationController
   include AccountConcern
   include AccountBrexitCheckerConcern
```
barrucadu, line 3: Could we merge the `AccountConcern` and the `AccountBrexitCheckerConcern`?  They were separate because not everything the checker needed was useful for the sessions controller, but that controller is gone now.

app/controllers/concerns/account_brexit_checker_concern.rb
```diff
@@ -42,11 +40,11 @@ def pre_update_results
   end
 
   def logged_out_pre_saved_results_path
-    transition_checker_new_session_path(redirect_path: transition_checker_saved_results_path, _ga: params[:_ga])
+    transition_checker_new_session_url(redirect_path: transition_checker_saved_results_path, _ga: params[:_ga])
   end
 
   def logged_out_pre_update_results_path
-    transition_checker_new_session_path(redirect_path: transition_checker_save_results_confirm_path(c: criteria_keys), _ga: params[:_ga])
+    transition_checker_new_session_url(redirect_path: transition_checker_save_results_confirm_path(c: criteria_keys), _ga: params[:_ga])
```
barrucadu, line 47: These methods only work if `SessionPathConcern` is also included in the controller.  This isn't great, because it means we have two separate modules which don't really work independently.  So I still think the `SessionPathConcern` should be merged into the existing concern(s).

If nothing else, having three separate concerns, all implementing different pieces of the same thing, and which are only used in a single controller, is a bit of a code smell.

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -3,262 +3,258 @@
...
 
-        @original_account_session_header = page.response_headers["GOVUK-Account-Session"]
-        expect(@original_account_session_header).to_not be_nil
-      end
+      stub_sign_in_callback_url(headers, state: "state", code: "code")
 
-      def log_out
-        visit transition_checker_end_session_path(done: "1")
-        expect(page.response_headers["GOVUK-Account-End-Session"]).to_not be_nil
-      end
```
barrucadu, line 222: So, it's a bit hard to read with all the diff going on, but I don't think we need these `log_in` and `log_out` methods any more (and so we can remove two of the stubs).  I think the `before` and `after` blocks which call them can be:

```ruby
    before do
      @original_headers = page.driver.options[:headers]
      @original_account_session_header = { "GOVUK-Account-Session" => "placeholder-session" }
      page.driver.options[:headers] ||= {}
      page.driver.options[:headers].merge!("GOVUK-Account-Session" => @original_account_session_header)
    end

    after do
      page.driver.options[:headers] = @original_headers
    end
```

spec/support/govuk_account_helper.rb
```diff
@@ -0,0 +1,18 @@
+module GovukAccountHelper
+  include SessionPathConcern
+
+  def stub_sign_in_url(**params)
+    stub_request(:get, transition_checker_new_session_url(params))
+      .to_return(status: 200)
+  end
+
+  def stub_sign_in_callback_url(headers = {}, **params)
```
barrucadu, line 9: This isn't needed with my suggested change

spec/support/govuk_account_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+    stub_request(:get, transition_checker_new_session_callback_url(params))
+      .to_return(status: 200, headers: headers)
+  end
+
+  def stub_sign_out_url(**params)
```
barrucadu, line 14: This isn't used anywhere

spec/support/govuk_account_helper.rb
```diff
@@ -0,0 +1,18 @@
+module GovukAccountHelper
+  include SessionPathConcern
+
+  def stub_sign_in_url(**params)
```
barrucadu, line 4: If the main motivation to have the paths be in a separate concern was so that they could be included in the tests, well, there's only one path now.

app/controllers/concerns/session_path_concern.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module SessionPathConcern
+  def transition_checker_new_session_url(**params)
+    "#{Plek.find('frontend')}/sign-in?#{params.compact.to_query}"
+  end
+
+  def transition_checker_new_session_callback_url(**params)
```
barrucadu, line 8: The transition checker shouldn't be calling this URL.  The users browser gets redirected to it from the account-manager, and it redirects the user back to the transition checker.

spec/support/govuk_account_helper.rb
```diff
@@ -1,20 +1,18 @@
 module GovukAccountHelper
   include SessionPathConcern
 
-  FRONTEND_BASE_URL = Plek.find("frontend").to_s.freeze
-
   def stub_sign_in_url(**params)
-    stub_request(:get, FRONTEND_BASE_URL + "sign-in" + params.to_query)
+    stub_request(:get, transition_checker_new_session_url(params))
```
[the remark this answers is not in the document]
huwd: sorry this turns out was stuck in PR Review comment limbo from yesterday instead of being a comment.
it's a couple day old!

app/controllers/brexit_checker_controller.rb
huwd: done here: https://github.com/alphagov/finder-frontend/pull/2472/commits/1d011a03b21c778cb843c83841048e1af146c22f

app/controllers/concerns/account_brexit_checker_concern.rb
huwd: You make a compelling case! I have merged them.

config/routes.rb
barrucadu, line 5: Could these be moved next to the `get "/email/subscriptions/new", to: proc { [200, {}, [""]] }, as: :email_alert_frontend_signup`, and with a comment above them saying something like `# Needed for the brexit checker tests`?

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -3,262 +3,242 @@
...
-
-    context "the user is not logged in" do
-      context "/transition-check/results" do
-        it "shows the normal call-to-action" do
-          given_i_am_on_the_results_page
-          expect(page).to have_content(I18n.t("brexit_checker.results.email_sign_up_title"))
-        end
+    context "/transition-check/saved-results" do
+      it "redirects to login page" do
+        stub_sign_in_url(redirect_path: "/transition-check/saved-results")
```
barrucadu, line 22: Is this `stub_sign_in_url` needed?  It should be the user's browser which gets sent to the sign-in URL, finder-frontend shouldn't make a request to it.

Since this is the only test which uses `stub_sign_in_url`, if it's not needed the stub can be deleted, and it's probably better to just put `transition_checker_new_session_url` in this file rather than have a one-method helper.

huwd: humm perhaps not... tests pass without it!

app/controllers/concerns/account_concern.rb
```diff
@@ -104,4 +78,72 @@ def logout!
...
+    nil
+  end
+
+  def transition_checker_new_session_url(**params)
+    "#{Plek.find('frontend')}/sign-in?#{params.compact.to_query}"
+  end
+
+  def transition_checker_end_session_url(**params)
+    "#{Plek.find('frontend')}/sign-out?#{params.compact.to_query}"
+  end
```
barrucadu, line 148: I've realised that we don't actually want `Plek.find("frontend")` *in production*, as that gives `"https://frontend.production.govuk-internal.digital"`.  But we *do* want it in development, or the links won't work.

So I think we want `Rails.env.production? ? Plek.new.website_root : Plek.find("frontend")`

huwd: I thought all the plek env vars on prod would prevent that?
Isn't that sort of flexibility across environments the point of Plek?

barrucadu: Hmm, `Plek.new.website_root` would actually work for dev, *if* you have a running router.  Which is admittedly usually not the case, but maybe it's good enough.  And it avoids the conditional here.

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -6,259 +6,237 @@
...
-          stub_account_api_get_sign_in_url(redirect_path: "/transition-check/saved-results")
-          given_i_am_on_the_saved_results_page
-          expect(current_path).to eq(transition_checker_new_session_path)
-        end
+    context "/transition-check/edit-saved-results" do
+      it "redirects to login page" do
+        stub_account_api_get_sign_in_url(redirect_path: "/transition-check/saved-results")
+        given_i_am_on_the_edit_saved_results_page
+        expect(page).to have_current_path(
+          "#{Plek.find('frontend')}/sign-in?#{CGI.escape('redirect_path=/transition-check/saved-results')}",
```
barrucadu, line 33: Shouldn't the querystring here be `redirect_path=/transition-check/edit-saved-results`?

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -6,259 +6,237 @@
...
 
-      context "/transition-check/saved-results" do
-        it "redirects to login page" do
-          stub_account_api_get_sign_in_url(redirect_path: "/transition-check/saved-results")
-          given_i_am_on_the_saved_results_page
-          expect(current_path).to eq(transition_checker_new_session_path)
-        end
+    context "/transition-check/edit-saved-results" do
+      it "redirects to login page" do
+        stub_account_api_get_sign_in_url(redirect_path: "/transition-check/saved-results")
```
barrucadu, line 30: finder-frontend isn't calling account-api to sign in, so it doesn't need this stub.

spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -6,259 +6,237 @@
...
 
-      context "/transition-check/edit-saved-results" do
-        it "redirects to login page" do
-          stub_account_api_get_sign_in_url(redirect_path: "/transition-check/saved-results")
-          given_i_am_on_the_edit_saved_results_page
-          expect(current_path).to eq(transition_checker_new_session_path)
-        end
+    context "/transition-check/save-your-results/confirm" do
+      it "redirects to login page" do
+        stub_account_api_get_sign_in_url(redirect_path: "/transition-check/save-your-results/confirm?c%5B%5D=nationality-eu")
```
barrucadu, line 40: Or this stub

barrucadu (approved):

barrucadu (approved):
Looks good to me

PR #2478, opened 2021-04-06
Correct header name

huwd:
[Trello](https://trello.com/c/6GWYTVMM/681-remove-the-redundant-header-logic-in-finder-frontend-and-account-api)

Now we know more about how rails changes request headers, (see work [we did over on frontend](https://github.com/alphagov/frontend/blob/master/app/controllers/concerns/account_concern.rb#L7)),  we can simplify this code and bring it more into line with other account-using codebases.

### DIFF
--- a/spec/features/brexit_checker/accounts_spec.rb
+++ b/spec/features/brexit_checker/accounts_spec.rb
@@ -45,16 +45,8 @@ RSpec.feature "Brexit Checker accounts", type: :feature do
   end
 
   context "the user is logged in" do
-    before do
-      @original_headers = page.driver.options[:headers]
-      @original_account_session_header = "placeholder-session"
-      page.driver.options[:headers] ||= {}
-      page.driver.options[:headers].merge!("GOVUK-Account-Session" => @original_account_session_header)
-    end
-
-    after do
-      page.driver.options[:headers] = @original_headers
-    end
+    before { page.driver.header("GOVUK-Account-Session", "placeholder") }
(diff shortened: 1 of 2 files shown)


spec/features/brexit_checker/accounts_spec.rb
```diff
@@ -49,7 +49,7 @@
       @original_headers = page.driver.options[:headers]
       @original_account_session_header = "placeholder-session"
       page.driver.options[:headers] ||= {}
-      page.driver.options[:headers].merge!("GOVUK-Account-Session" => @original_account_session_header)
+      page.driver.options[:headers].merge!("HTTP_GOVUK_ACCOUNT_SESSION" => @original_account_session_header)
```
barrucadu, line 52: No, the request header sent to Rails is `GOVUK-Account-Session`.  The name-munging happens inside the Rails middleware.

See the frontend tests: https://github.com/alphagov/frontend/blob/9ce9aaf8248715c0fac8df87a293232c4e48b4da/test/support/session_helpers.rb#L3 https://github.com/alphagov/frontend/blob/master/test/integration/sessions_test.rb#L9

barrucadu (approved):

spec/features/brexit_checker/accounts_spec.rb
barrucadu: Discussed offline: this actually had to be `HTTP_GOVUK_ACCOUNT_SESSION` because `page.driver.options[:headers]` doesn't mangle the header names... but we should have been using `page.driver.header` anyway.

PR #2490, opened 2021-04-26
Remove legacy /healthcheck route

barrucadu:
govuk-puppet and govuk-aws have now been updated to only use the new
routes, so the old one is no longer needed.

See RFC 141 for more information.

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)


huwd (approved):
👍🏻

Issue #2446, opened 2021-03-01
Missing Filter for GOV.UK Search


nnagewad:
# Note

- I've noticed a small bug on search where the `Search` disappears complete when changing the Browser's window from desktop to mobile, back to desktop.

## Video

https://user-images.githubusercontent.com/8485154/109534108-e33b1700-7ab2-11eb-8c2d-a4523624522c.mov

andysellick:
Thanks for finding this @nnagewad. We'll take a look.

PR #2493, opened 2021-04-28
Fix minor filter bug when switching breakpoints

danacotoran:
Fixes the issue described in #2446 wherein the filters disappear on desktop when the browser window is resized from desktop to smaller breakpoints and back.

Instead of programatically applying inline styles to the facets, apply a class to toggle visibility on/off. This way we can ensure that the filters are always visible on desktop using CSS media queries.


huwd (approved):
LGTM

PR #2495, opened 2021-04-29
Do not consider someone with an empty session logged in

barrucadu:
We have a bug in our CDN config which always sets the
GOVUK-Account-Session request header, even if the cookie isn't set.
Since the `do_or_logout` method checks for nil-ness and not presence,
every request to the Transition Checker results page causes a request
to the account-api, which returns a 401.

Thankfully there's no bad side-effect of this, it just makes the page
a little slower and increases the volume of requests to account-api
unnecessarily.

I noticed that account-api was getting a lot of requests, far more
than we'd expect from the number of account holders we have, and
tracked it down to this.  I'll also update our CDN config to only set
the header if the cookie is set.


huwd (approved):
Great catch

PR #2507, opened 2021-05-11
Fix registry cache update schedule

sihugh:
At present all the schedules are run once per hour, but they're all clustered between the 45th and 60th minute. It'd be nicer if they were randomly spread out. It's not actually a problem, it just looks a bit untidy on the graphs.

The `every 45.minutes do` line translates into https://crontab.guru/#45_*_*_*_* which means "at the 45th minute", not "every 45 minutes".

https://grafana.production.govuk.digital/dashboard/file/finder-frontend.json?panelId=5&fullscreen&orgId=1&from=now-1d&to=now


hannako (approved):
What's the advantage of picking a random number of minutes? Rather than it always being 30 for example.

PR #2508, opened 2021-05-11
Fix cache warming schedule again

sihugh:
It turns out that Whenever does something weird when you provide a minutes argument that is less than or equal to 30. It assumes you mean 'every n minutes' whereas for values greater than 30, it assumes you mean 'the nth minute'.

[Whenever specs describing the behaviour](https://github.com/javan/whenever/blob/e916b58f29ea4da76de9ddc5aca8c85a5f29f897/test/unit/cron_test.rb#L17-L19)

So, when the random number generates an 8, whenever generates a schedule like this (edited for brevity):

> 8,16,24,32,40,48,56 * * * * /bin/bash -l -c 'rake registries:cache_refresh'

We'll switch to using raw cron syntax instead. [From the Readme](https://github.com/javan/whenever/blob/e916b58f29ea4da76de9ddc5aca8c85a5f29f897/README.md):

> every '0 0 27-31 * *' do
>  command "echo 'you can use raw cron syntax too'"
> end

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -8,8 +8,8 @@ set :output, error: "log/cron.error.log", standard: "log/cron.log"
 bundler_prefix = ENV.fetch("BUNDLER_PREFIX", "/usr/local/bin/govuk_setenv finder-frontend")
 job_type :rake, "cd :path && #{bundler_prefix} bundle exec rake :task :output"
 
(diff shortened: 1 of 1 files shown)


bilbof (approved):
So good 👍

hannako (commented):
🤯

PR #2516, opened 2021-05-20
Add new default option for Research & Statistics Finder

brucebolt:
The default option for the Research & Statistics Finder was to show only published statistics.

However, some organisations do not publish statistics, in which case users would see no results despite there being research available.

User research done by the Social Mobility Commission showed this was problematic for their users (as they do not publish statistics).

Therefore this adds a new default option that encompasses all documents that can be returned by this finder. The user can then filter the results down using the provided facets.

Whilst doing this, I noticed we still use the term `rummager_api` in the feature tests, despite the name Rummager being dropped several years ago.  I've renamed all affected methods so anyone unfamiliar with the term is not confused.

Trello card: https://trello.com/c/t4EKt4va

### DIFF
--- a/app/models/filters.rb
+++ b/app/models/filters.rb
@@ -23,15 +23,20 @@ module Filters
           "value" => "statistics_published",
           "label" => "Statistics (published)",
           "filter" => {
-            "content_store_document_type" => %w[statistics national_statistics statistical_data_set official_statistics],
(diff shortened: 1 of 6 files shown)


MuriloDalRi (approved):
I have to say I'm excited about the double glazing at Hogwarts

PR #2517, opened 2021-05-21
Remove jQuery from application.js and live_search.js

JonathanHallam:
Part of a wider effort to remove JQuery from GOV.UK

## application.js

This creates and calls the `live_search` code, originally by passing a jQuery object. This was a small change but it couldn't be done in isolation for this reason.

## live_search.js

Controls the JS on the search pages on gov.uk/search, and related search pages. Used jQuery's Ajax function to dynamically update sections of the page. This has been replaced with `XMLHttpRequest` to allow compatibility with as many browsers as possible. This has involved changing the code slightly in terms of the way callbacks were used to manage the page updates and manually serialising the form data. See commit messages for more specifics.

One part of this code still uses jQuery here: https://github.com/alphagov/finder-frontend/pull/2517/files#diff-fe1355439297aa459f871621b4f95627ea330655c7c6eb08ef10bda7ab66c0feR196 and needs separate work to enable a non-jQuery object to be passed to the modules start code.

## taxonomy_select

Heavily linked to the live_search code. Controls the topic/subtopic select boxes on some searches (e.g. https://www.gov.uk/search/news-and-communications).

## Jasmine tests and IE

Currently all of the `live_search` jasmine tests pass in IE10+, apart from two that create keypress events using a technique not supported by IE10. This can hopefully change to use the GOVUK triggerEvent code when that is updated to allow keyCodes to be passed to it, but this is not a blocking issue for this PR - fixing the tests in IE10+ is a separate issue (several other tests unrelated to this change are also failing).

Trello card: https://trello.com/c/jK6hrK3d/417-remove-jquery-from-finder-frontend

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -28,25 +28,21 @@
 //= require_tree ./modules
 //= require_tree ./components
 
-/* eslint-env jquery */
+var $elementsRequiringJavascript = document.querySelectorAll('.js-required')
 
-jQuery(function ($) {
-  var $form = $('.js-live-search-form')
+for (var i = 0; i < $elementsRequiringJavascript.length; i++) {
+  $elementsRequiringJavascript[i].style.display = 'block'
+}
 
-  var $results = $('.js-live-search-results-block')
+var $form = document.querySelector('.js-live-search-form')
+var $results = document.querySelector('.js-live-search-results-block')
+var $atomAutodiscoveryLink = document.querySelector("link[type='application/atom+xml']")
 
-  var $elementsRequiringJavascript = $('.js-required')
-
-  var $atomAutodiscoveryLink = $("link[type='application/atom+xml']").eq('0')
-
-  $elementsRequiringJavascript.show()
-
-  if ($form.length && $results.length) {
-    // eslint-disable-next-line
-    new GOVUK.LiveSearch({
-      $form: $form,
-      $results: $results,
-      $atomAutodiscoveryLink: $atomAutodiscoveryLink
-    })
-  }
-})
+if ($form && $results && $atomAutodiscoveryLink) {
+  // eslint-disable-next-line no-new
+  new GOVUK.LiveSearch({
+    $form: $form,
+    $results: $results,
+    $atomAutodiscoveryLink: $atomAutodiscoveryLink
+  })
+}
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -356,6 +356,7 @@
     var searchState = this.serializeState(this.state)
     var cachedResultData = this.cache(searchState)
     var liveSearch = this
+
     if (typeof cachedResultData === 'undefined') {
       this.showLoadingIndicator()
       var xhr = new XMLHttpRequest()
@@ -364,12 +365,8 @@
       var done = function (e) {
         if (xhr.readyState === 4 && xhr.status === 200) {
           var response = JSON.parse(e.target.response)
-          var newPath = encodeURI(window.location.pathname + '?' + liveSearch.serializeState(liveSearch.state))
-          window.history.pushState(liveSearch.state, '', newPath)
-          liveSearch.trackingInit()
-          liveSearch.setRelevantResultCustomDimension()
-          liveSearch.trackPageView()
-
+          liveSearch.updateUrl()
+          liveSearch.trackSearch()
           liveSearch.cache(liveSearch.serializeState(liveSearch.state), response)
           liveSearch.displayResults(response, searchState)
         } else {
@@ -382,8 +379,21 @@
       xhr.addEventListener('load', done)
       xhr.send()
     } else {
+      this.updateUrl()
+      this.trackSearch()
       this.displayResults(cachedResultData, searchState)
     }
+  }
+
+  LiveSearch.prototype.updateUrl = function () {
+    var newPath = encodeURI(window.location.pathname + '?' + this.serializeState(this.state))
+    window.history.pushState(this.state, '', newPath)
+  }
+
+  LiveSearch.prototype.trackSearch = function () {
+    this.trackingInit()
+    this.setRelevantResultCustomDimension()
+    this.trackPageView()
   }
 
   LiveSearch.prototype.updateLinks = function updateLinks () {
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -1,4 +1,3 @@
-/* eslint-env jquery */
 (function () {
   'use strict'
 
@@ -11,30 +10,41 @@
     this.resultCache = {}
 
     this.$form = options.$form
-    this.$searchSubmitButton = this.$form.find('.gem-c-search__submit')
-    this.$resultsWrapper = this.$form.find('.js-live-search-results-block')
-    this.$suggestionsBlock = this.$form.find('#js-spelling-suggestions')
-    this.$resultsBlock = options.$results.find('#js-results')
-    this.$countBlock = options.$results.find('#js-result-count')
-    this.$mobileResultsCount = this.$form.find('.js-result-count')
-    this.$selectedFilterCount = this.$form.find('.js-selected-filter-count')
-    this.$facetTagBlock = options.$results.find('#js-facet-tag-wrapper')
-    this.$mobileFacetTagBlock = this.$form.find('.js-mobile-facet-tag-block')
-    this.$loadingBlock = options.$results.find('#js-loading-message')
-    this.$sortBlock = options.$results.find('#js-sort-options')
-    this.$paginationBlock = options.$results.find('#js-pagination')
-    this.action = this.$form.attr('action') + '.json'
+    this.$searchSubmitButton = this.$form.querySelector('.gem-c-search__submit')
+    this.$resultsWrapper = this.$form.querySelector('.js-live-search-results-block')
+    this.$suggestionsBlock = this.$form.querySelector('#js-spelling-suggestions')
+    this.$resultsBlock = options.$results.querySelector('#js-results')
+    this.$countBlock = options.$results.querySelector('#js-result-count')
+    this.$mobileResultsCount = this.$form.querySelector('.js-result-count')
+    this.$selectedFilterCount = this.$form.querySelector('.js-selected-filter-count')
+    this.$facetTagBlock = options.$results.querySelector('#js-facet-tag-wrapper')
+    this.$mobileFacetTagBlock = this.$form.querySelector('.js-mobile-facet-tag-block')
+    this.$loadingBlock = options.$results.querySelector('#js-loading-message')
+    this.$sortBlock = options.$results.querySelector('#js-sort-options')
+    this.$paginationBlock = options.$results.querySelector('#js-pagination')
+    this.action = this.$form.getAttribute('action') + '.json'
     this.$atomAutodiscoveryLink = options.$atomAutodiscoveryLink
-    this.baseTitle = $("meta[name='govuk:base_title']").attr('content') || document.title
-    this.$resultsCountMetaTag = $("meta[name='govuk:search-result-count']")
-    this.$emailLink = $('a[href*="email-signup"]')
+
+    this.baseTitle = document.querySelector("meta[name='govuk:base_title']")
+    if (this.baseTitle) {
+      this.baseTitle = this.baseTitle.getAttribute('content')
+    } else {
+      this.baseTitle = document.title
+    }
+
+    this.$resultsCountMetaTag = document.querySelector("meta[name='govuk:search-result-count']")
+    this.$emailLinks = document.querySelectorAll('[href*="email-signup"]')
     this.previousSearchTerm = ''
 
-    this.emailSignupHref = this.$emailLink.attr('href')
-    this.$atomLink = $('a[href*=".atom"]')
-    this.atomHref = this.$atomLink.attr('href')
+    if (this.$emailLinks[0]) {
+      this.emailSignupHref = this.$emailLinks[0].getAttribute('href')
+    }
+    this.$atomLinks = document.querySelectorAll('[href*=".atom"]')
+    if (this.$atomLinks[0]) {
+      this.atomHref = this.$atomLinks[0].getAttribute('href')
+    }
     this.bindSortElements()
-    this.getTaxonomyFacet().update()
+    this.getAndUpdateTaxonomyFacet()
 
     if (window.ga) {
       // Use navigator.sendBeacon
@@ -46,72 +56,87 @@
 
     // prevent page refresh on search submit button click
     // instead trigger the AJAX results fetch
-    this.$searchSubmitButton.on('click',
-      function (e) {
+    if (this.$searchSubmitButton) {
+      this.$searchSubmitButton.addEventListener('click', function (e) {
         e.preventDefault()
         this.formChange(e)
-      }.bind(this)
-    )
+      }.bind(this))
+    }
 
     if (GOVUK.support.history()) {
       this.saveState()
 
-      this.$form.on('change', 'input[type=checkbox], input[type=radio], select',
-        function (e) {
-          this.formChange(e)
-        }.bind(this)
-      )
+      this.$form.addEventListener('change', this.formChange.bind(this))
+
       // custom event listener on the form, that fires the update only once
       // when we clear of filters
-      // fired from javascripts/modules/mobile-filters-modal.js:139
-      this.$form.on('customFormChange', this.$form,
-        function (e) {
-          this.formChange(e)
-        }.bind(this)
-      )
+      // fired from javascripts/modules/mobile-filters-modal.js
+      this.$form.addEventListener('customFormChange', this.formChange.bind(this))
 
-      this.$form.on('change keypress', 'input[type=text],input[type=search]',
-        function (e) {
-          var ENTER_KEY = 13
+      this.handleKeyPress = function (e) {
+        var ENTER_KEY = 13
 
-          if (e.keyCode === ENTER_KEY || e.type === 'change') {
-            // cater for jQuery and native events
-            var suppressAnalytics = e.suppressAnalytics || (e.detail && e.detail.suppressAnalytics)
-
-            if (e.currentTarget.value !== this.previousSearchTerm && !suppressAnalytics) {
-              LiveSearch.prototype.fireTextAnalyticsEvent(e)
-            }
-            this.formChange(e)
-            this.previousSearchTerm = e.currentTarget.value
-            e.preventDefault()
+        if (e.keyCode === ENTER_KEY || e.type === 'change') {
+          // cater for jQuery and native events
+          var suppressAnalytics = e.suppressAnalytics || (e.detail && e.detail.suppressAnalytics)
+          if (e.currentTarget.value !== this.previousSearchTerm && !suppressAnalytics) {
+            LiveSearch.prototype.fireTextAnalyticsEvent(e)
           }
-        }.bind(this)
-      )
+          this.formChange(e)
+          this.previousSearchTerm = e.currentTarget.value
+          e.preventDefault()
+        }
+      }
+
+      var inputs = this.$form.querySelectorAll('input[type=text],input[type=search]')
+      for (var i = 0; i < inputs.length; i++) {
+        inputs[i].addEventListener('change', this.handleKeyPress.bind(this))
+        inputs[i].addEventListener('keypress', this.handleKeyPress.bind(this))
+      }
 
       this.indexTrackingData()
 
-      $(window).on('popstate', this.popState.bind(this))
+      document.addEventListener('popstate', this.popState.bind(this))
     } else {
-      this.$form.find('.js-live-search-fallback').show()
+      var fallback = this.$form.querySelector('.js-live-search-fallback')
+      fallback.style.display = 'block'
     }
   }
 
   LiveSearch.prototype.startEnhancedEcommerceTracking = function startEnhancedEcommerceTracking () {
-    this.$resultsWrapper.attr('data-search-query', this.currentKeywords())
-    this.$suggestionsBlock.attr('data-search-query', this.currentKeywords())
+    if (this.$resultsWrapper) {
+      this.$resultsWrapper.setAttribute('data-search-query', this.currentKeywords())
+    }
+    if (this.$suggestionsBlock) {
+      this.$suggestionsBlock.setAttribute('data-search-query', this.currentKeywords())
+    }
     if (GOVUK.Ecommerce) { GOVUK.Ecommerce.start() }
   }
 
-  LiveSearch.prototype.getTaxonomyFacet = function getTaxonomyFacet () {
-    this.taxonomy = this.taxonomy || new GOVUK.TaxonomySelect({ $el: $('.js-taxonomy-select') })
-    return this.taxonomy
+  LiveSearch.prototype.getAndUpdateTaxonomyFacet = function getAndUpdateTaxonomyFacet () {
+    var taxonomySelect = document.querySelector('.js-taxonomy-select')
+    if (taxonomySelect) {
+      this.taxonomy = this.taxonomy || new GOVUK.TaxonomySelect({ $el: taxonomySelect })
+      // return this.taxonomy
+      this.taxonomy.update()
+    }
   }
 
   LiveSearch.prototype.getSerializeForm = function getSerializeForm () {
-    var serialized = this.$form.serializeArray()
-    var filtered = serialized.filter(function (field) {
-      return field.value !== '' && field.name !== 'option-select-filter'
+    var serialized = new FormData(this.$form)
+
+    var filtered = []
+    serialized.forEach(function (value, key) {
+      if (key !== '' && key !== 'option-select-filter' && value !== '') {
+        filtered.push(
+          {
+            name: key,
+            value: value
+          }
+        )
+      }
     })
+
     return filtered
   }
 
@@ -136,7 +161,7 @@
   LiveSearch.prototype.formChange = function formChange (e) {
     var pageUpdated
     if (this.isNewState()) {
-      this.getTaxonomyFacet().update()
+      this.getAndUpdateTaxonomyFacet()
       this.saveState()
       this.updateOrder()
       this.updateLinks()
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -134,50 +160,63 @@
...
-    $spellingSuggestionMetaTag.attr('content', suggestion)
+    var $spellingSuggestionMetaTag = document.querySelector("meta[name='govuk:spelling-suggestion']")
+    if (this.$suggestionsBlock && $spellingSuggestionMetaTag) {
+      var spellingSuggestionAvailable = this.$suggestionsBlock.querySelectorAll('a').length > 0
+      // currently there's ever only one suggestion
+      var link = this.$suggestionsBlock.querySelector('a')
+      var suggestion = ''
+      if (link) {
+        link = JSON.parse(link.getAttribute('data-track-options'))
+        suggestion = spellingSuggestionAvailable ? link.dimension81 : ''
```
kevindew, line 216: I don't think we need this condition since to have a link a spelling suggestion should be available (unless I'm misunderstanding something).

app/assets/javascripts/live_search.js
```diff
@@ -187,40 +226,41 @@
    * rewrite the appropriate tracking data attribute to delineate the group and document index.
    */
   LiveSearch.prototype.indexTrackingData = function indexTrackingData () {
-    var $groupEls = $('.filtered-results__group')
+    var $groupEls = document.querySelectorAll('.filtered-results__group')
     if ($groupEls.length > 0) {
```
kevindew, line 230: This conditional seems redundant, as the following lines will just not execute anything if the query returns an empty NodeList.

app/assets/javascripts/live_search.js
```diff
@@ -187,40 +226,41 @@
...
-      })
+          trackingAction = [trackingAction, g + 1, d + 1].join('.')
+          $documentLink.setAttribute('data-track-action', trackingAction)
+        }
+      }
     }
 
-    var $results = $('.js-finder-results')
+    var $results = document.querySelector('.js-finder-results') || []
     if ($results.length > 0) {
```
kevindew, line 246: I think the array fallback is unnecessary here, unless I'm missing something.

```suggestion
    var $results = document.querySelector('.js-finder-results')
    if ($results) {
```

app/assets/javascripts/live_search.js
```diff
@@ -254,27 +294,30 @@
...
+    }
   }
 
   LiveSearch.prototype.currentKeywords = function currentKeywords () {
     return this.getTextInputValue('keywords', this.state)
   }
 
   LiveSearch.prototype.updateOrder = function updateOrder () {
-    if (!this.$orderSelect.length) {
+    if (!this.$orderSelect || !this.$orderSelect.length) {
```
kevindew, line 320: I don't think this needs the 2nd half of the conditional. I can't see a scenario where this is a NodeList - perhaps I'm missing something?

kevindew (commented):
Amazing job @andysellick - this looks like it's been quite the mission. Super effort 💯 

I've run out of time again on putting a review together so felt it was best if I submitted a partial one

app/assets/javascripts/application.js
```diff
@@ -28,25 +28,21 @@
...
-  if ($form.length && $results.length) {
-    // eslint-disable-next-line
-    new GOVUK.LiveSearch({
-      $form: $form,
-      $results: $results,
-      $atomAutodiscoveryLink: $atomAutodiscoveryLink
-    })
-  }
-})
+if ($form && $results) {
```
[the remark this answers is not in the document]
andysellick: Good point, have updated.

app/assets/javascripts/live_search.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-env jquery */
+/* eslint-env */
```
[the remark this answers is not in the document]
andysellick: This was already removed but I assume not in the commit you were viewing. Let me know if you think the change needs to go in a different commit.

app/assets/javascripts/live_search.js
```diff
@@ -11,30 +11,35 @@
...
+    this.$countBlock = options.$results.querySelector('#js-result-count')
+    this.$mobileResultsCount = this.$form.querySelector('.js-result-count')
+    this.$selectedFilterCount = this.$form.querySelector('.js-selected-filter-count')
+    this.$facetTagBlock = options.$results.querySelector('#js-facet-tag-wrapper')
+    this.$mobileFacetTagBlock = this.$form.querySelector('.js-mobile-facet-tag-block')
+    this.$loadingBlock = options.$results.querySelector('#js-loading-message')
+    this.$sortBlock = options.$results.querySelector('#js-sort-options')
+    this.$paginationBlock = options.$results.querySelector('#js-pagination')
+    this.action = this.$form.getAttribute('action') + '.json'
     this.$atomAutodiscoveryLink = options.$atomAutodiscoveryLink
```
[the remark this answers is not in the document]
andysellick: That is an excellent question. There was quite a lot about this code during this work that raised similar questions. I worry that this is one of our larger chunks of JS and its mostly been left to itself to stagnate for several years now.

app/assets/javascripts/live_search.js
```diff
@@ -11,30 +11,35 @@
...
-    this.baseTitle = $("meta[name='govuk:base_title']").attr('content') || document.title
-    this.$resultsCountMetaTag = $("meta[name='govuk:search-result-count']")
-    this.$emailLink = $('a[href*="email-signup"]')
+    this.baseTitle = document.querySelector("meta[name='govuk:base_title']")
+    this.baseTitle = this.baseTitle.getAttribute('content') || document.title
```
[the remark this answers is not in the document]
andysellick: The original line that this replaced was

```
this.baseTitle = $("meta[name='govuk:base_title']").attr('content') || document.title
```

In jQuery if the tag wasn't found then the first part of this statement would return empty, then fall back to the document title. So yes, a better solution (without jQuery's forgiving nature) would be:

```
this.baseTitle = document.querySelector("meta[name='govuk:base_title']")
if (this.baseTitle) {
  this.baseTitle = this.baseTitle.getAttribute('content')
} else {
  this.baseTitle = document.title
}
```

Good spot. So many extra lines!

app/assets/javascripts/live_search.js
```diff
@@ -45,73 +50,94 @@
...
 
     if (GOVUK.support.history()) {
       this.saveState()
 
-      this.$form.on('change', 'input[type=checkbox], input[type=radio], select',
-        function (e) {
-          this.formChange(e)
-        }.bind(this)
-      )
+      var checkboxRadioSelects = this.$form.querySelectorAll('input[type=checkbox], input[type=radio], select')
```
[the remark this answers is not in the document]
andysellick: No, they all come from a Rails template so that the non-JS version still has all the same options. JS is used to show/hide/disable/select/unselect some of them on page load. Is that okay? I didn't fully understand what event you were referring to.

app/assets/javascripts/live_search.js
```diff
@@ -45,73 +50,94 @@
...
+      for (var c = 0; c < checkboxRadioSelects.length; c++) {
+        checkboxRadioSelects[c].addEventListener('change', this.formChange.bind(this))
+      }
+
       // custom event listener on the form, that fires the update only once
       // when we clear of filters
       // fired from javascripts/modules/mobile-filters-modal.js:139
```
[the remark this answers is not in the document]
andysellick: Good shout, dropped. Related/unrelated: that script could now use the GOVUK.triggerEvent script...

app/assets/javascripts/live_search.js
```diff
@@ -45,73 +50,94 @@
...
+    if (this.$searchSubmitButton) {
+      this.$searchSubmitButton.addEventListener('click', function (e) {
         e.preventDefault()
         this.formChange(e)
-      }.bind(this)
-    )
+      }.bind(this))
+    }
 
     if (GOVUK.support.history()) {
```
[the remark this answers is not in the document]
andysellick: I'm not certain but I think if this bit doesn't run then the search falls back to the non-JS version, which is what we want for IE less than 10.

app/assets/javascripts/live_search.js
```diff
@@ -45,73 +50,94 @@
...
-    var serialized = this.$form.serializeArray()
-    var filtered = serialized.filter(function (field) {
-      return field.value !== '' && field.name !== 'option-select-filter'
-    })
+    var formElements = this.$form.elements
+    var filtered = []
+
+    for (var i = 0; i < formElements.length; i++) {
+      var el = formElements[i]
+      if ((el.type && el.type !== 'checkbox' && el.type !== 'radio') || el.checked) {
```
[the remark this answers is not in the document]
andysellick: The only disabled element I can think of is the sub topic select element, but when it's disabled it also has no value, which is caught by the next check inside this `if`, so I think we're okay.

app/assets/javascripts/live_search.js
```diff
@@ -45,73 +50,94 @@
...
+        var name = el.getAttribute('name')
+        var value = el.value
+        if (name !== '' && name !== 'option-select-filter' && name !== null && value !== '' && value !== undefined) {
```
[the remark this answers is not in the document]
andysellick: Thanks that's much simpler. Sorry that was one of those 'oh and I need to check this as well' overly long statements that I should have pruned earlier. Improved.

app/assets/javascripts/live_search.js
```diff
@@ -134,50 +160,63 @@
...
-      )
+      this.updateResults()
+    }
+  }
+
+  LiveSearch.prototype.serializeState = function (state) {
+    var params = []
+    if (Array.isArray(state)) {
+      for (var i = 0; i < state.length; i++) {
+        params.push(state[i].name + '=' + state[i].value)
```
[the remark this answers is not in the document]
andysellick: I think it's safe to assume that the `name` will already include `[]` if that's appropriate (e.g. if it's a checkbox or radio). Otherwise we might end up with things like `vessel_type[][]=merchant-vessel`.

On the encoding question, the state gets encoded in a single line in `updateResults` (currently on line 370) which feels more efficient to do it all in one gone that potentially call encodeURIComponent hundreds of times.

app/assets/javascripts/live_search.js
```diff
@@ -134,50 +160,63 @@
...
+      }
+    } else {
+      for (var key in state) {
+        if (Object.prototype.hasOwnProperty.call(state, key)) {
+          params.push(encodeURIComponent(key) + '=' + encodeURIComponent(state[key]))
+        }
+      }
     }
+    return params.join('&')
   }
```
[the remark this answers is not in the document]
andysellick: I'm generally against polyfills because they can add more code than needed (or known) and once added, they need to be maintained and probably removed further down the road. Writing it this way is a bit delicate but I think clearer (and I learned plenty about form serialisation in the process!).

If we did more form handling like this then I'd definitely want a way of modernising and reusing the code better, but as the search code is a weird standalone anomaly I think it's okay.

app/assets/javascripts/live_search.js
```diff
@@ -134,50 +160,63 @@
...
-    var $mostRelevantDocumentLink = $('.js-finder-results').find('.gem-c-document-list__item--highlight')
-    var dimensionValue = $mostRelevantDocumentLink.length ? 'yes' : 'no'
+    var $mostRelevantDocumentLink = this.$form.querySelector('.gem-c-document-list__item--highlight')
+    var dimensionValue = $mostRelevantDocumentLink ? 'yes' : 'no'
     GOVUK.SearchAnalytics.setDimension(83, dimensionValue)
   }
 
   LiveSearch.prototype.trackingInit = function trackingInit () {
-    GOVUK.modules.start($(this.$resultsWrapper))
+    GOVUK.modules.start($(this.$resultsWrapper)) // this still needs to pass a jQuery object until we update the modules code
```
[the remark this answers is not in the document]
andysellick: When I encountered this I added an item to that card to remember to convert this.

app/assets/javascripts/live_search.js
```diff
@@ -134,50 +160,63 @@
...
 
   LiveSearch.prototype.trackSpellingSuggestionsImpressions = function trackSpellingSuggestionsImpressions ($suggestions) {
-    var $spellingSuggestionMetaTag = $("meta[name='govuk:spelling-suggestion']")
-    // currently there's ever only one suggestion
-    var spellingSuggestionAvailable = this.$suggestionsBlock.find('a').length > 0
-    var suggestion = spellingSuggestionAvailable ? this.$suggestionsBlock.find('a').data('track-options').dimension81 : ''
-    $spellingSuggestionMetaTag.attr('content', suggestion)
+    var $spellingSuggestionMetaTag = document.querySelector("meta[name='govuk:spelling-suggestion']")
+    if (this.$suggestionsBlock && $spellingSuggestionMetaTag) {
+      var spellingSuggestionAvailable = this.$suggestionsBlock.querySelectorAll('a').length > 0
```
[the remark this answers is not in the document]
andysellick: Argh, that was garbled. Good shout, thanks.

app/assets/javascripts/live_search.js
andysellick: Good point, have rewritten this section.

app/assets/javascripts/live_search.js
andysellick: Good point, will change.

app/assets/javascripts/live_search.js
andysellick: Yes, with the same logic applied to the `if ($mostRelevantDocumentLink) {` line that follows.

app/assets/javascripts/live_search.js
andysellick: Good point, changed.

app/assets/javascripts/application.js
```diff
@@ -28,25 +28,21 @@
...
-    // eslint-disable-next-line
-    new GOVUK.LiveSearch({
-      $form: $form,
-      $results: $results,
-      $atomAutodiscoveryLink: $atomAutodiscoveryLink
-    })
-  }
-})
+if ($form && $results && $atomAutodiscoveryLink) {
+  // eslint-disable-next-line
```
injms, line 42: What rules is this exemption needed for?

app/assets/javascripts/live_search.js
```diff
@@ -45,73 +56,94 @@
...
 
     // prevent page refresh on search submit button click
-    // instead trigger the AJAX results fetch
-    this.$searchSubmitButton.on('click',
-      function (e) {
+    // instead trigger the results fetch
+    if (this.$searchSubmitButton) {
+      this.$searchSubmitButton.addEventListener('click', function (e) {
         e.preventDefault()
         this.formChange(e)
```
injms, line 63: Looking at the `formChange` function, I don't think it actually does anything with the event variable `e` passed into it. Could it be removed?

https://github.com/alphagov/finder-frontend/blob/c5940fdeacadb5f21b843c41904020b04e8dbfb8/app/assets/javascripts/live_search.js#L168-L177

injms (approved):
A couple of non-blocking questions - but nothing that should stop this massive improvement from being merged in.

app/assets/javascripts/application.js
andysellick: I didn't know, so I took it out and ran the linter, and it's for `Do not use 'new' for side effects`: https://eslint.org/docs/rules/no-new.html

Not sure if there's a reasonable way around that. Another thing to add to the list for refactoring this code at some point...

app/assets/javascripts/live_search.js
andysellick: Good point, seems not, have removed.

app/assets/javascripts/application.js
injms: Ah, I see - would it be worth disable the `Do not use 'new' for side effects` rule explictly? It's clearer what's being done and doesn't disable all of ESLint's rules:

```suggestion
  // eslint-disable-next-line no-new
```

features/step_definitions/analytics_steps.rb
```diff
@@ -39,7 +39,7 @@
 And "I search for lunch" do
   stub_search_api_request_with_query_param_no_results("lunch")
 
-  fill_in "Search", with: "lunch"
+  fill_in "Search", with: "lunch", fill_options: { clear: :backspace }
```
kevindew, line 42: I did a bit of digging into why this was needed but couldn't work out why aside from this bug: https://github.com/redux-form/redux-form/issues/686 - which I assume is what you found too.

I found that I also got this problem with the main branch locally as well so I don't think it was something in this code that introduces it.

It'd be good if we put this in a separate commit with some documentation as to why we added it since it's unconventional.

All I can really determine is that when we tell Capybara to fill_in a field it [sends a JS message before it to select() the contents so that the new information replaces the old](https://github.com/teamcapybara/capybara/blob/f7ab0b5cd5da86185816c2d5c30d58145fe654ed/lib/capybara/selenium/node.rb#L281). I'm not sure if we have any events that change focus which perhaps interfere with this somehow and make it a race condition.

kevindew (commented):
Thanks Andy, this is looking great. Just a few minor comments to deal with then good to merge.

app/assets/javascripts/live_search.js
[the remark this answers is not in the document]
andysellick: Ah of course, thanks, yes, have now removed.

app/assets/javascripts/live_search.js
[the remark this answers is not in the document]
andysellick: I understand now. That's a very good point which I hadn't considered, but fortunately for me the form elements don't get updated when the search changes - only the results portion of the page is changed.

However, it turns out I was being needlessly specific - I can just set the 'change' eventListener on the form itself, and the event bubbles up. Code's a lot simpler, only one event listener set instead of hundreds. Thanks!

app/assets/javascripts/live_search.js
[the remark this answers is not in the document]
andysellick: I think I'll put this on the 'things to consider when this code gets a proper refactor' list - I'd rather not break it.

app/assets/javascripts/live_search.js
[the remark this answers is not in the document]
andysellick: I put some logging into the code and this function is called with arrays and objects in what looks to be broadly similar numbers when the jasmine tests are run. A quick test of the code running in a browser showed only calls with arrays passed, but I haven't comprehensively checked all the possible searches.

app/assets/javascripts/application.js
andysellick: Good shout, done.

app/assets/javascripts/taxonomy-select.js
```diff
@@ -20,52 +17,59 @@
...
 
   TaxonomySelect.prototype.disableSubTaxonFacet = function disableSubTaxonFacet () {
-    var topLevelTaxonSelected = !!this.$topLevelTaxon().val()
-    this.$subTaxon().attr('disabled', !topLevelTaxonSelected)
+    var topLevelTaxonSelected = !!this.$topLevelTaxon().value
+    if (!topLevelTaxonSelected) {
+      this.$subTaxon().setAttribute('disabled', true)
+    } else {
+      this.$subTaxon().removeAttribute('disabled')
+    }
```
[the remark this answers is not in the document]
andysellick: I tried and it didn't work. I think the problem is that the disabled attribute needs to be completely removed in the `else` - if it's present on an element, most browsers don't care if it's set to `true` or `false`, either means the element is disabled.

app/assets/javascripts/taxonomy-select.js
```diff
@@ -20,52 +17,59 @@
...
+    var taxons = this.options[this.$topLevelTaxon().value]
     var subtaxon = this.$subTaxon()
+    var options = subtaxon.querySelectorAll('option')
 
-    subtaxon.find('option').each(function () {
-      if ($(this).val()) { $(this).remove() }
-    })
-
-    subtaxon.append(taxons)
+    for (var o = options.length - 1; o >= 0; o--) {
```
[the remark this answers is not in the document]
andysellick: I don't remember why I wrote it like this, but it seems to work the other way round too, maybe I was being paranoid. Have changed.

spec/javascripts/support/jasmine.yml
```diff
@@ -12,6 +12,7 @@
 #
 src_files:
   - spec/javascripts/vendor/jquery-1.12.4.js
+  - spec/javascripts/vendor/mock-ajax.js
```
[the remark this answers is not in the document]
andysellick: Good point, changed.

features/step_definitions/analytics_steps.rb
andysellick: Good point, have separated this out into a separate commit.

spec/javascripts/live_search_spec.js
```diff
@@ -197,7 +180,7 @@ describe('liveSearch', function () {
 
     it('should have an order state selected when keywords are present', function () {
       liveSearch.state = 'find-eu-exit-guidance-business.json?keywords=123'
-      expect(liveSearch.$orderSelect.val()).not.toBe(null)
+      expect(liveSearch.$orderSelect.value).not.toBe(null)
```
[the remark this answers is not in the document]
andysellick: Weirdly even if a `select` doesn't have an empty `option`, I think it might be still possible to set the value to `null` using JS. I'm going to kick the can down the road again - I'm not comfortable removing this test based on my limited understanding of this code, and would rather look at it again in a proper refactor.

app/assets/javascripts/taxonomy-select.js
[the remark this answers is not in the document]
kevindew: The situation you're describing is the case if you're operating on the HTML attribute, i.e. needing to use setAttribute and similar methods. However form inputs have a [JS property of disabled](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/disabled) which can only be true/false and can never be removed (as it's a property of the element) 

Not a blocker for this PR though.

spec/javascripts/live_search_spec.js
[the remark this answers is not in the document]
kevindew: Sure, no worry to defer it. I can open a PR afterwards.

The code doesn't actually do anything. If we change `liveSearch.state` then nothing happens except `liveSearch.state` has a different value. So any assertions on other properties aren't going to change, let alone element values.

Btw, even with JS a select.value can't be null:

```
document.querySelector('select').value
> "af07d5a5-df63-4ddc-9383-6a666845ebe9"
document.querySelector('select').value = null
> null
document.querySelector('select').value
> ""
```

kevindew (approved):
Nice one Andy, looks good to merge. Sorry it's taken so long to get back to this.

andysellick:
Found a small bug while doing some last minute checks, which I need to resolve before merging.

spec/javascripts/live_search_spec.js
```diff
@@ -170,6 +174,44 @@ describe('liveSearch', function () {
...
   })
 
+  describe('should update the URL correctly', function () {
+    beforeEach(function () {
+      // clear these options to simplify the URLs
+      $form.find('input[name=field]').prop('checked', false)
+      $form.find('input[name=published_at]').val('')
+    })
+
+    it('when search options change', function () {
```
kevindew, line 184: This reads a bit strange as "it when search options change" I can see it's a continuation of "should update the URL correctly" but would suggest doing something like:

describe('updating the URL correctly')
it('should update when the search options change')
it('should update when there is a cached response')

spec/javascripts/live_search_spec.js
```diff
@@ -170,6 +174,44 @@ describe('liveSearch', function () {
...
+          status: 200,
+          response: '{}'
+        })
+      }
+      // set options so that a cached state is triggered
+      $form.find('input[value="john"]').click()
+      $form.find('input[value="paul"]').click()
+      $form.find('input[value="paul"]').click()
+      expect(window.location.search).toContain('people%5B%5D=john')
+      expect(window.location.search).not.toContain('people%5B%5D=paul')
```
kevindew, line 211: I'm a bit confused by this test and how it's testing a not cached state. My confusion is based around an expectation I have that this has 3 states (load first results - john, load second results - john & paul, return to cached first result - john) but has 3 stubs, so I'm not sure I'm understanding this wrong.

app/assets/javascripts/live_search.js
```diff
@@ -382,10 +378,19 @@
       xhr.addEventListener('load', done)
       xhr.send()
     } else {
+      this.xmlHttpDone()
       this.displayResults(cachedResultData, searchState)
     }
   }
 
+  LiveSearch.prototype.xmlHttpDone = function () {
```
kevindew, line 386: I think this name is a little misleading as it can be called when no XHR has been made.

I can't think of a particularly better name for the method, since it seems to do a few things, but I did wonder if it's actually all part of displaying results (since they seem to be coupled) and should be a part of that method.

kevindew (commented):
Great job resolving the bug Andy. I've made a few suggestions of the organisation of the implementation.

spec/javascripts/live_search_spec.js
andysellick: Fair enough. Have tweaked, now looks like this:

Screenshot 2021-08-09 at 08 44 00

spec/javascripts/live_search_spec.js
andysellick: My mistake, I left in a stub that wasn't needed. You're right, this only has two states. Have added some comments to clarify.

app/assets/javascripts/live_search.js
andysellick: Thanks for this comment - it's always worth remembering that code should be readable for the next developer. With that in mind, I've split this into two functions, one for the URL change and one for calling the tracking. It feels a little verbose to do this, but I think it makes the code clearer in terms of what it's actually doing.

spec/javascripts/live_search_spec.js
```diff
@@ -187,6 +174,42 @@ describe('liveSearch', function () {
...
   })
 
+  describe('the URL should change', function () {
+    beforeEach(function () {
+      // clear these options to simplify the URLs
+      $form.find('input[name=field]').prop('checked', false)
+      $form.find('input[name=published_at]').val('')
+    })
+
+    it('when the search options change', function () {
```
kevindew, line 184: Thanks for amending. This still reads wrong to me with "it when the search options change", which makes it not feel like an idiomatic way to express a scenario. Can it be changed so that the it forms the expectation? i.e. `it('should ...')`

spec/javascripts/live_search_spec.js
kevindew: Thank you!

kevindew (commented):
Just a couple of minor follow-ups from the last review.

spec/javascripts/live_search_spec.js
andysellick: Okay, I've had another crack at it. Hopefully... better?

kevindew (approved):
Super 👍  Thanks for working through all of this.

PR #2519, opened 2021-05-25
Remove JQuery from EnableAriaControls module

leenagupte:
Trello: https://trello.com/c/jK6hrK3d

# What's changed?

Replace [jQuery](https://jquery.com/) code in GOV.UK JavaScript (JS) code with vanilla (plain/normal) JavaScript in the EnableAriaControls module.

# Why?

Co-authored-by: Andy Sellick <andy.sellick@digital.cabinet-office.gov.uk>


owenatgov (approved):
Good looks peachy.

One non-blocking query I have is what this code is for. I wasn't patient enough to go looking for a live example but from what it looks like, we're sending markup to the frontend with a data-attribute containing aria controls. Why can't we just set aria controls at the template level and bypass the need for this module altogether? Is this a govspeak thing or something?

PR #2529, opened 2021-06-02
Use GovukPersonalisation::AccountConcern

AlanGabbianelli:
The new govuk_personalisation gem extracted some of the functionality
that was present in AccountConcern so we can use that.

Trello card: https://trello.com/c/40YbnkRp/783-add-a-rails-concern-to-govukpersonalisation


app/controllers/concerns/account_concern.rb
```diff
@@ -3,17 +3,13 @@
...
-  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
-  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
-  ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-End-Session"
-  ACCOUNT_SESSION_DEV_COOKIE_NAME = "govuk_account_session"
-
   ATTRIBUTE_NAME = "transition_checker_state"
 
   included do
-    before_action :fetch_account_session_header
+    include GovukPersonalisation::AccountConcern
```
huwd, line 9: Ah interesting, was this the solution to some of those instance methods not being visible?

So if we included this into the `AccountConcern` the methods from `GovukPersonalisation::AccountConcern` would be mixed into `AccountConcern` right?

But in the included block they get spread into anything the `AccountConcern` is included into? So the like how those `before_action` start to function as a part of the controller we do `include AccountConcern` in?

If this line goes up to line 4, do things stop wokring then?

AlanGabbianelli: Moving this to line 4 results in:
```
An error occurred while loading ./spec/controllers/brexit_checker_controller_spec.rb.
Failure/Error: include GovukPersonalisation::AccountConcern

NoMethodError:
  undefined method `before_action' for AccountConcern:Module
# ./app/controllers/concerns/account_concern.rb:4:in `include'
# ./app/controllers/concerns/account_concern.rb:4:in `<module:AccountConcern>'
# ./app/controllers/concerns/account_concern.rb:3:in `<top (required)>'
# ./app/controllers/brexit_checker_controller.rb:2:in `<class:BrexitCheckerController>'
# ./app/controllers/brexit_checker_controller.rb:1:in `<top (required)>'
# ./spec/controllers/brexit_checker_controller_spec.rb:3:in `<top (required)>'


Finished in 0.00008 seconds (files took 10.93 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

Both `AccountConcern` and `GovukPersonalisation::AccountConcern` provide some functionalities (methods and `before_action`s) that we want available in `BrexitCheckerController`.

Adding `include GovukPersonalisation::AccountConcern` in the `included do` block of `AccountConcern` makes sure that wherever we include `AccountConcern`, we also include `GovukPersonalisation::AccountConcern`.

huwd: ah gotcha!
Of course, so it's not just our instance variables it's that we've got before actions in our gem that also need to be in thsi block.

makes sense, cheers for checking!

huwd (approved):

app/controllers/concerns/account_concern.rb
AlanGabbianelli: We could have added it directly to the controller
```ruby
class BrexitCheckerController < ApplicationController
  include GovukPersonalisation::AccountConcern
  include AccountConcern
  # ...
end
```
but since we know that when we include `AccountConcern` we always also want to include `GovukPersonalisation::AccountConcern`, might as well do it this way.

Also, the order of how we include them matters - we couldn't include `AccountConcern` first.

huwd: Yeah your approach makes much more sense, it's just interesting first time you see it.
I've only ever seen moduels being included directly into the class.

Including something in an `includes do` block struck me as a bit inception initially.
But the before_actions make perfect sense!

Thanks for taking the time to explain

PR #2532, opened 2021-06-04
Remove JQuery from the expander

JonathanHallam:
## What

Remove jQuery from the expander file.

## How

I've done this by sort of reverting [this PR](https://github.com/alphagov/finder-frontend/pull/1748). I've moved the listener into both the `select` and `input[type="text"]` parts of the `if / else` loop in order to remove the need for jQuery. I've also removed the jQuery that was triggering the events that it listens for in order to ensure that the bug outlined in that PR remains fixed.

## Why

### DIFF
--- a/app/assets/javascripts/components/expander.js
+++ b/app/assets/javascripts/components/expander.js
@@ -31,9 +31,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
       this.$toggleButton.click()
     }
 
-    // listen for the change event
-    $module.on('change', 'select, input[type="text"]', this.updateSelectedCount.bind(this))
-
     // Attach listener function to update selected count
     var boundChangeEvents = this.bindChangeEvents.bind(this)
     boundChangeEvents()
@@ -42,6 +39,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
   Expander.prototype.bindChangeEvents = function (e) {
     for (var i = 0; i < this.$allInteractiveElements.length; i++) {
       var $el = this.$allInteractiveElements[i]
+      if ($el.tagName === 'SELECT') {
+        $el.addEventListener('change', this.updateSelectedCount.bind(this))
+      }
+
       // but for inputs we need both change and enter key event
       if ($el.tagName === 'INPUT') {
         $el.addEventListener('change', this.handleInputEvent.bind(this))
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -5,11 +5,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict'
 
   GOVUK.Modules.RemoveFilter = function RemoveFilter () {
-    var onChangeSuppressAnalytics = {
-      type: 'change',
-      suppressAnalytics: true
-    }
-
     this.start = function (element) {
       $(element).on('click', '[data-module="remove-filter-link"]', toggleFilter)
     }
@@ -36,7 +31,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       if (inputType === 'checkbox') {
         $input.prop('checked', false)
-        window.GOVUK.triggerEvent($input[0], 'change', onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input[0], 'change', { detail: { suppressAnalytics: true } })
       } else if (inputType === 'text' || inputType === 'search') {
         /* By padding the haystack with spaces, we can remove the
          * first instance of " $needle ", and this will catch it in
@@ -57,9 +52,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var haystack = ' ' + currentVal.trim() + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
         var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
-        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', { detail: { suppressAnalytics: true } })
       } else if (elementType === 'OPTION') {
-        window.GOVUK.triggerEvent($('#' + removeFilterFacet).val('')[0], 'change', onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($('#' + removeFilterFacet).val('')[0], 'change', { detail: { suppressAnalytics: true } })
       }
     }
 
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -36,7 +36,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       if (inputType === 'checkbox') {
         $input.prop('checked', false)
-        $input.trigger(onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input[0], 'change', onChangeSuppressAnalytics)
       } else if (inputType === 'text' || inputType === 'search') {
         /* By padding the haystack with spaces, we can remove the
          * first instance of " $needle ", and this will catch it in
@@ -57,10 +57,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var haystack = ' ' + currentVal.trim() + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
         var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
-
-        $input.val(newVal).trigger(onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', onChangeSuppressAnalytics)
       } else if (elementType === 'OPTION') {
-        $('#' + removeFilterFacet).val('').trigger(onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($('#' + removeFilterFacet).val('')[0], 'change', onChangeSuppressAnalytics)
       }
     }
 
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/components/expander.js
```diff
@@ -42,8 +39,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
...
       var $el = this.$allInteractiveElements[i]
+      if ($el.tagName === 'SELECT') {
+        $el.addEventListener('change', this.updateSelectedCount.bind(this))
+      }
+
       // but for inputs we need both change and enter key event
       if ($el.tagName === 'INPUT') {
+        if ($el.type === 'text') {
+          $el.addEventListener('change', this.updateSelectedCount.bind(this))
+        }
```
andysellick, line 50: I think this `if` statement isn't necessary - if you follow the logic of the code if this if statement is true and executes, lines 52 and 53 run regardless - and line 49 appears to be identical to line 52, so it runs twice.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -36,7 +35,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       if (inputType === 'checkbox') {
         $input.prop('checked', false)
-        $input.trigger(onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input[0], 'change', suppressAnalytics)
```
andysellick, line 38: If you're extracting element `[0]` of `$input`, that means that `$input` is still a jQuery object. You're going to need to update line 68 inside the `getInput` function.

andysellick: On the same subject, line 19 is also creating a jQuery object: `var $el = $(e.target)`

andysellick: And lines 21-24 need to change from `data` to `getAttribute('data-`

andysellick: And lines 32-34... `prop` is jQuery, and so is `val()`

andysellick: Line 61 also creates a jQuery object - `$('#' + removeFilterFacet)`

andysellick (changes requested):
Sorry, I'm afraid there's still some jQuery in them thar hills...

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -5,8 +5,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict'
 
   GOVUK.Modules.RemoveFilter = function RemoveFilter () {
-    var onChangeSuppressAnalytics = {
-      type: 'change',
+    var suppressAnalytics = {
       suppressAnalytics: true
     }
```
kevindew, line 10: Minor, I don't think it's necessary to have a variable of `suppressAnalytics` that equals `{ suppressAnalytics: true }` I think we should just pass the object literal in as an argument to triggerEvent.

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -57,10 +56,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var haystack = ' ' + currentVal.trim() + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
         var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
-
-        $input.val(newVal).trigger(onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', suppressAnalytics)
```
kevindew, line 59: The suppressAnalytics options will need to be passed in as a detail argument:

```
window.GOVUK.triggerEvent($input[0], 'change', { detail: { suppressAnalytics: true } )
```

Otherwise that argument won't be sent to IE (https://github.com/alphagov/govuk_publishing_components/blob/98aea354ba78e9b242b097fefd97ce015ffe9a69/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js#L15) I also don't think it works on modern browsers either:

```
e = new CustomEvent('change', { detail: { detailOption: true }, nonDetailOption: true })
=> CustomEvent {isTrusted: false, detail: {…}, type: "change", target: null, currentTarget: null, …}
e.detail.detailOption
=> true
e.nonDetailOption
=> undefined
```

Once that change is made we'll need to adapt some other code which receives the event so it can handle a non jQuery event. There's code in this commit that you can use for that handling: https://github.com/alphagov/finder-frontend/commit/efb52d2086fab43db08fc1aaf59aa2556e38d9c6

@andysellick should the parameters option on https://github.com/alphagov/govuk_publishing_components/blob/98aea354ba78e9b242b097fefd97ce015ffe9a69/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js#L5 actually be just a detail option? as that's the only key you could set on the argument that will actually affect IE - any other properties set will be silently dropped.

kevindew (commented):
I've added a couple of other comments - I think there's some event stuff to dig into a bit further too. Nice going though - it's great seeing this progress, I hope you're enjoying learning more about the exciting DOM API.

app/assets/javascripts/modules/remove-filter.js
andysellick: That's true, IE won't receive any other parameters. But we're starting to move away from IE a little, in that more things are going to be a non-JS fallback as the browser support drops. Since other browsers can understand other parameters, I think we should keep that script as it is for now, but we may need to raise awareness among the community of this limitation.

kevindew: Gotcha, I'd have thought that if you wanted to do something IE didn't support then you'd probably use CustomEvent directly and you'd only use that function if you wanted it to be IE compatible. It could be a bit surprising for other CustomEvent usages since it sets bubbles and cancelable which you might not expect.

andysellick: Yeah, the script could do with a bit of work. It's only being used in a few places right now, but something I need to monitor and expand on.

app/assets/javascripts/modules/remove-filter.js
leenagupte: So: `window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', suppressAnalytics: true)` ?

kevindew: It'll be `window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', { detail: { suppressAnalytics: true } })` (as per: https://github.com/alphagov/finder-frontend/pull/2532#discussion_r649807087)

You also can't omit the curly braces like Ruby lets you.

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: So i might be misunderstanding things here. Do these bits need to be done for this PR? I was only trying to remove jQuery from the expander. The changes to `remove-filter` were needed to get it working because I'd changed an event listener to be JS and it couldn't hear jQuery triggers. These comments seem to be related to removing jQuery from the `remove-filter` file which isn't something I was trying to do in this PR. Planning on doing that later

app/assets/javascripts/components/expander.js
JonathanHallam: good shout, line 49 and 52 aren't quite identical but the `handleInputEvent` on 52 calls `updateSelectedCount` so we're all good without the `if` :) 🎉

app/assets/javascripts/modules/remove-filter.js
andysellick: Oh I see... sorry. Got confused. That's fine then, although if you're already doing part of it and you're going to do it anyway, I'd go for including it in this PR as well, but up to you.

JonathanHallam: yeah, i think I'd prefer to merge as is just to keep each PR as nuclear as possible

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: Sorted :)

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: Also sorted :)

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -57,10 +52,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var haystack = ' ' + currentVal.trim() + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
         var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
-
-        $input.val(newVal).trigger(onChangeSuppressAnalytics)
+        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', { detail: { suppressAnalytics: true } })
```
owenatgov, line 55: JQuery spotted❗ 👀 ❗ I'm trying to understand what's happening here and struggling. The way jQuery applies new input values is using a function call instead of the declaration that plain js uses (so `$input.value = newVal` instead of `$input.val(newVal)`). It looks like we're passing the output of the jQuery input value update function to the trigger event? This feels a bit grubby, and also looks like we're hiding the update of an input value in here. If there's an explanation I'm all ears, otherwise **get that jQuery out of here**.

owenatgov (commented):
One thing that looks a bit sus, otherwise looks lovely!

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: Hi Owen, basically this PR is only concerned with the expander. The changes made here to the remove-filter are the absolute bare minimum that this needs to pass. I've got [a branch](https://github.com/alphagov/finder-frontend/compare/remove-filter) that's based on this one that's going to properly address all of the jQuery in this file separately. I promise all of the jQuery is going :D

owenatgov: I seeeeeeee. Cool! Will drop a tick on this now.

owenatgov (approved):

PR #2556, opened 2021-06-22
Remove jQuery from the remove-filter

JonathanHallam:
## What

Remove jQuery from the remove-filter file.

## Why

## Trello

https://trello.com/c/jK6hrK3d/417-remove-jquery-from-finder-frontend

### DIFF
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -16,12 +16,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     function toggleFilter (e) {
       e.preventDefault()
       e.stopPropagation()
-      var $el = $(e.target)
+      var $el = e.target
 
-      var removeFilterName = $el.data('name')
-      var removeFilterValue = $el.data('value')
-      var removeFilterLabel = $el.data('track-label')
-      var removeFilterFacet = $el.data('facet')
+      var removeFilterName = $el.getAttribute('data-name')
+      var removeFilterValue = $el.getAttribute('data-value')
+      var removeFilterLabel = $el.getAttribute('data-track-label')
+      var removeFilterFacet = $el.getAttribute('data-facet')
 
       var $input = getInput(removeFilterName, removeFilterValue, removeFilterFacet)
       fireRemoveTagTrackingEvent(removeFilterLabel, removeFilterFacet)
@@ -29,9 +29,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
 
     function clearFacet ($input, removeFilterValue, removeFilterFacet) {
-      var elementType = $input.prop('tagName')
-      var inputType = $input.prop('type')
-      var currentVal = $input.val()
+      var elementType = $input.tagName
+      var inputType = $input.type
+      var currentVal = $input.value
 
       if (inputType === 'checkbox') {
         $input.prop('checked', false)
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -66,9 +66,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
 
     function getInput (removeFilterName, removeFilterValue, removeFilterFacet) {
-      var selector = (removeFilterName) ? " input[name='" + removeFilterName + "']" : " [value='" + removeFilterValue + "']"
+      var selector = (removeFilterName) ? "input[name='" + removeFilterName + "']" : "[value='" + removeFilterValue + "']"
+      var element = document.getElementById(removeFilterFacet)
 
-      return $('#' + removeFilterFacet).find(selector)
+      return element.querySelector(selector)
     }
 
     function fireRemoveTagTrackingEvent (filterValue, filterFacet) {
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -6,7 +6,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   GOVUK.Modules.RemoveFilter = function RemoveFilter () {
     this.start = function (element) {
-      $(element).on('click', '[data-module="remove-filter-link"]', toggleFilter)
+      element[0].addEventListener('click', function (e) {
+        if (e.target.getAttribute('data-module') === 'remove-filter-link') {
+          toggleFilter(e)
+        }
+      })
     }
 
     function toggleFilter (e) {
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -34,8 +34,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var currentVal = $input.value
 
       if (inputType === 'checkbox') {
-        $input.prop('checked', false)
-        window.GOVUK.triggerEvent($input[0], 'change', { detail: { suppressAnalytics: true } })
+        $input.checked = false
+        window.GOVUK.triggerEvent($input, 'change', { detail: { suppressAnalytics: true } })
       } else if (inputType === 'text' || inputType === 'search') {
         /* By padding the haystack with spaces, we can remove the
          * first instance of " $needle ", and this will catch it in
@@ -56,9 +56,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var haystack = ' ' + currentVal.trim() + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
         var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
-        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', { detail: { suppressAnalytics: true } })
+        $input.value = newVal
+        window.GOVUK.triggerEvent($input, 'change', { detail: { suppressAnalytics: true } })
       } else if (elementType === 'OPTION') {
-        window.GOVUK.triggerEvent($('#' + removeFilterFacet).val('')[0], 'change', { detail: { suppressAnalytics: true } })
+        var element = document.getElementById(removeFilterFacet)
+        element.value = ''
+        window.GOVUK.triggerEvent(element, 'change', { detail: { suppressAnalytics: true } })
       }
     }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/remove-filter.js
```diff
@@ -6,7 +6,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   GOVUK.Modules.RemoveFilter = function RemoveFilter () {
     this.start = function (element) {
-      $(element).on('click', '[data-module="remove-filter-link"]', toggleFilter)
+      element[0].addEventListener('click', function (e) {
+        var target = e.target
+        if (target.getAttribute('data-module') === 'remove-filter-link') {
```
kevindew, line 11: Nitpick: We can save a line here

```suggestion
        if (e.target.getAttribute('data-module') === 'remove-filter-link') {
```

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -57,9 +57,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var haystack = ' ' + currentVal.trim() + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
         var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
-        window.GOVUK.triggerEvent($input.val(newVal)[0], 'change', { detail: { suppressAnalytics: true } })
+        $input.value = newVal
+        window.GOVUK.triggerEvent($input, 'change', { detail: { suppressAnalytics: true } })
       } else if (elementType === 'OPTION') {
-        window.GOVUK.triggerEvent($('#' + removeFilterFacet).val('')[0], 'change', { detail: { suppressAnalytics: true } })
+        var element = document.getElementById(removeFilterFacet)
```
kevindew, line 63: Are we sure this element will always exist? should we have a conditional so we don't try set value on null

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -67,9 +67,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
 
     function getInput (removeFilterName, removeFilterValue, removeFilterFacet) {
-      var selector = (removeFilterName) ? " input[name='" + removeFilterName + "']" : " [value='" + removeFilterValue + "']"
+      var selector = (removeFilterName) ? "input[name='" + removeFilterName + "']" : "[value='" + removeFilterValue + "']"
+      var element = document.getElementById(removeFilterFacet)
 
-      return $('#' + removeFilterFacet).find(selector)
+      return element.querySelector(selector)
```
kevindew, line 73: Similar question here about our certainty of the elements existence. jQuery is very forgiving about elements not existing:

```
> jQuery('asdfasdf').find('adfasd').toArray()
[]
> document.getElementById('asdfasdf').querySelector('adfasd').toArray()
VM801:1 Uncaught TypeError: Cannot read property 'querySelector' of null
    at <anonymous>:1:36
```

app/assets/javascripts/modules/remove-filter.js
```diff
@@ -54,9 +54,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
          * Just removing " beta " from the haystack would result in
          * "alphagamma", which is wrong.
          */
-        var haystack = ' ' + currentVal.trim() + ' '
+        var haystack = ' ' + currentVal.replace(/^\s+|\s+$/g, '') + ' '
         var needle = ' ' + decodeEntities(removeFilterValue.toString()) + ' '
-        var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').trim()
+        var newVal = haystack.replace(needle, ' ').replace(/ {2}/g, ' ').replace(/^\s+|\s+$/g, '')
```
kevindew: We're not removing jQuery methods here, this is is JS String.prototype.trim, so can remain.

kevindew (commented):
Looks great @JonathanHallam - a few questions and some suggestions

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: As far as I've been able to manually test yes. This has to exist for the thing to be clicked in the first place. It does cause errors if it doesn't exist (as with your other comment) which is why i had to add [this commit](https://github.com/alphagov/finder-frontend/pull/2556/commits/7cc4280cb4da67f89a6d89f018a826f06a322d2d) but I haven't seen it when spinning the app up locally and the tests don't suggest that it can happen.

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: nice spot, was being too literal, fixed :)

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: oh right, okay, thanks :) I've dropped the last commit

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: This caused me such a headache, couldn't work out why the tests were broken. As far as I know these will always exist, I've not been able to break it when running the app locally

spec/javascripts/modules/remove-filter-spec.js
```diff
@@ -37,14 +37,14 @@ describe('remove-filter', function () {
 
   var $dropdown = $(
     '<div data-module="remove-filter">' +
-      '<button href="/search/news-and-communications?[][]=level_one_taxon&amp;[][]=ba3a9702-da22-487f-86c1-8334a730e559&amp;[][]=level_two_taxon&amp;[][]" class="remove-filter" role="button" aria-label="Remove filter Entering and staying in the UK" data-module="remove-filter-link" data-facet="level_one_taxon" data-value="ba3a9702-da22-487f-86c1-8334a730e559" data-track-label="Entering and staying in the UK" data-name="">✕</button>' +
+      '<button href="/search/news-and-communications?[][]=dropdown&amp;[][]=ba3a9702-da22-487f-86c1-8334a730e559&amp;[][]=level_two_taxon&amp;[][]" class="remove-filter" role="button" aria-label="Remove filter Entering and staying in the UK" data-module="remove-filter-link" data-facet="dropdown" data-value="ba3a9702-da22-487f-86c1-8334a730e559" data-track-label="Entering and staying in the UK" data-name="">✕</button>' +
```
andysellick, line 40: What's the reason behind this change? It seems like it's changing the name of something, which is reflected further on in the file, but I can't see why.

spec/javascripts/modules/remove-filter-spec.js
```diff
@@ -54,10 +54,24 @@ describe('remove-filter', function () {
...
+      '</select>' +
+    '</div>' +
+    '<div id="level_two_taxon">' +
+      '<select name="level_two_taxon">' +
+        '<option value="">All topics</option>' +
+        '<option value="bb3a9702-da22-487f-86c1-8334a730e559">Entering and staying in the UK</option>' +
+      '</select>' +
+    '</div>' +
+    '<div>' +
+      '<select name="dropdown" id="dropdown">' +
```
andysellick, line 70: This looks like a duplication of the 'All topics' select element. In reality I think there'd only ever be two selects, but there are three here now - why?

andysellick (commented):
Looks good, just some questions about changes to the tests 👍

app/assets/javascripts/modules/remove-filter.js
kevindew: Thanks 👍 

I've been doing some reading of this to try form an understanding of what could go wrong. 

This `getInput` method strikes me as something that we should be prepared for the element not to exist. As it depends on another element existing in the facet. In fact I think I've found an example where it would fail: https://www.gov.uk/search/all?organisations%5B%5D=department-for-international-trade&order=updated-newest&parent=department-for-international-trade where there's an ability to filter on facets that aren't in the facet column (link originates from "See all latest documents" on [an organisation](https://www.gov.uk/government/organisations/department-for-international-trade)). 

I think you're ok on not adding a check to https://github.com/alphagov/finder-frontend/pull/2556/files#r657362674 as the element has to exist to get past this code (a nice optimisation would be to only look it up once)

I don't think GOV.UK has really formed a strategy for deciding when to care if elements exist or not (seems a fresh post jQuery problem) but to me it seems reasonable to assume elements exist if they're children of the component/module, and to always be cautious with anything beyond that. Does that make sense/sounds reasonable?

JonathanHallam: So i think the input actually refers to the object that you're clicking on to remove. In the example that you posted `element.querySelector(selector)` finds `<input type="text" name="organisations[]" id="organisations_0" value="department-for-international-trade">` so it manages to find it even though there's no parent item on the page.

spec/javascripts/modules/remove-filter-spec.js
JonathanHallam: it's basically because all three elements need to exist independently to make the tests pass. Because of the move to JS we now have to really care if the elements exist on the page before removing them otherwise it falls over. This is so that all three elements (`dropdown`, `level_one_taxon` and `level_two_taxon`) exist separately and can be used by the tests

spec/javascripts/modules/remove-filter-spec.js
JonathanHallam: it's to keep the tests happy, basically the same reason as the above comment. They don't need to be `selects` but there needs to be some element with the correct `name` and `id` for the tests not to fail

andysellick (approved):
I think I'm happy with this now to approve it, but I'd like @kevindew to also approve before it gets merged.

spec/javascripts/modules/remove-filter-spec.js
JonathanHallam: Fixed after a chat :)

spec/javascripts/modules/remove-filter-spec.js
JonathanHallam: Fixed after a chat :)

app/assets/javascripts/modules/remove-filter.js
JonathanHallam: We've spoken about this offline and I've updated the 1st commit message to reflect our findings

kevindew (approved):
Looks good to me too. Thanks @JonathanHallam and @andysellick 🎉

PR #2576, opened 2021-07-07
Fix duplicated banner on brexit checker results page.

danacotoran:
There is an edge case on the Brexit checker results page, where if the user must NOT reauthenticate (i.e they're already logged in), and their results differ from the ones already stored in their account, the "These results are different from the ones stored in your account" message appears twice. This is due to the same conditions being present in the view at two points, [here](https://github.com/alphagov/finder-frontend/blob/6f9a17f291ddb5fa788bdb2273272e7e50d4fdc3/app/views/brexit_checker/results.html.erb#L72) and [here](https://github.com/alphagov/finder-frontend/blob/6f9a17f291ddb5fa788bdb2273272e7e50d4fdc3/app/views/brexit_checker/results.html.erb#L90).

image

This removes the second message, so that the notice will only appear once in those specific conditions.

-----

### DIFF
--- a/app/views/brexit_checker/results.html.erb
+++ b/app/views/brexit_checker/results.html.erb
@@ -86,17 +86,8 @@
     <%= render 'results_citizen_actions', citizen_results_groups: @presenter.citizen_results_groups if @presenter.citizen_results_groups.any? %>
     <%= render 'results_with_no_actions', criteria: @presenter.criteria if (@presenter.business_results.empty? && @presenter.citizen_results_groups.empty?) %>
 
-    <% if !must_reauthenticate? %>
-      <% if @results_differ %>
-        <%= render 'govuk_publishing_components/components/notice', {
-          title: t('brexit_checker.results.accounts.results_differ.message'),
-          description: sanitize(t('brexit_checker.results.accounts.results_differ.description', link: transition_checker_save_results_confirm_path(c: criteria_keys)))
-        } %>
-      <% end %>
-    <% else %>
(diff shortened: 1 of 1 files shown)


barrucadu:
Showing the banner both at the top and bottom of the results list was intentional - because if there's a bunch of results you can't see the banner-at-the-top when you scroll down to the bottom.

But it doesn't look good when there *isn't* more than a page of results.  We could get rid of one of the banners, but it would be good to check with a designer first.

danacotoran:
> Showing the banner both at the top and bottom of the results list was intentional - because if there's a bunch of results you can't see the banner-at-the-top when you scroll down to the bottom.
> 
> But it doesn't look good when there _isn't_ more than a screen of results. We could get rid of one of the banners, but it would be good to check with a designer first.

@barrucadu Ah. Thought I was missing something. My bad, I completely thought this was introduced twice by accident. 
I'm not quite sure I understand the logic of it though... because since the banner at the top is before the actions, it is always seen first right? Presumably people would take note of it. Is the implication that people might either:
1. skip over it
2. read it but forget about it once they've read their results (which introduces the need to have that information duplicated after the results to make sure they don't miss it)?

Perhaps we can do something with the logic here so that the second notice only appears when there's a lengthy list of results.

huwd (approved):

danacotoran:
Closed for now, because it was determined this was on purpose and we probably don't want to do something fancy like only show the second banner when the list of results is long enough.

PR #2578, opened 2021-07-12
Remove the Transition Checker email page from account journey

barrucadu:
We've decided that, if the user signs up via an account, they should
get email notifications by default.  This is because it's a bit
strange to ask the user if they want an account OR email notifications
and then, after they choose account, ask them again if they want email
notifications.

See also https://github.com/alphagov/govuk-account-manager-prototype/pull/899

---

[Trello card](https://trello.com/c/JuzRNvA4/832-remove-the-email-signup-from-the-transition-checker-registration-journey)


danacotoran (approved):
This seems to do what it's meant to do and looks good to me, but it would probably be best to get another (backend) dev to sense check some of this stuff.

huwd (approved):

PR #2587, opened 2021-07-19
Fix js tests in ie10

andysellick:

## What
Make Javascript tests pass in IE10+.

## Why
Part of a wider effort to ensure our tests pass in older but still supported browsers.

## Visual changes
None.

Trello card: https://trello.com/c/CAkmuuTK/424-make-js-tests-pass-in-all-browsers


spec/javascripts/modules/remove-filter-spec.js
```diff
@@ -79,6 +79,7 @@ describe('remove-filter', function () {
   })
 
   afterEach(function () {
+    $('.js-remove').remove()
```
alex-ju, line 82: Not familiar with the filter, so take my suggestion with a pinch of salt, but could we remove the whole `$facets` from the DOM, instead of one element at a time?

app/assets/javascripts/components/option-select.js
```diff
@@ -243,7 +243,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     // check whether this is hidden by progressive disclosure,
     // because height calculations won't work
-    if (this.$optionsContainer.offsetParent === null) {
+    // would use offsetParent === null but for IE10+
+    var parent = this.$optionSelect.parentElement
+    var parentIsHidden = !(parent.offsetWidth || parent.offsetHeight || parent.getClientRects().length)
+    if (parentIsHidden) {
```
alex-ju, line 249: this seems twisted. to me, `offsetParent` being `null` could mean many things: the element itself or its parent have `display` set to `none`, or the element's `position` is set to `fixed` which makes the previous conditional statement unclear. now, I don't think we're making it any clear with this update.

can we instead look for the parent optionSelect and check if it's open or not? I see some `js-opened` `js-closed` classes that we can potentially check. may not be the best suggestion, but I'm looking at ways we can simplify this because it feels that we're moving from obscure to twisted.

alex-ju (commented):
Looks good overall, except with two areas where I think we can improve things.

spec/javascripts/modules/remove-filter-spec.js
andysellick: Yes, I wanted to do that, but I tried `$facets.remove()` and it throws a `$facets.remove is not a function` error for every test. I don't know why that doesn't work, I thought it should do. Previously the test wasn't cleaning up after itself at all.

app/assets/javascripts/components/option-select.js
andysellick: Totally with you on the obscureness here. Unfortunately we can't check the parent optionSelect - on mobile the search filters are inside a panel that appears/disappears like a modal, and the visibility is toggled way up the DOM.

Annoyingly IE10 supports the `offsetParent` null check, but it wasn't working in this test (although I wrote a quick test to check it and it should pass, but for some reason doesn't when run on the actual code). I replaced it with this because it seemed more accurate - it's actually checking that the parent element is hidden, rather than hidden or fixed, as you point out.

Really not sure what else to try with this - open to suggestions.

alex-ju (approved):

PR #2589, opened 2021-07-22
Replace jQuery in module initialisers

Rosa-Fox:
[Trello](https://trello.com/c/02ot7gbP/482-convert-finder-frontend-module-initialisers-to-not-use-jquery)

# What? 
Replaces jQuery module initialisers with vanilla JavaScript for `expander` and `option-select`. 

See [this PR description](https://github.com/alphagov/govuk_publishing_components/pull/2095) for some nice examples of module initialisation. 

# Why?
We're using an old and unsupported version of jQuery for browser support reasons.

Rather than upgrade, it's far better to remove our dependence.

### Expander

expander

### Option Select

option-select

Co-authored-by: Andy Sellick @andysellick

### DIFF
--- a/spec/javascripts/components/expander-spec.js
+++ b/spec/javascripts/components/expander-spec.js
@@ -2,7 +2,7 @@ describe('An expander module', function () {
   'use strict'
 
   var $element
-  var expander
+
   /* eslint-disable */
   var html = '\
     <div class="app-c-expander" data-module="expander">\
@@ -19,9 +19,9 @@ describe('An expander module', function () {
 
   describe('in default mode', function () {
     beforeEach(function () {
-      expander = new GOVUK.Modules.Expander()
-      $element = $(html)
-      expander.start($element)
+      $element = document.createElement('div')
+      $element.innerHTML = html
+      new GOVUK.Modules.Expander($element.querySelector('.app-c-expander')).init()
     })
 
     afterEach(function () {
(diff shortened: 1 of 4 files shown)


spec/javascripts/components/expander-spec.js
```diff
@@ -19,62 +19,63 @@ describe('An expander module', function () {
 
   describe('in default mode', function () {
     beforeEach(function () {
-      expander = new GOVUK.Modules.Expander()
-      $element = $(html)
-      expander.start($element)
+      $element = document.createElement('div')
+      $element.innerHTML = html
+      new GOVUK.Modules.Expander($element.querySelector('.app-c-expander')).init()
```
kevindew, line 24: Not a blocker but you could have got away with a simpler change to these tests that just changed these beforeEach blocks.

E.g:

```diff
-      expander = new GOVUK.Modules.Expander()
       $element = $(html)
-      expander.start($element)
+      expander = new GOVUK.Modules.Expander($element[0])
+      expander.init()
```

I don't think you should change the code you've commited, as I think it's better with the lower jQuery usage but I can imagine it adds up some extra work in completing this task and reviewing the changes, so it shouldn't be a necessary step in each of these conversions.

kevindew (approved):
Nice one @Rosa-Fox - looks good.

spec/javascripts/components/expander-spec.js
Rosa-Fox: thanks! I will keep that in mind for the next module initialiser conversions.

PR #2595, opened 2021-07-30
Remove enable-aria-controls JS module

kevindew:
Trello: https://trello.com/c/n8V5lG8F/518-convert-finder-frontend-module-initialisers-in-enable-aria-controlsjs-to-not-use-jquery

This is being removed as this code no longer does anything. It
previously used to set the aria-controls attribute dynamically on
elements and was replaced in 2018 [1] [2] (which seems an odd decision
I'm not sure why it wasn't just part of the HTML 🤔). Since [2] was merged
there is no code with a data-aria-controls attribute anymore so this
module was essentially a no-op.

[1]: https://github.com/alphagov/finder-frontend/commit/2744c07ef25e2c1edb53dea2ae62ee6c5c64a774#diff-f90fee1f2d7d2b2341b403e10727271b70bd61e365b4ea19e7df1e9fcf789e19
[2]: https://github.com/alphagov/finder-frontend/commit/af5b72012911bc2d84ec4af86811f5109d063d6a#diff-f90fee1f2d7d2b2341b403e10727271b70bd61e365b4ea19e7df1e9fcf789e19


leenagupte (approved):

andysellick:
Historical information - I think the original intention of this script was to apply the aria-controls attribute only when JS was enabled. If JS was disabled, aria-controls would be lying as dynamic updates of the page wouldn't work.

kevindew:
> Historical information - I think the original intention of this script was to apply the aria-controls attribute only when JS was enabled. If JS was disabled, aria-controls would be lying as dynamic updates of the page wouldn't work.

Thanks - that makes some. I was pretty baffled why this used JS.

PR #2596, opened 2021-08-03
Convert remove-filter.js to use init initialisation approach

kevindew:
Trello: https://trello.com/c/DdSnmUQy/522-convert-finder-frontend-module-initialisers-in-remove-filterjs-to-not-use-jquery

This replaces the jQuery start initialisation pattern in the module to use the non-jQuery init approach, to help us remove jQuery. It then refactors the module organisation to be the standard GOV.UK approach. Finally it does some clean-up in the tests and removes the jQuery from that too.

This is easiest reviewed commit by commit.

### DIFF
--- a/app/assets/javascripts/modules/remove-filter.js
+++ b/app/assets/javascripts/modules/remove-filter.js
@@ -1,100 +1,102 @@
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
-(function (global, GOVUK) {
+(function (Modules) {
   'use strict'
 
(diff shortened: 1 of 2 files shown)


andysellick (approved):
also 🚀 

Extra points for removing jQuery from the test - you don't have to do that if it's not necessary.

kevindew:
> also 🚀

Thanks for the review 👍  

> Extra points for removing jQuery from the test - you don't have to do that if it's not necessary.

Yup, that wasn't an intentional project it just all seemed a bit of a mess - lots of `$('<div>')[0]` code

PR #2598, opened 2021-08-03
Fix feature test

JonathanHallam:
This requires the explicit clearing of the input box for the test to work, otherwise the new value is added along with the existing value, resulting in 'breakfastlunch' instead of the intended 'lunch'


leenagupte (approved):
That's a bit of a gnarly issue.  Well done for figuring it out 💯

PR #2604, opened 2021-08-09
replace start with init for ecommerce in live search

JonathanHallam:
As it turns out init is defined on the [Ecommerce variable in publishing components](https://github.com/alphagov/govuk_publishing_components/blob/e3b952e0365c8174da7ed7e8e964951bd3cbc074/app/assets/javascripts/govuk_publishing_components/analytics/ecommerce.js#L12) so this should be fine.

This is part of a wider effort to remove jQuery from the stack because the version we're using is outdated

Trello - https://trello.com/c/pATsfsuD/524-convert-finder-frontend-module-initialisers-in-livesearchjs-to-not-use-jquery

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -99,7 +99,7 @@
   LiveSearch.prototype.startEnhancedEcommerceTracking = function startEnhancedEcommerceTracking () {
     this.$resultsWrapper.attr('data-search-query', this.currentKeywords())
     this.$suggestionsBlock.attr('data-search-query', this.currentKeywords())
-    if (GOVUK.Ecommerce) { GOVUK.Ecommerce.start() }
+    if (GOVUK.Ecommerce) { GOVUK.Ecommerce.init() }
   }
 
   LiveSearch.prototype.getTaxonomyFacet = function getTaxonomyFacet () {


app/assets/javascripts/live_search.js
```diff
@@ -99,7 +99,7 @@
   LiveSearch.prototype.startEnhancedEcommerceTracking = function startEnhancedEcommerceTracking () {
     this.$resultsWrapper.attr('data-search-query', this.currentKeywords())
     this.$suggestionsBlock.attr('data-search-query', this.currentKeywords())
-    if (GOVUK.Ecommerce) { GOVUK.Ecommerce.start() }
+    if (GOVUK.Ecommerce) { GOVUK.Ecommerce.init() }
```
andysellick, line 102: I'm not sure this change is needed - `start` and `init` do different things, and the Ecommerce code isn't a module, it just happens to have functions named `start` and `init`.

andysellick (changes requested):

app/assets/javascripts/live_search.js
JonathanHallam: I have misunderstood how this works, shall I just close this then?

andysellick: The `start` function accepts an element, but it's optional (as seen in this instance) as it can find its own element. But it then calls the `init` function using that element, so if we called `init` instead of `start`, I don't think the code would do anything.

I'm not suggesting we need to change this to pass an element as I think calling the `start` function is the way the code is intended to be called, I'm just looking through it and trying to make sense of it 🤔

andysellick: For future reference, there are several indicators of whether a bit of code is part of our modules system or not.

- on line 100 (https://github.com/alphagov/govuk_publishing_components/blob/e3b952e0365c8174da7ed7e8e964951bd3cbc074/app/assets/javascripts/govuk_publishing_components/analytics/ecommerce.js#L100) it adds the Ecommerce object to the `window.GOVUK` object, modules are kept in `window.GOVUK.Modules`.
- modules require no code to execute themselves, this is all handled centrally by the modules initialiser (https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/modules.js). Modules should have a `start` or `init` function that accepts an element to act on. In this case the Ecommerce code finds an element to act on by itself, in the `start` function.

JonathanHallam:
Closing this after working out it's not needed

PR #2609, opened 2021-08-10
Return A/B response headers for all variants

injms:
## What

Sets the `Vary` header to be returned whatever the variant being view is.

Tests have been updated to check that this is working correctly, with checks for the Vary header, the meta tag and the correct Slimmer template header.

## Why
The `Vary` header was only being returned for the "B" variant - which makes it tricky for Fastly to know how to cache the pages correctly. This has been updated to always pass the `Vary` header back to Fastly regardless of the variant.

---


sihugh (approved):
Looks great!

PR #2676, opened 2021-11-04
Add notice to indicate closure of the checker

sihugh:
This will be displayed on all question pages and the results page. This is a bit invasive, but it's important to flag, because people will have to think about their options if they want to keep getting Brexit emails.

Screenshot 2021-11-09 at 09 53 45
Screenshot 2021-11-09 at 09 53 40


hannako (approved):
👍

PR #2691, opened 2021-11-30
Redirect Brexit checker requests

sihugh:
We're retiring the Brexit checker today. This will send requests for checker paths to the Brexit home page.

I'm using a 302 status "Found" even though we expect this to be permanent because it allows us to backtrack easily should we need to. If we use 301 off the bat, then clients may well have cached the redirect already and
we'll be stuck.

We'll remove the plumbing of the checker at a later date.


sihugh:
> Are you leaving in the rake tasks and associated code to import checker stuff for now?

yep, it's a bit messy so wanted to leave in as much as possible for now to make reviewing easier

PR #2693, opened 2021-12-01
Remove Brexit checker code

sihugh:
We [retired the checker yesterday](https://github.com/alphagov/finder-frontend/pull/2691), so this code is no longer needed. Fortunately, most of the files have `brexit_checker` in the path which makes it easier to delete/review. 

The facility to send bulk emails still exists in Email Alert API, so we can use that to send things should we need to.

The latter two commits remove other dead code that I found (which got the deleted lines of code count over 6000!)


huwd (commented):
Can this go too?
https://github.com/alphagov/finder-frontend/blob/remove-checker-code/app/views/development/index.html.erb#L10-L15

hannako (approved):

huwd:
I guess these can go too:
https://github.com/alphagov/finder-frontend/blob/main/public/robots.txt#L7

huwd:
Humm not sure about these two, are they related to this or something else "QA"?

- https://github.com/alphagov/finder-frontend/blob/remove-checker-code/spec/javascripts/modules/track-brexit-qa-choices.spec.js
- https://github.com/alphagov/finder-frontend/blob/remove-checker-code/app/assets/javascripts/modules/track-brexit-qa-choices.js
-

huwd:
🤔 Humm this seems to referene get-ready-for-brexit emails which was ours.
But the thing it tests looks generic:
https://github.com/alphagov/finder-frontend/blob/remove-checker-code/spec/lib/services/email_alert_api_spec.rb

I can't work out if this file can go... or if we can just change the let example to be something else you can subscribe to through a finder?

huwd:
This line can go too I think:
https://github.com/alphagov/finder-frontend/blob/remove-checker-code/docs/pull_request_template.md?plain=1#L9

huwd:
Also this CSS:
https://github.com/alphagov/finder-frontend/blob/remove-checker-code/app/assets/stylesheets/print.scss#L55-L58

huwd:
Last but not least:
The readme:
- [here](https://github.com/alphagov/finder-frontend/blob/c09e456606f9d0515eb435f4e9b21a3bfc272317/README.md?plain=1#L9)

sihugh:
> 🤔 Humm this seems to referene get-ready-for-brexit emails which was ours. But the thing it tests looks generic: https://github.com/alphagov/finder-frontend/blob/remove-checker-code/spec/lib/services/email_alert_api_spec.rb
> 
> I can't work out if this file can go... or if we can just change the let example to be something else you can subscribe to through a finder?

I'll have a look at this separately if that's OK? Good shout though, and top spotting!

huwd (approved):

PR #2694, opened 2021-12-02
Remove temporary redirects for Brexit checker paths

sihugh:
I have replaced the previous content items with [redirects to /brexit](https://www.gov.uk/api/content/transition-check/questions) so there's no need for these to be here any more.

I originally put these redirects in because the 302 response was uncached which was useful if we'd changed our minds and wanted to roll back quickly. The [301s](https://httpstatusdogs.com/301-moved-permanently) that redirect content items produce are cached for half an hour and also indicate that the thing has actually gone somewhere else for good.


hannako (approved):
👍  Satisfying tidying up!

PR #2706, opened 2021-12-24
Honour Content Store TTL

bilbof:
Content Store provides a TTL in the Cache-Control response header which frontend apps should respect.

This ensures that this app honours the cache TTL provided by Content Store. There is a 5 minute TTL on the `FindersController` which I have left as-is, since we want content in search to be fresh and the current TTL provided by Content Store is 30 minutes.

This partially implements RFC 144.

Links:
* Content Store TTL: https://github.com/alphagov/content-store/blob/9661e2047f2167f87066fbb6ef38da71668a1327/app/controllers/content_items_controller.rb#L14
* RFC 144: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-144-consistent-cache-expiry.md

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,11 +1,13 @@
 class SearchController < ApplicationController
   layout "search_layout"
-  before_action :set_expiry
 
   def index
     search_params = SearchParameters.new(params)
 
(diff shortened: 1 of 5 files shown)


bilbof:
Superceded by https://github.com/alphagov/finder-frontend/pull/2713

PR #2713, opened 2022-01-20
Honour content store TTL

kentsanggds:
Content Store provides a TTL in the Cache-Control response header which frontend apps should respect.

This ensures that this app honours the cache TTL provided by Content Store. There is a 5 minute TTL on the FindersController which I have left as-is, since we want content in search to be fresh and the current TTL provided by Content Store is 30 minutes.

This partially implements RFC 144.

Links:

Content Store TTL: https://github.com/alphagov/content-store/blob/9661e2047f2167f87066fbb6ef38da71668a1327/app/controllers/content_items_controller.rb#L14
RFC 144: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-144-consistent-cache-expiry.md

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -2,10 +2,9 @@ class FindersController < ApplicationController
   layout "finder_layout"
 
   before_action do
-    expires_in(5.minutes, public: true)
+    set_expiry(content_item)
   end
 
-  ATOM_FEED_MAX_AGE = 300
   def show
     respond_to do |format|
       format.html do
@@ -27,7 +26,6 @@ class FindersController < ApplicationController
           redirect_to_destination
         else
           @search_query = initialize_search_query(is_for_feed: true)
-          expires_in(ATOM_FEED_MAX_AGE, public: true)
           @feed = AtomPresenter.new(content_item, results, facet_tags)
         end
       end
@@ -77,10 +75,6 @@ private
 
   def render_component(partial, locals)
     (render_to_string(formats: %w[html], partial: partial, locals: locals) || "").squish
-  end
-
-  def content_item
-    @content_item ||= ContentItem.from_content_store(finder_base_path)
   end
 
   def result_set_presenter
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,11 +1,11 @@
 class SearchController < ApplicationController
   layout "search_layout"
-  before_action :set_expiry
 
   def index
     search_params = SearchParameters.new(params)
 
-    @content_item = Services.cached_content_item("/search")
+    @content_item = ContentItem.from_content_store("/search")
+    set_expiry(@content_item)
 
     # Redirect all requests to all content finder, where either search params have been supplied or the user is
     # requesting the JSON endpoint.
@@ -17,12 +17,6 @@ class SearchController < ApplicationController
   end
 
 protected
-
-  def set_expiry(duration = 30.minutes)
-    unless Rails.env.development?
-      expires_in(duration, public: true)
-    end
-  end
 
   def fill_in_slimmer_headers(result_count)
     set_slimmer_headers(
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -15,7 +15,13 @@ module Services
   def self.cached_content_item(base_path)
     Rails.cache.fetch("finder-frontend_content_items#{base_path}", expires_in: 5.minutes) do
       GovukStatsd.time("content_store.fetch_request_time") do
-        content_store.content_item(base_path).to_h
+        content_item = content_store.content_item(base_path)
+        content_item_hash = content_item.to_h
+        content_item_hash["cache_control"] = {
+          "max_age" => content_item.cache_control["max-age"],
+          "public" => !content_item.cache_control.private?,
+        }
+        content_item_hash
       end
     end
   end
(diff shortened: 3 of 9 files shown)


app/controllers/search_controller.rb
```diff
@@ -1,11 +1,13 @@
 class SearchController < ApplicationController
   layout "search_layout"
-  before_action :set_expiry
 
   def index
     search_params = SearchParameters.new(params)
 
-    @content_item = Services.cached_content_item("/search")
+    content_item = Services.cached_content_item("/search")
+    @content_item = content_item.to_h
```
bilbof, line 8: Is the `content_item` already a hash here?

app/lib/services.rb
```diff
@@ -13,9 +13,14 @@ def self.content_store
   end
 
   def self.cached_content_item(base_path)
-    Rails.cache.fetch("finder-frontend_content_items#{base_path}", expires_in: 5.minutes) do
+    content_item = content_store.content_item(base_path)
```
bilbof, line 16: You're skipping the cache here - the call to content store needs to be within the `Rails.cache.fetch(...) {}` block otherwise we'll call the content store every time.

app/lib/services.rb
```diff
@@ -13,9 +13,14 @@ def self.content_store
...
+    max_age = content_item.cache_control["max-age"]
+
+    Rails.cache.fetch("finder-frontend_content_items#{base_path}", expires_in: max_age) do
```
bilbof, line 19: You can keep `max_age` as 5 minutes here - we don't mind how long we cache content items, it's just how long we cache the whole page that matters.

app/lib/services.rb
```diff
@@ -13,9 +13,14 @@ def self.content_store
...
       GovukStatsd.time("content_store.fetch_request_time") do
-        content_store.content_item(base_path).to_h
+        content_item_hash = content_item.to_h
+        content_item_hash["cache_control"] = { "max_age" => max_age, "public" => content_item.cache_control.include?("public") }
```
bilbof, line 22: Here could you do `"public" => !content_item.cache_control.private?` to use the existing method.

Mutating the content item hash is a _bit_ of a smell but it's not so bad.

app/controllers/finders_controller.rb
```diff
@@ -2,10 +2,10 @@ class FindersController < ApplicationController
   layout "finder_layout"
 
   before_action do
-    expires_in(5.minutes, public: true)
+    content_item = ContentItem.from_content_store(finder_base_path)
```
bilbof, line 5: In the worst case (the cache is unavailable) we're calling the content store twice. In the best case (cache is warm) we're calling the cache twice. I think we can do a bit better and have a memoized method (e.g. def content_store; @content_item ||= ContentItem.from_content_store(finder_base_path); end`) so we don't call it twice.

bilbof (changes requested):
Thanks for your work here. I've requested changes, and I wonder if you could put a commit message in to explain the yarn change.

app/lib/services.rb
kentsanggds: ok, I'm trying to use the existing `private` method

Is there another way to add cache_control without mutating the hash?

app/lib/services.rb
kentsanggds: ok, have changed it, only did this as wasn't quite sure when we want to adhere to the content store ttl

app/controllers/search_controller.rb
kentsanggds: yes, I've updated

app/lib/services.rb
bilbof: I think mutating the hash is the pragmatic choice, no need to change it 👍

app/controllers/finders_controller.rb
```diff
@@ -2,10 +2,10 @@ class FindersController < ApplicationController
   layout "finder_layout"
 
   before_action do
-    expires_in(5.minutes, public: true)
+    @content_item ||= ContentItem.from_content_store(finder_base_path)
+    expires_in(content_item.max_age, public: content_item.cache_public)
```
bilbof, line 6: On line 82 there's a `content_item` method you could use instead, so you could delete line 5.

```ruby
def content_item
    @content_item ||= ContentItem.from_content_store(finder_base_path)
  end
```

app/controllers/finders_controller.rb
```diff
@@ -27,7 +27,7 @@ def show
           redirect_to_destination
         else
           @search_query = initialize_search_query(is_for_feed: true)
-          expires_in(ATOM_FEED_MAX_AGE, public: true)
+          expires_in(content_item.max_age, public: content_item.cache_public)
```
bilbof, line 30: You no longer need this line since there's a `before_action` that does the same thing.

kentsanggds: ok, I'll remove

app/controllers/finders_controller.rb
kentsanggds: 👍

bilbof (approved):

app/controllers/search_controller.rb
```diff
@@ -1,11 +1,11 @@
 class SearchController < ApplicationController
   layout "search_layout"
-  before_action :set_expiry
 
   def index
     search_params = SearchParameters.new(params)
 
     @content_item = Services.cached_content_item("/search")
+    set_expiry(@content_item)
```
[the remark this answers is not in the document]
kentsanggds: I'll try that out

bilbof (approved):

PR #2729, opened 2022-02-10
Remove height on filter menu

chris-gds:
## What 

Remove `height` from filter menu on mobile

## Why

On mobile this height limit restricts the viewport of this section and whats visible. This leads to a potentially difficult experience for mobile users, this fix removes this so the user can scroll naturally

## Visuals

Before: 

https://user-images.githubusercontent.com/71266765/153610573-1f688c44-b606-4a8e-9af7-0e394341085e.mov

After:

https://user-images.githubusercontent.com/71266765/153607934-80d3a07e-7eaa-4512-8dd8-b69bc8a542f0.mov

### Desktop

https://user-images.githubusercontent.com/71266765/153608061-0ab11b80-516c-403a-b249-6e0a907f1ab7.mov

## Anything else

Desktop `max-height` has been removed

### Testing 

https://www.gov.uk/administrative-appeals-tribunal-decisions
https://finder-front-filter-men-i3xnrn.herokuapp.com/

### DIFF
--- a/app/assets/stylesheets/components/_option-select.scss
+++ b/app/assets/stylesheets/components/_option-select.scss
@@ -84,8 +84,6 @@
 
 .app-c-option-select__container {
   position: relative;
-  max-height: 200px;
-  overflow-y: auto;
   overflow-x: hidden;
   background-color: govuk-colour("white");
 
@@ -129,7 +127,9 @@
   }
 
   .app-c-option-select__container {
-    height: 200px;
+    @include govuk-media-query($from: tablet) {
+      height: 200px;
+    }
   }
 
   [data-closed-on-load="true"] .app-c-option-select__container {
--- a/app/assets/javascripts/components/option-select.js
+++ b/app/assets/javascripts/components/option-select.js
@@ -18,6 +18,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   OptionSelect.prototype.init = function () {
+    var SETTINGS = {
+      breakpoint: {
+        desktop: 769
+      }
+    }
+
+    // Returns what screen size the window is currently. Returns a string of
+    // either `desktop` or `mobile`
+    this.windowSize = function () {
+      return document.documentElement.clientWidth >= SETTINGS.breakpoint.desktop ? 'desktop' : 'mobile'
+    }
+
     if (this.hasFilter.length) {
       var filterEl = document.createElement('div')
       filterEl.innerHTML = this.hasFilter
@@ -212,7 +224,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   OptionSelect.prototype.setContainerHeight = function setContainerHeight (height) {
-    this.$optionsContainer.style.height = height + 'px'
+    // Only apply inline height should the user be on desktop
+    if (this.windowSize() === 'desktop') {
+      this.$optionsContainer.style.height = height + 'px'
+    }
   }
 
   OptionSelect.prototype.isCheckboxVisible = function isCheckboxVisible (option) {


app/assets/stylesheets/components/_option-select.scss
```diff
@@ -82,7 +82,6 @@
 
 .app-c-option-select__container {
   position: relative;
-  max-height: 200px;
   overflow-y: auto;
```
maxgds, line 85: I think we can remove the overflows along with this?

maxgds (changes requested):

app/assets/stylesheets/components/_option-select.scss
chris-gds: Thanks @maxgds - this wasn't quite ready for eyes just yet - i'm getting errors rendering `http://static.dev.gov.uk/templates/header_footer_only.html.erb` and wanted to see this on an environment. 

It's good to know the review request comes through once tagged rather than when marked ready.

maxgds: Ah, my mistake, sorry.

maxgds:
Hmm, to be honest I'd anticipated this being a lot simpler than the js you've tweaked here. I anticipated this css
```
.app-c-option-select__container {
  position: relative;
  max-height: 200px;
  overflow-y: auto;
  overflow-x: hidden;
  background-color: govuk-colour("white");

  &:focus {
    outline: 0;
  }
}
```

being augmented with the design system queries:

```
@include govuk-media-query($from: desktop) {
```

Do you think that would be a usable option? (It should probably be from tablet, actually).

However your screenshots have uncovered something I'd not anticipated. Where are the filters you're showing there? I was only aware of a much more simple search filter like this: https://www.gov.uk/search/all?keywords=pension&order=relevance and I'm not sure having such a massive list to traverse as shown in your screenshot works well on mobile. I'm interested to know how you got to those screens and whether you think the functionality could be coded to diverge across your example and mine if that were desirable? We may need more design input on this, I think.

chris-gds:
> Do you think that would be a usable option? (It should probably be from tablet, actually).

It looked like this initially, however what's also happening is theres a dynamic height that's being placed inline on these elements - I would have gone down the `important` route but figured this was a better option.

image

> However your screenshots have uncovered something I'd not anticipated. 

> Where are the filters you're showing there? 

My fault - here a link https://www.gov.uk/administrative-appeals-tribunal-decisions

I was only aware of a much more simple search filter like this: 

> https://www.gov.uk/search/all?keywords=pension&order=relevance and I'm not sure having such a massive list to traverse as shown in your screenshot works well on mobile. 

> I'm interested to know how you got to those screens and whether you think the functionality could be coded to diverge across your example and mine if that were desirable? 

How do you mean?

> We may need more design input on this, I think.

Sure, happy to loop in

maxgds:
What I'm thinking is that making the scroll more natural makes sense for the sort of list like on the standard basic search page I linked, but on others it makes things horrific, the one in your video looks bad to me (but Nik seemed happy with it going by the comment on the trello card), but in this example it's terrible: 

https://finder-front-filter-men-i3xnrn.herokuapp.com/search/transparency-and-freedom-of-information-releases

Go to the organisations and there are loads of them to scroll past. And the majority of them are closed. If you've found the one you want on your mobile and you want to get to the button at the bottom of the pages you're going to spend absolutely ages swiping. I think the existing behaviour is better there. Is there a way we can pick and choose, perhaps based on the number of options that it seems sensible to scroll past? That's what I meant by diverging code, have it open on some pages and in a scrollable box on others. Or should we just stick with what's on live already - it looked like this would be an improvement (and quick to update!) but is it?

andysellick:
Dropping in some quick context on this in case it's useful (although I see Max already noticed that there are pages where there are literally hundreds of options so removing the height limit altogether might not be ideal). Part of this code is very complicated and tries to ensure that the last visible checkbox in the list is always cut in half - so the user is given a visual cue that scrolling down is possible (since scroll bars are often invisible these days). Hopefully this PR shouldn't impact that but wanted to highlight it.

chris-gds:
> What I'm thinking is that making the scroll more natural makes sense for the sort of list like on the standard basic search page I linked, but on others it makes things horrific..

> Go to the organisations and there are loads of them to scroll past. And the majority of them are closed. If you've found the one you want on your mobile and you want to get to the button at the bottom of the pages you're going to spend absolutely ages swiping. I think the existing behaviour is better there. 

Agreed, great flag - this introduces a new problem. 

> Is there a way we can pick and choose, perhaps based on the number of options that it seems sensible to scroll past?

It's definitely an idea, I'll propose this. Thinking about how this might work, perhaps any "long list" the original behaviour is present. (The mix of interactions I'm not sure is ideal)

> Or should we just stick with what's on live already - it looked like this would be an improvement (and quick to update!) but is it?

It's a difficult one. The original solution not only creates a second, hidden scrollbar but also within a small container to scroll an equally long list.  This is just as un-usable as this new issue, arguably this is worse - the user is  is unable to get to the bottom item with more intense scrolling. 

This new solution would 

- allow a user to use “find / search” in a page more easily
- it’s more of a natural scroll
- potentially the user can scroll up and close the accordion
- for all smaller lists it's an improvement

I have a bit of theory that this type of filtering is more intense work and a user may be more likely to interact with the filtering on desktop. This interaction on mobile; to be effective would need a real re-think.  It’s been suggested that the volume of users interacting in this way is quite small, I've not been able to verify this though.

>Part of this code is very complicated and tries to ensure that the last visible checkbox in the list is always cut in half - so the user is given a visual cue that scrolling down is possible (since scroll bars are often invisible these days). 

Thanks for highlighting peak / partial reveal design, this makes sense.

On balance I'm thinking this is still an iterative  improvement - however I'm really close to the fence on this one. I'll continue to chat to see if there's anything else we can do.

chris-gds:
Closing PR as solution is less than ideal and priority call was made, in this context not to continue.
Acknowledging the problem still persists and [capturing findings for future](https://trello.com/c/AtxV10Ww)

---

Learning from Design

> Jeremy Yun advised Learned that the `min-height` was already set when he joined the team. We talked about the issue of two scroll bars being an issue, while acknowledging the long list of filters is not great. Learned that particular [Research & Stats](https://www.gov.uk/search/research-and-statistics) are aimed at web-proficient users. Was advised to reach out to Simon Hughesdon

> Spoke to Simon Hughesdon today. He noted that he wasn't sure about the `min-height` requirements. He did note that when the team existed they were hoping to further iterate on the filters by only showing applicable filters based on returned search results. Ex. if someone searches for "Boris Johnson" why include the filter option "Canada"
As they were about to further iterate the team got disbanded.

> I've also found the team's old [roadmap](https://docs.google.com/spreadsheets/d/1EKVsMW7tWOuj_W0ZgqN4rMEC3VrdJ2qnTeAGB5HeAVg/edit?usp=sharing) and one task that has not been marked as either "done" or "in progress" is "Prototype and accessibility review new search / finders UI".

PR #2734, opened 2022-02-15
Remove cache key checks from readiness probe

### LINKED ISSUE #2732: Remove cache content check from readiness probe endpoint
The [cache readiness](https://github.com/alphagov/finder-frontend/blob/main/config/routes.rb#L8) probe should be removed from the `/healthcheck/readiness` check.

We should export the status of connectivity to memcached and search-api as Prometheus metrics rather than as healthcheck.

Readiness probes should tell us whether a pod is ready to serve requests, and should not check their dependencies.

[GOV RFC 141](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-141-application-healthchecks.md) incorrectly states:

> If an app can't talk to a backing service it relies on, then it probably can't be trusted to handle any requests.

This is untrue since application should be expected to degrade gracefully, and an external dependency may not be required for all request paths. For example, this app can serve the `/search` homepage without the Search API being reachable, and can serve all requests if the memcached instance is down, and will degrade gracefully if Content Store is unavailable.

We should not be causing cascading failures by healthchecking an app's dependencies.

For instance, if memcached is unavailable this app cannot be started in Kubernetes.

bilbof:
Resolves #2732

This will enable Finder Frontend to serve requests before its cache is warm. The readiness probe should only check
the hard dependencies of the application (ie are required to serve all requests). This allows the application to degrade gracefully when external dependencies fails, and prevents a cascading failure when those dependencies fail.

I've left the Cache check class in as I'd like to reuse it to export cache status as Prometheus metrics.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,9 +4,7 @@ Rails.application.routes.draw do
   get "/sign-in", to: proc { [200, {}, %w[OK]] }
 
   get "/healthcheck/live", to: proc { [200, {}, %w[OK]] }
-  get "/healthcheck/ready", to: GovukHealthcheck.rack_response(
-    Healthchecks::RegistriesCache,
-  )
+  get "/healthcheck/ready", to: proc { [200, {}, [JSON.generate({ status: :ok })]] }
 
   root to: redirect("/development") unless Rails.env.test?
   get "/development" => "development#index"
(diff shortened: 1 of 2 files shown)


alexbakervr (approved):
LGTM

PR #2738, opened 2022-03-01
Use Internal Worldwide API

bilbof:
FinderFrontend will use the internal /api/world-locations
API rather than going out of the VPC and then back in.

This removes an oddity - and also removes the need for
finder-frontend to provide basic authentication when accessing
the API at www-origin.


fredericfran-gds (approved):
looks good

PR #2758, opened 2022-03-24
Remove jQuery duplication

andysellick:

## What / why

- a test was failing due to a component not being initialised properly
- to [fix this](https://github.com/alphagov/finder-frontend/pull/864), the dependencies.js file was included from the components gem directly into finder-frontend, which initialises our components
- unfortunately, dependencies.js also included jQuery, which was already included via the same file in static for all of GOV.UK
- this meant that finder-frontend has been receiving two copies of jQuery for some time
- fix this problem by moving the dependencies.js call into a separate file and only including that if the environment is test

## Visual changes
None.


spec/javascripts/live_search_spec.js
```diff
@@ -1,5 +1,8 @@
 describe('liveSearch', function () {
   var $form, $results, _supportHistory, liveSearch, $atomAutodiscoveryLink, $count
+  var GOVUK = window.GOVUK || {}
+  GOVUK.modules = GOVUK.modules || {}
+  GOVUK.modules.start = GOVUK.modules.start || function () {}
```
kevindew, line 5: I expect you can remove the need for this by adding test-dependencies.js to https://github.com/alphagov/finder-frontend/blob/d3b4771780258ea9250c44439f93428d64411ce1/spec/javascripts/support/jasmine.yml#L16

kevindew (commented):
Looks great - just a suggestion on the final commit fix.

spec/javascripts/live_search_spec.js
andysellick: @kevindew thanks for the suggestion, works perfectly, all updated.

kevindew (approved):

PR #2786, opened 2022-05-06
Topic finder pages use the full width layout 

patrickpatrickpatrick:
## What

Add a line of code which replaces the default value of `slimmer_template` (`gem_layout`) to `gem_layout_full_width` only
 if the page is a topic finder page.

## Why

On topic finder pages, a different styling is used for the page header. The header is wrapper in a banner which gives the text a blue background (so that it is differentiated from standard finder pages). The background is intended to be as wide as the page but it is currently being constrained by the layout. This was not previously seen before as the layout when this header style change was made for topic pages because the layout being used at that time for all finder pages was  `gem_layout_full_width` (because an emergency banner was being deployed on the pages). After this banner was removed, the gem layout was restored to the 'regular' width so that the blue bar below the navigation bar (which is present on gem_layout but not present on `gem_layout_full_width`) was restored on all finder pages. This change means that the correct layout will be used on the correct type of finder page.

### Before

Screenshot 2022-05-06 at 12 44 08

Screenshot 2022-05-06 at 12 41 01

### After
 
Screenshot 2022-05-06 at 12 40 21

Screenshot 2022-05-06 at 12 41 21

[Relevant Trello Card](https://trello.com/c/ehmV9NDk/975-bug-finder-pages-are-missing-padding-on-blue-header), [Jira issue NAV-5358](https://gov-uk.atlassian.net/browse/NAV-5358)

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -6,6 +6,8 @@ class FindersController < ApplicationController
   end
 
   def show
+    slimmer_template "gem_layout_full_width" if i_am_a_topic_page_finder
+
     respond_to do |format|
       format.html do
         raise UnsupportedContentItem unless content_item.is_finder?


chao-xian (approved):
LGTM but as it's quite a visual change can we check that Designers are ok too?

PR #2800, opened 2022-05-31
Create `/search/latest` endpoint

brucebolt:
, [Jira issue PP-792](https://gov-uk.atlassian.net/browse/PP-792)We are retiring `/government/latest` from being rendered in Whitehall.

Finder Frontend's `/search/all` finder provides the same search response, so we are redirecting `/government/latest` to this finder.

However there is a mismatch in the query string parameters used between the two applications. Therefore we need to redirect any queries containing `departments` to match the expectation of the finder (`organisations`).

Additionally, we need to add a default search order to see the latest documents instead of the most relevant documents.

[Trello card](https://trello.com/c/93adHMBQ)

### DIFF
--- a/app/controllers/redirection_controller.rb
+++ b/app/controllers/redirection_controller.rb
@@ -7,6 +7,16 @@ class RedirectionController < ApplicationController
     redirect_to(finder_path(params[:slug], params: covid_topic_and_other_params))
   end
 
+  def redirect_latest
+    redirect_params = {
+      order: "updated-newest",
+    }
+    redirect_params[:organisations] = params[:departments] if params[:departments]
+    redirect_params[:topical_events] = params[:topical_events] if params[:topical_events]
+    redirect_params[:world_locations] = params[:world_locations] if params[:world_locations]
+    redirect_to(finder_path("search/all", params: redirect_params))
+  end
+
   def advanced_search
     conversion_hash =
       {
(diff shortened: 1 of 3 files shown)


app/controllers/redirection_controller.rb
```diff
@@ -7,6 +7,16 @@ def redirect_covid
     redirect_to(finder_path(params[:slug], params: covid_topic_and_other_params))
   end
 
+  def redirect_latest
+    redirect_params = {
+      order: "updated-newest",
+    }
+    redirect_params[:organisations] = params[:departments] if params[:departments]
+    redirect_params[:topical_events] = params[:topical_events] if params[:topical_events]
+    redirect_params[:world_locations] = params[:world_locations] if params[:world_locations]
```
BeckaL, line 16: Just a thought, I'm wondering here if there's a way of only adjusting the departments param?

Was playing around with something like 

`params.transform_keys { |key| key == :departments ? :organisations : key }`

but getting non permitted param errors, so maybe need to filter for only permitted ones.

Basically, the motivation being that we aren't actually doing anything to the topical events or world location params, so feels noisy to being doing this. Also do we have a guarantee that these are the only params we'll want to pass through?

BeckaL (approved):
Looks good to me! Just wondering if there's a cleaner way of creating the redirection params (there might not be) as per my comment, but non-blocking.

PR #2803, opened 2022-05-31
Guard against invalid page param

KludgeKML:

Add tests for page being set to invalid values, as here: https://sentry.io/organizations/govuk/issues/3314230696/?project=202224

(page=[2], etc).

### DIFF
--- a/app/lib/search/query_builder.rb
+++ b/app/lib/search/query_builder.rb
@@ -57,6 +57,8 @@ module Search
 
     def current_page
       [params["page"].to_i, 1].max
+    rescue StandardError
+      1
     end
 
     def documents_per_page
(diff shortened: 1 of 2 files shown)


app/lib/search/query_builder.rb
```diff
@@ -56,7 +56,10 @@ def pagination_start
     end
 
     def current_page
-      [params["page"].to_i, 1].max
+      page = Integer(params["page"] || 1)
+      [page, 1].max
+    rescue StandardError => _e
```
1pretz1, line 61: Think we can remove assigning the exception to a variable (`=> _e`) as it's not in use. It is underscored which implies it's not being used, but we could go one further and remove that part entirely.

1pretz1 (approved):
Thanks for sorting this Keith - I've left two comments but approved as they are non blocking / minor tweaks.

app/lib/search/query_builder.rb
KludgeKML: Ah, I'm so used to old rubocops requiring an assignment I didn't even think to test whether govuk's cop required it. Have removed the assignment. Thanks!

PR #2815, opened 2022-06-29
Add XSS tests to allow removal from Smokey

kevindew:

This adds a capybara feature test and a controller test to verify that, if a user searches with malicious input, we don't execute or return their JavaScript.

It has been done to allow the removal of https://github.com/alphagov/smokey/blob/7553a432259d785fb6d4fb81cfd4fe3572185ad3/features/apps/finder_frontend.feature#L11-L31 and https://github.com/alphagov/smokey/blob/7553a432259d785fb6d4fb81cfd4fe3572185ad3/features/apps/finder_frontend.feature#L61-L77

More info in commits


features/finders.feature
```diff
@@ -24,12 +24,19 @@ Feature: Filtering documents
...
     Given a collection of documents exist
-    When I search documents by keyword
+    When I search documents by keyword: "keyword searchable"
     Then I see all documents which contain the keywords
     And there is not a zero results message
     And the page title is updated
     And I can see that Google won't index the page
 
+  @javascript
+  Scenario: User tries to filter with malicious input
```
benthorner, line 34: Did you find a way to test this locally?

As in, make a change that means a raw `<script>` tag does appear on the page and see the final step fails? I had a go myself - just putting an `alert()` in the ERB file, except something seems to be scrubbing it. I can get it to work for the initial page load - requires a corresponding `accept_confirm` in the `visit` method call - but not for the actual search.

Looking at the page content, a significant amount of the structure is added or changed by JS, but I couldn't see anywhere that would be filtering out this tag.

Do you have any ideas?

benthorner (commented):
Really pleased to see these tests getting migrated 😃.

I agree the first two commits replace [one of the Smokey tests](https://github.com/alphagov/smokey/blob/7553a432259d785fb6d4fb81cfd4fe3572185ad3/features/apps/finder_frontend.feature#L11-L31). I'm not sure about the other one, though: site search is distinct from finders and has its own controller. Not a blocker for this PR, just for which tests we may then go and delete.

I also found a couple of issues that I think are worth looking at before this goes in. Also, while I don't think it's a blocker, ideally I'd like to show the feature test works by creating the condition for it to fail - I've left a comment about what I tried.

features/step_definitions/filtering_steps.rb
```diff
@@ -60,6 +60,16 @@
   expect(page).to have_css(".finder-results .gem-c-document-list__item", count: 1)
 end
 
+Then(/I should see the keyword$/) do
+  expect(page).to have_content(@keyword_search)
```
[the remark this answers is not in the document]
kevindew: Ah yes, sorry that was left over from an earlier approach. Thanks for catching. 

Seems low value to test the search box in this respect since we've just entered text into it.

features/finders.feature
```diff
@@ -24,12 +24,19 @@ Feature: Filtering documents
...
+    Given a collection of documents exist
+    When I search documents by keyword: "<script>alert(0)</script>"
+    Then the page title is updated
+    And there should not be an alert
```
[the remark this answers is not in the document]
kevindew: Argh yes, it took me quite a lot of effort to replicate that. This repo is set up pretty weird by actually having a cache in the test environment: https://github.com/alphagov/finder-frontend/blob/18d8347b9744ec7661023a2eaef4eda5cad97e2c/config/environments/test.rb#L13 and looking through the tests they're littered with `Rails.cache.clear` - sadly changing cache to the regular `null_store` creates a fresh batch of test failures.

Anyway.... 

It was quite interesting how this is flaky. It seems to be that the request occurs after the test is complete. So webmock has been disabled. I was able to resolve this by adding an extra step to the test that ensured it waited for the page to load. It's a bit contrived.

features/finders.feature
kevindew: I was able to replicate it with a rather contrived bit of code. You can add `document.write(this.currentKeywords())`.

With the way this functionality has changed there is no longer the embedding of the search term via ruby in the live search (it's all done through JS dom methods currently) so it'd be quite hard to have an XSS issue. The controller test is the only one where ERB embedding is in play.

I could well be persuaded to remove this test, what do you think? I see it's inclusion somewhat justified by it being a regression test, however it is essentially a regression test for behaviour that is now removed.

kevindew:
> Really pleased to see these tests getting migrated 😃.
> 
> I agree the first two commits replace [one of the Smokey tests](https://github.com/alphagov/smokey/blob/7553a432259d785fb6d4fb81cfd4fe3572185ad3/features/apps/finder_frontend.feature#L11-L31). I'm not sure about the other one, though: site search is distinct from finders and has its own controller. Not a blocker for this PR, just for which tests we may then go and delete.

Site search is no longer distinct from finders. The controller it has is just an index page and that submits a form that redirects you to the finders controller (search/all) so testing that for XSS is all that is needed.

features/finders.feature
benthorner: > I was able to replicate it with a rather contrived bit of code.

I'm not sure what you mean here. I was able to get an alert dialogue but only on the initial page load, not after the form submission. The removal of `<script>` tags - but not other tags - remains mysterious.

However, I'm not sure it matters as what you say makes sense: injection isn't a problem if it happens via JS, and if JS isn't enabled then any injection has no consequence.

While it's annoying there's no way to make the test fail today, I'm happy to think of it as an omission from before live search existed - it would likely still be here if it had been added originally.

And like you say, there's historical precedent for being overcautious here.

benthorner (approved):
Good to go 🚢.

> Site search is no longer distinct from finders.

So it is - thanks for explaining. This does indeed replace all the smoke tests you identified 🎉.

PR #2837, opened 2022-07-28
Fix search ecommerce and result click tracking

andysellick:

## What
Apply fixes so that tracking works after search parameters are changed and parts of the page have been dynamically updated. See individual commits for details.

## Why
Tracking was working fine on initial page load but stopped working after a second search.

## Visual changes
None.

Trello card: https://trello.com/c/a9GavpZp/87-fix-missing-ua-search-finder-event-clicks-after-applying-filters-or-changing-the-search-term

### DIFF
--- a/features/step_definitions/analytics_steps.rb
+++ b/features/step_definitions/analytics_steps.rb
@@ -1,7 +1,7 @@
 Then(/^the links on the page have tracking attributes$/) do
   visit finder_path("government/policies/benefits-reform")
 
-  expect(page).to have_selector('.finder-results[data-module="gem-track-click"]')
+  expect(page).to have_selector('.js-live-search-results-block[data-module="gem-track-click"]')
 
   document_links = page.all(".gem-c-document-list__item-title")
   expect(document_links.count).to be_positive
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -366,9 +366,9 @@
         if (xhr.readyState === 4 && xhr.status === 200) {
           var response = JSON.parse(e.target.response)
           liveSearch.updateUrl()
-          liveSearch.trackSearch()
           liveSearch.cache(liveSearch.serializeState(liveSearch.state), response)
           liveSearch.displayResults(response, searchState)
+          liveSearch.trackSearch()
         } else {
           liveSearch.showErrorIndicator()
         }
@@ -380,8 +380,8 @@
       xhr.send()
     } else {
       this.updateUrl()
-      this.trackSearch()
       this.displayResults(cachedResultData, searchState)
+      this.trackSearch()
     }
   }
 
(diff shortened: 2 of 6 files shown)


Rosa-Fox (approved):
Hi Andy,

To me it looks good. From what I can tell, [this](https://github.com/alphagov/finder-frontend/blob/b89fe2820ac425372b0e30a0b0a90fd4e1a7f9ba/app/assets/javascripts/live_search.js#L386) starts the tracking for results in the `.js-live-search-results-block`  when the search is updated. [Previously](https://github.com/alphagov/finder-frontend/blob/b89fe2820ac425372b0e30a0b0a90fd4e1a7f9ba/app/assets/javascripts/live_search.js#L200) it was only set for the original search. 

I ran it locally on `main` and could see there were no tracking events when links were clicked after the filters/reordering etc had been applied. I ran this branch and tried out various combinations of sorting a filtering and the events were firing as expected.

tracking-filters-reordering

I clicked around on the other links in the results section (pagination, subscribe etc) and they all still worked fine. It seems the `track-category` and other stuff defined in the Trello card was already set [here](https://github.com/alphagov/finder-frontend/blob/919a38385d5ce526bb4b33449e6cf591ce55db38/app/presenters/search_result_presenter.rb#L102).

I'll approve it but you might want a frontender to also have a quick look over.

features/step_definitions/analytics_steps.rb
```diff
@@ -1,7 +1,7 @@
 Then(/^the links on the page have tracking attributes$/) do
   visit finder_path("government/policies/benefits-reform")
 
-  expect(page).to have_selector('.finder-results[data-module="gem-track-click"]')
+  expect(page).to have_selector('.js-live-search-results-block[data-module="gem-track-click"]')
```
jon-kirwan, line 5: Does the query selector here need to be so specific? Perhaps it could be changed to `js-live-search-results-block`. i think there is only one occurrence of this selector.

jon-kirwan (commented):
I think this looks fine to me. I've tested and can see both events are fired when the filter options are updated and a link then clicked from the updated search results page.
```
#4 event⤑ ['navFinderLinkClicked',...
#3 event⤑ ['UX','click', 'Results', '', 0]
```
I've added a very tiny suggestion below and also wondered if `data-module="gem-track-click"` can be removed from `_filter_button.html.erb` since it's nested within the same `js-live-search-results-block` container?

features/step_definitions/analytics_steps.rb
andysellick: I think the intent of this test is to check that this element both exists and includes the `gem-track-click` module, so I think the specific selector is intentional.

jon-kirwan: Ah yes, my mistake - I see it's mentioned in the test description 👍

jon-kirwan (approved):
Thanks for the change. All looks good to me 🚀

andysellick:
Going to hold off merging this until it can be checked again by the performance analyst on our team, who's currently on leave. Thanks for the reviews all 👍

Issue #2841, opened 2022-08-02
Is hidden search 'Skip to results' link in the right place?


edwardhorsford:
The GOV.UK search has a hidden link `Skip to results` shown on focus. I'd like to query if it's in the right location on the page / working as expected.

I *think* it used to be placed at the start of the filters. Now it appears to be right next to the results:
<img width="1075" alt="Screenshot 2022-08-02 at 17 34 15" src="https://user-images.githubusercontent.com/2204224/182426912-b99fcda4-6a12-4f2e-8154-a38549c681fe.png">

If the purpose is to save keyboard only and screen reader users the pain of tabbing through a bunch of things - I don't think we're helping much as it's so close to the results. I think it used to be before the filters themselves where it saved a bunch of tabbing.

andysellick:
Interesting, thanks @edwardhorsford. Yes, I think it used to be at the start of the filters too. I'll have a look.

edwardhorsford:
Thanks @andysellick!

PR #2667, opened 2021-10-27
Update how users see a change in return results (search filter)

jon-kirwan:
## What

- Display *No. of results* label on mobile (remove media query) and displayed above the 'filter' button
- On filter overlay (mobile) display *No. of results* label below the 'Filter' `h1`
- On filter overlay (mobile) remove the number of results text from the submit button so the button text now just reads 'Show search results'

## Why

- Small improvement to GOV.UK Search
- Especially after making the Search button on the filter screen (in mobile) no longer sticky.
- Also, addresses an issue, though not a WCAG fail, with information changing on the page automatically after a user interaction.
- More info about this can be found in the following ticket: https://trello.com/c/1c8oQKhJ

## Visuals

| Before (all_content_finder)  | After (all_content_finder) |
|---|---|
|  |  |
| | |
|  |  |

| Before (!all_content_finder)  | After (!all_content_finder) |
|---|---|
|  |  |
| | |
|  |  |

## How to test

- https://www.gov.uk/search/all (all_content_finder)
- https://www.gov.uk/search/services (!all_content_finder)

## Anything else
- In the 'filter' overlay view I have marked up the no. of results (e.g. 520,927 results or 7,615 services), which is displayed below the main 'Filter' title (`h1`) with a `h2` since this is how no. of results is formatted on larger screens.
- I've also moved the skip link component to sit below the newly repositioned 'filter' button (so the tab order is the same as it currently is).

PR #2842, opened 2022-08-03
Move skip link on search page

andysellick:

## What
Moving the skip link on the search pages to be before the search filters instead of after them. This is intended to allow keyboard users to skip to search results without having to tab through the filters.

Have left it after the main search box rather than before, on the assumption that users will want to access that even if they don't want to access the filters.

## Why
Was moved unintentionally as part of this change https://github.com/alphagov/finder-frontend/pull/2667

Fixes https://github.com/alphagov/finder-frontend/issues/2841

## Visual changes

Before | After
------ | ------
Screenshot 2022-08-03 at 09 54 14 | Screenshot 2022-08-03 at 09 53 57

### DIFF
--- a/app/views/finders/_facet_collection.html.erb
+++ b/app/views/finders/_facet_collection.html.erb
@@ -17,6 +17,14 @@
   <% end %>
 
   <% if facets.select(&:filterable?).any? %>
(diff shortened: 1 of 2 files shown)


davidtrussler (approved):
LGTM 👍

PR #2852, opened 2022-08-20
Clean up ENV values from Dockerfile

nsabri1:
https://trello.com/c/KgpXafSS/983-update-app-dockerfiles-to-remove-env-values-and-update-e2e-test

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,9 +20,7 @@ RUN bundle exec rails assets:precompile && rm -fr /app/log
 
 FROM $base_image
 
-ENV GOVUK_APP_NAME=finder-frontend PORT=3062
-
-HEALTHCHECK CMD curl --silent --fail localhost:$PORT/healthcheck/ready || exit 1
+ENV GOVUK_APP_NAME=finder-frontend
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,9 +20,7 @@ RUN bundle exec rails assets:precompile && rm -fr /app/log
 
 FROM $base_image
 
-ENV GOVUK_APP_NAME=finder-frontend PORT=3062
-
-HEALTHCHECK CMD curl --silent --fail localhost:$PORT/healthcheck/ready || exit 1
```
sengi, line 25: Do you need to [move this to docker-compose](https://docs.docker.com/compose/compose-file/#healthcheck)? Or is it definitely not relevant for publishing-e2e-tests? (Does anything in the test fixtures rely on waiting for finder-frontend to become ready before sending request to it?)

sengi (approved):

Dockerfile
nsabri1: Oh yes i forgot to add that - i will add that now good catch.

I suspect not likely needed - but i i will add it in the docker compose

sengi (approved):

PR #2880, opened 2022-09-14
Add enhanced ecommerce tracking

JamesCGDS:

## What
This PR adds:

- `ga4-enhanced-ecommerce-tracker` to the `data-module` attribute on the DOM element wrapping the search results so that our `Ga4EnhancedEcommerceTracker` tracking module can be used
- The necessary JS to start tracking every time the search results are updated via filtering or sorting

## Why
This change has been made as part of our migration to GA4.

## Visual changes

N/A

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -52,6 +52,8 @@
       window.ga('set', 'transport', 'beacon')
     }
 
+    this.Ga4EcommerceTracking(true)
+
     this.focusErrorMessagesOnLoad(this.$form)
 
     // prevent page refresh on search submit button click
@@ -115,6 +117,22 @@
       this.$suggestionsBlock.setAttribute('data-search-query', this.currentKeywords())
     }
     if (GOVUK.Ecommerce) { GOVUK.Ecommerce.start() }
+
+    this.Ga4EcommerceTracking(false)
+  }
+
+  LiveSearch.prototype.Ga4EcommerceTracking = function (isNewPage) {
+    if (GOVUK.analyticsGA4 && GOVUK.analyticsGA4.Ga4EnhancedEcommerceTracker) {
+      var consentCookie = GOVUK.getConsentCookie()
+
+      if (consentCookie && consentCookie.settings) {
+        GOVUK.analyticsGA4.Ga4EnhancedEcommerceTracker.init(isNewPage)
+      } else {
+        window.addEventListener('cookie-consent', function () {
+          GOVUK.analyticsGA4.Ga4EnhancedEcommerceTracker.init(isNewPage)
+        })
+      }
+    }
   }
 
   LiveSearch.prototype.getAndUpdateTaxonomyFacet = function getAndUpdateTaxonomyFacet () {
--- a/app/views/finders/show.html.erb
+++ b/app/views/finders/show.html.erb
@@ -44,6 +44,7 @@
         role="region"
         aria-label="<%= content_item.title %> search results"
         data-analytics-ecommerce
+        data-ga4-ecommerce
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -115,6 +115,14 @@
       this.$suggestionsBlock.setAttribute('data-search-query', this.currentKeywords())
     }
     if (GOVUK.Ecommerce) { GOVUK.Ecommerce.start() }
+
+    if (GOVUK.Modules) {
```
andysellick, line 119: If the intent of this check is to make sure that the GA4 ecommerce tracker exists before running, would it be better to change it to `if (GOVUK.Modules.Ga4EnhancedEcommerceTracker) {`?

app/assets/javascripts/live_search.js
```diff
@@ -115,6 +115,14 @@
...
+      var searchResults = document.querySelectorAll("[data-module='gem-track-click ga4-enhanced-ecommerce-tracker']")
```
andysellick, line 120: Is it possible to only search here for `document.querySelectorAll("[data-module='ga4-enhanced-ecommerce-tracker']")`? The reason I ask is that `gem-track-click` is from the existing analytics code, and will disappear or change eventually, at which point this will stop working.

app/assets/javascripts/live_search.js
```diff
@@ -115,6 +115,14 @@
...
+
+      for (var i = 0; i < searchResults.length; i++) {
+        new GOVUK.Modules.Ga4EnhancedEcommerceTracker(searchResults[i]).init(false)
```
andysellick, line 123: I haven't looked at the accompanying PR for this yet but I notice that this has been written as a GOVUK Module, which might not be the right approach. Modules are automatically initialised on a page wherever the Modules start script finds a `data-module` attribute, so in this situation it would have already been initialised by the time this code is run.

However I think we might need to re-initialise this ecommerce code after search updates, which is why I suspect this line is here? In which case it might make more sense to not have the new ecommerce code as a Module, but structure it differently somehow. Open to having a conversation about this.

app/assets/javascripts/live_search.js
JamesCGDS: Good point! I've made this change in the new implementation.

app/assets/javascripts/live_search.js
JamesCGDS: Just to document/reflect the discussion we had yesterday - the ecommerce tracking code will no longer be structured as a module so we won't be using this approach any more.

app/assets/javascripts/live_search.js
JamesCGDS: Just to document/reflect the discussion we had yesterday - the ecommerce tracking code will no longer be structured as a module so we won't be using this approach any more.

JamesCGDS:
Thanks for your feedback @andysellick, I’ve made some changes to this approach now. I’ve left it as a separate commit as it might be easier to see what’s changed and then I can squash the commits together if all’s good.

andysellick (approved):
👍 (although you have some failing tests)

JamesCGDS:
Thanks @andysellick for your help on fixing the failing tests. Just did a quick bit of local manual testing and everything's working fine so I've squashed the commits together 👍

andysellick (approved):
Looks good. We should probably hold off merging this until the ecommerce stuff in the gem is deployed, although I think these changes will be fine to deploy without them - it'll just not do anything, right?

JamesCGDS:
Yeah - as long as `Ga4EnhancedEcommerceTracker` doesn't exist, this code won't do anything.

PR #2906, opened 2022-10-21
Fix referrer bug

JamesCGDS:

## What
This PR fixes a GA4 analytics bug relating to `page_views`.

When a user is on a search results page and applies a new filter or sort option, the URL of the page changes and this needs to be reflected in the `page_view.referrer` property that is pushed to the `dataLayer`. Currently, the  `page_view.referrer` property is not updating when a dynamic update has been triggered (e.g. by applying a filter). To address this, the `location` of the page is stored in a `referrer` variable before any dynamic updates have been triggered and it is ultimately passed to `GOVUK.analyticsGa4.Ga4EcommerceTracker.init()` as part of an object (namely `dynamicPageViewData.referrer`).

In addition to the `dynamicPageViewData.referrer` property, a `dynamicPageViewData.dynamic` property is also passed in order to enable PA's to differentiate between a fresh page load and a dynamic page update.

The `referrer` and `dynamic` values have been passed as properties on an object because this removes a reliance on the parameters being passed in the correct order and, if additional properties need to be passed in future, it can be easily extended.

## Why
Without these fixes, PA's will be unable to:

- Track user journeys accurately when filters or sorting options have been added, changed, or removed
- Differentiate between fresh page loads (e.g. by clicking a link or using the pagination) and dynamic page updates

## Visual changes
None.

## N.B.
Related PR - https://github.com/alphagov/govuk_publishing_components/pull/3032

As this PR changes how parameters are passed to `GOVUK.analyticsGa4.Ga4EcommerceTracker.init()`, these changes are breaking.

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -52,7 +52,7 @@
       window.ga('set', 'transport', 'beacon')
     }
 
-    this.Ga4EcommerceTracking(true)
+    this.Ga4EcommerceTracking()
 
     this.focusErrorMessagesOnLoad(this.$form)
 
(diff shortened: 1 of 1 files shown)


AshGDS (commented):
Looks good - I wonder if an alternative to passing through `referrer` multiple times, is to set the variable you need on the object scope. For example you would do `this.previousSearchURL = window.location.href`, and then you could do: 

```
    this.Ga4EcommerceTracking({
      dynamic: 'true',
      referrer: this.previousSearchURL
    })
```
What do you think? The code might be a bit more self-commenting naming it `previousSearchURL` too as it's easier to see that the referrer is the previous search URL.

andysellick (approved):
I think this makes sense. Is this a breaking change without the corresponding change in `govuk_publishing_components`? If it is, we'd normally make it more obvious on the PR - put `[DO NOT MERGE]` at the start of the title, and put the warning in bold as the first thing in the PR description, so it doesn't get overlooked.

JamesCGDS:
Thanks @andysellick and @AshGDS for your comments. @AshGDS - I've implemented your suggestion and, agreed, it does make things a bit neater. I'll hold off on merging this for the moment as it is dependent on changes being made in the components gem.

JamesCGDS:
@andysellick @AshGDS - following on from the discussion in [this PR](https://github.com/alphagov/govuk_publishing_components/pull/3032), I've made a small change to how `referrer` is being passed. In essence, it's no longer a property in an object and I'm just passing the string.

AshGDS (approved):
Looks good - if possible, I'd squash the commits before merging

andysellick (approved):
Looks good with the changes 🚀

PR #2944, opened 2022-12-13
Implement redirects from Whitehall finder paths

brucebolt:
These redirects used to be rendered by Whitehall, but were removed in https://github.com/alphagov/whitehall/pull/6950.

As Whitehall Frontend is being retired, we need to bring these into another application.  The previous routes are already being redirected to Finder Frontend (`/search/all` for publications and `/search/news-and-communications` for announcements), so are best processed here.

This replicates the behaviour in an identical way to Whitehall, the commit messages link to the redirect code before it was removed.

[Trello card](https://trello.com/c/Drb1H01H/383-bug-fix-redirects-from-whitehall-rendered-finders), [Jira issue PP-892](https://gov-uk.atlassian.net/browse/PP-892)

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -12,6 +12,8 @@ class FindersController < ApplicationController
       format.html do
         raise UnsupportedContentItem unless content_item.is_finder?
 
+        transform_legacy_publication_params_and_redirect if legacy_params_present? && content_item.base_path == "/search/all"
+
         show_page_variables
       end
       format.json do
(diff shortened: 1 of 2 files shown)


app/controllers/finders_controller.rb
```diff
@@ -196,4 +198,143 @@ def include_ecommerce?
       /government/organisations/hm-revenue-customs/contact
     ].exclude?(content_item.base_path)
   end
+
+  def identify_and_redirect_legacy_params
```
BeckaL, line 202: Can we just in-line this method? Seems readable enough to have this in the body

app/controllers/finders_controller.rb
```diff
@@ -196,4 +198,143 @@ def include_ecommerce?
...
+      subtaxons
+      taxons
+      to_date
+    ].freeze
+    given_params = params.keys.map(&:to_sym)
+
+    (legacy_params & given_params).any?
+  end
+
+  def redirect_legacy_params
```
BeckaL: Naming suggestion - `redirect_transforming_legacy_params`? Maybe just me, but when I read this method I assume it's returning me some params for a redirect, not actually doing the redirect itself.

BeckaL: Also, do you think there's a sensible way of splitting this method up into slightly smaller ones? Find it slightly hard to parse this. Edit: Ah, another thing I should have waited to read the next commit for! Maybe you could move the refactor splitting out the taxons stuff into this commit?

app/controllers/finders_controller.rb
BeckaL: Oh I see this method is added to in the next commit, ignore this.

BeckaL (approved):
Nice ✨ ! Looking good to me. Just a few very minor comments, approving anyway as non-blocking refactoring suggestions.

app/controllers/finders_controller.rb
brucebolt: Thanks for reviewing @BeckaL.  I've moved all of the refactoring into the first commit, so the diff is clearer between the two.

KludgeKML (approved):
LGTM

PR #2974, opened 2023-01-14
Remove disused account_api stub.

sengi:
Finder Frontend no longer makes any calls to Account API. The last usage was removed in fea96b0.

This means we'll have one fewer API user bearer token to manage (and they're currently manually rotated, so eliminating one of them is a win).


gclssvglx (approved):
Great stuff - thanks @sengi 👍

PR #2979, opened 2023-01-17
Change mobile search filters modal to a regular 'in page' component

AshGDS:
## What

- Changes the Filters on mobile from being a modal/popup, to a regular "In Page" element thats visibility is triggered by a `<button>` styled as a link.
- Fixes a bug where the amount of filters you selected would not disappear after clicking "Clear all filters". This was because `<span>` was not included in elements to remove in the JS code for this.
- Fixes a visual bug where on Desktop/Tablet, the Topic and Subtopics filters had a `min-width` which would make them spill over onto the search results.

## Why
- The filters change was needed as part of our accessibility fixes.

## Testing

To test, you don't need any local static. You can just use `../startup.sh --live`. Then navigate to http://127.0.0.1:3062/search/all and reduce your screen size until it's a mobile view.

Accessibility wise, I have tested the keyboard functionality, MacOS VoiceOver, Android Talkback, JAWS, NVDA, Lighthouse, WebAim WAVE, navigating at 400% zoom, the behaviour when JavaScript is disabled, and checked the heading structure with headingsMap.

### Before - Mobile Filters
https://user-images.githubusercontent.com/8880610/219028588-90215c3a-a271-44fe-82dd-3827611d7d92.mov

### After - Mobile Filters

https://user-images.githubusercontent.com/8880610/219028825-f43558c7-69c5-4f81-8e06-8e8444823432.mov

## Before - Visual Bug
image

## After - Visual Bug

image

This is a draft PR for an accessibility fix, the code will change quite a bit and we cleaned up before it is ready for a PR.

### DIFF
--- a/app/assets/javascripts/modules/mobile-filters-modal.js
+++ b/app/assets/javascripts/modules/mobile-filters-modal.js
@@ -2,104 +2,46 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  function MobileFiltersModal ($module) {
-    this.$module = $module
-    this.$facetsBox = this.$module.querySelector('.facets__box')
-    this.$closeTriggers = this.$module.querySelectorAll('.js-close-filters')
-    this.$showResultsButton = this.$module.querySelector('.js-show-results')
-    this.$clearFiltersTrigger = this.$module.querySelector('.js-clear-selected-filters')
-    this.$body = document.querySelector('body')
-
-    this.$module.open = this.handleOpen.bind(this)
-    this.$module.close = this.handleClose.bind(this)
-    this.$module.clearFilters = this.handleClearFilters.bind(this)
-    this.$module.ModalFocus = this.handleModalFocus.bind(this)
-    this.$module.boundKeyDown = this.handleKeyDown.bind(this)
+  function MobileFiltersModal (module) {
+    this.module = module
+    this.triggerElement = document.querySelector('[data-toggle="mobile-filters-modal"][data-target="' + this.module.id + '"]')
+    this.clearFiltersTrigger = this.module.querySelector('.js-clear-selected-filters')
+    this.module.toggle = this.handleToggle.bind(this)
+    this.module.open = this.handleOpen.bind(this)
+    this.module.close = this.handleClose.bind(this)
+    this.module.clearFilters = this.handleClearFilters.bind(this)
   }
 
   MobileFiltersModal.prototype.init = function () {
-    var $triggerElement = document.querySelector(
-      '[data-toggle="mobile-filters-modal"][data-target="' + this.$module.id + '"]'
-    )
-    if ($triggerElement) {
-      $triggerElement.addEventListener('click', this.$module.open)
+    if (this.triggerElement) {
+      this.triggerElement.addEventListener('click', this.module.toggle)
+      this.triggerElement.setAttribute('aria-controls', this.module.id)
+      this.triggerElement.setAttribute('aria-expanded', 'false')
     }
 
-    if (this.$closeTriggers) {
-      for (var t = 0; t < this.$closeTriggers.length; t++) {
-        this.$closeTriggers[t].addEventListener('click', this.$module.close)
-      }
-    }
-
-    if (this.$clearFiltersTrigger) {
-      this.$clearFiltersTrigger.addEventListener('click', this.$module.clearFilters)
+    if (this.clearFiltersTrigger) {
+      this.clearFiltersTrigger.addEventListener('click', this.module.clearFilters)
     }
   }
 
-  MobileFiltersModal.prototype.handleOpen = function (event) {
-    if (event) {
-      event.preventDefault()
-    }
-    // when our support browser matrix is above iOS 13
-    // we can use overflow: hidden on the <body> and remove position: fixed
-    this.$body.style.top = '-' + window.scrollY + 'px'
-    this.$body.style.position = 'fixed'
-    this.$focusedElementBeforeOpen = document.activeElement
-    this.$module.classList.add('facets--visible')
-    this.$facetsBox.setAttribute('aria-modal', true)
-    this.$facetsBox.setAttribute('tabindex', 0)
-    this.$facetsBox.focus()
+  MobileFiltersModal.prototype.handleToggle = function (event) {
+    event.preventDefault()
 
-    document.addEventListener('keydown', this.$module.boundKeyDown, true)
+    if (this.module.classList.contains('facets--visible')) {
+      this.module.close()
+    } else {
+      this.module.open()
+    }
   }
 
-  MobileFiltersModal.prototype.handleClose = function (event) {
-    if (event) {
-      event.preventDefault()
-    }
-
-    var offsetTop = this.$body.style.top
-    this.$body.style.position = ''
-    this.$body.style.top = ''
-    window.scrollTo(0, parseInt(offsetTop || '0') * -1)
-    this.$module.classList.remove('facets--visible')
-    this.$facetsBox.removeAttribute('aria-modal')
-    this.$facetsBox.removeAttribute('tabindex')
-    this.$focusedElementBeforeOpen.focus()
-
-    document.removeEventListener('keydown', this.$module.boundKeyDown, true)
+  MobileFiltersModal.prototype.handleOpen = function () {
+    this.triggerElement.setAttribute('aria-expanded', 'true')
+    this.module.classList.add('facets--visible')
   }
 
-  MobileFiltersModal.prototype.handleModalFocus = function () {
-    this.$facetsBox.focus()
-  }
-
-  // while open, prevent tabbing to outside the dialogue
-  // and listen for ESC key to close the dialogue
-  MobileFiltersModal.prototype.handleKeyDown = function (event) {
-    var KEY_TAB = 9
-    var KEY_ESC = 27
-
-    switch (event.keyCode) {
-      case KEY_TAB:
-        if (event.shiftKey) {
-          if (document.activeElement === this.$facetsBox) {
-            event.preventDefault()
-            this.$showResultsButton.focus()
-          }
-        } else {
-          if (document.activeElement === this.$showResultsButton) {
-            event.preventDefault()
-            this.$facetsBox.focus()
-          }
-        }
-        break
-      case KEY_ESC:
-        this.$module.close()
-        break
-      default:
-        break
-    }
+  MobileFiltersModal.prototype.handleClose = function () {
+    this.triggerElement.setAttribute('aria-expanded', 'false')
+    this.module.classList.remove('facets--visible')
   }
 
   MobileFiltersModal.prototype.handleClearFilters = function (event) {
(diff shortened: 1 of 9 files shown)


AshGDS:
Hi @andysellick - would you be able to review this? It's a bit of a big change so may be easier to look at it commit by commit or to look at `mobile-filters-modal.js` in your editor. It's my first time contributing SASS/HTML so hopefully everything is OK code wise. Thanks :+1:

app/assets/javascripts/modules/mobile-filters-modal.js
```diff
@@ -2,104 +2,57 @@ window.GOVUK = window.GOVUK || {}
...
-      }
+    if (this.clearFiltersTrigger) {
+      this.clearFiltersTrigger.addEventListener('click', this.module.clearFilters)
     }
+  }
 
-    if (this.$clearFiltersTrigger) {
-      this.$clearFiltersTrigger.addEventListener('click', this.$module.clearFilters)
+  MobileFiltersModal.prototype.handleToggle = function (event) {
+    if (event) {
```
andysellick, line 28: Not sure this check needs to be in place? Is there a situation where `event` wouldn't be defined?

app/assets/javascripts/modules/mobile-filters-modal.js
```diff
@@ -2,104 +2,57 @@ window.GOVUK = window.GOVUK || {}
...
+      event.preventDefault()
+    }
+
+    if (this.module.classList.contains('facets--visible')) {
+      var handleClose = this.handleClose.bind(this)
```
andysellick, line 33: Why do you need to re-bind `handleClose` - can these lines not be `this.handleClose(event)` ?

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -354,3 +354,9 @@ mark {
   padding-left: 13px;
   padding-right: 13px;
 }
+
+.facets__content {
+  .gem-c-select__select--full-width {
```
andysellick, line 359: We shouldn't be styling components in applications - if this is a change that needs to be made to a component, it should be made in the gem, otherwise we're breaking [component isolation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_principles.md#a-component-is-isolated-when).

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -354,3 +354,9 @@ mark {
...
+    min-width: 100%;
```
andysellick, line 360: A select component with the full_width option (which this is) already has width set to 100%, so there's something odd going on here. It looks like the select element itself is escaping from the bounds of the component itself, which I think should be investigated at the component level.

app/views/finders/_facet_collection.html.erb
```diff
@@ -1,4 +1,11 @@
 <div class="filter-form">
+<% if content_item.all_content_finder? %>
+  <%= render partial: 'filter_button'%>
+  <% elsif !content_item.all_content_finder? %>
```
andysellick, line 4: Nitpick: indentation.

app/views/finders/_filter_button.html.erb
```diff
@@ -1,4 +1,5 @@
-<button class="app-c-button-as-link app-mobile-filters-link js-show-mobile-filters"
+<div style="position:relative;">
```
andysellick, line 1: This should be a class rather than an inline style - it's better practice and will soon be outlawed under our CSP.

app/assets/javascripts/modules/mobile-filters-modal.js
```diff
@@ -2,104 +2,57 @@ window.GOVUK = window.GOVUK || {}
...
-    this.$module.clearFilters = this.handleClearFilters.bind(this)
-    this.$module.ModalFocus = this.handleModalFocus.bind(this)
-    this.$module.boundKeyDown = this.handleKeyDown.bind(this)
+  function MobileFiltersModal (module) {
+    this.module = module
+    this.triggerElement = document.querySelector('[data-toggle="mobile-filters-modal"][data-target="' + this.module.id + '"]')
+    this.clearFiltersTrigger = this.module.querySelector('.js-clear-selected-filters')
+    this.module.toggle = this.handleToggle.bind(this)
+    this.module.open = this.handleOpen.bind(this)
+    this.module.close = this.handleClose.bind(this)
```
andysellick, line 11: Not sure where to put this comment, so adding here to allow you to reply. I wanted to ask about the behaviour of the `Show search results` button. It seems to be a jump link to the search results immediately below it, which makes sense now I've thought about it but initially confused me as I assumed it would hide the search filters. Is this intended behaviour?

app/assets/javascripts/modules/mobile-filters-modal.js
```diff
@@ -108,34 +61,37 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-            $el.checked = false
-          } else if ($el.type === 'text' && $el.value !== '') {
-            $el.value = ''
+          if (el.type === 'checkbox' && el.checked === true) {
+            el.checked = false
+          } else if (el.type === 'text' && el.value !== '') {
+            el.value = ''
           }
           break
         case 'SELECT':
```
andysellick, line 79: Again not sure where to put this comment. On Chrome on the mobile view the contents of the select component appear very small and sometimes not apparently attached to the element itself. Not seeing the same thing in Firefox, so not sure what's going on, probably more to do with the component than this change. Something to investigate separately, perhaps.

Screenshot 2023-02-21 at 15 39 35

app/assets/javascripts/modules/mobile-filters-modal.js
AshGDS: @andysellick Yep this is intentional - at first I removed the button entirely, and even though the results update automatically the designer thought it's better to have a button which jumps to the results.

app/assets/javascripts/modules/mobile-filters-modal.js
AshGDS: @andysellick I believe it's because the tests call the functions without firing an event, so you wouldn't be able to call `.preventDefault()` on something that's undefined. Is this OK?

app/assets/javascripts/modules/mobile-filters-modal.js
AshGDS: Good spot, I think it can be `this.module.close()` which is already binded to `this`. I think the binding is needed because within that close function, it calls `this.module` etc, and without the binding, the scope of `this` would be the function itself.

app/assets/stylesheets/finder_frontend.scss
AshGDS: @andysellick Yep it has `width: 100%` on it, but `.govuk-select` has `min-width: 11.5em;` which causes the issue. Should I add the `min-width: 100%` to the select component instead?

app/views/finders/_filter_button.html.erb
AshGDS: Whoops, good spot, was meant to be temporary for testing!

app/assets/javascripts/modules/mobile-filters-modal.js
AshGDS: @andysellick Are you using an external monitor? Mine looks like this 
image

AshGDS:
Thanks @andysellick - I've resolved the changes I could make, but I have a question regarding the `min-width` thing. Is it easier to fix that as a separate PR instead?

andysellick:
> Thanks @andysellick - I've resolved the changes I could make, but I have a question regarding the `min-width` thing. Is it easier to fix that as a separate PR instead?

Thanks, I'll do another review. Yes, probably best to sort out that issue in a separate PR, as I guess it might involve changing the component.

app/assets/stylesheets/finder_frontend.scss
andysellick: Agree that it's probably easier to address this in a separate PR.

app/assets/javascripts/modules/mobile-filters-modal.js
andysellick: Thanks. I think now that this might be a significant enough change that we should do some user testing to make sure we've got it right. That would make this piece of work a lot bigger and more complex though. I'll put a note on the trello card and we can discuss it further offline.

app/assets/javascripts/modules/mobile-filters-modal.js
andysellick: It happens when I switch to mobile view in Chrome but keep the window wide enough to still see the desktop site. Have raised an issue about it to avoid blocking this PR: https://github.com/alphagov/govuk_publishing_components/issues/3280

app/views/finders/_facet_collection.html.erb
```diff
@@ -1,4 +1,11 @@
 <div class="filter-form">
+  <% if content_item.all_content_finder? %>
+      <%= render partial: 'filter_button'%>
+    <% elsif !content_item.all_content_finder? %>
+      <% if facets.select(&:filterable?).any? %>
+        <%= render partial: 'filter_button'%>
+      <% end %>
+  <% end %>
```
andysellick, line 8: Nitpick: there still seems to be something wrong with the indentation here.

app/views/finders/_facet_collection.html.erb
```diff
@@ -1,4 +1,11 @@
 <div class="filter-form">
+  <% if content_item.all_content_finder? %>
```
andysellick, line 2: Is this `if` block over-complicated? Could it be written instead as

```Ruby
<% if content_item.all_content_finder? or facets.select(&:filterable?).any? %>
  <%= render partial: 'filter_button'%>
<% end %>
```

app/views/finders/_facet_collection.html.erb
AshGDS: Whoops, my bad!

app/views/finders/_facet_collection.html.erb
AshGDS: No worries, yeah I didn't change any Ruby logic in this PR luckily! If it was on one line, wouldn't it need to be:

```Ruby
<% if content_item.all_content_finder? or !content_item.all_content_finder? and facets.select(&:filterable?).any? %>
  <%= render partial: 'filter_button'%>
<% end %>
```

as in the original, the `elsif` only runs if `content_item.all_content_finder?` is `false`, whereas that simplified solution wouldn't check for that.

app/assets/stylesheets/finder_frontend.scss
AshGDS: Thanks, I removed it for now.

app/assets/javascripts/modules/mobile-filters-modal.js
AshGDS: @andysellick  The UI changes were posted in the accessibility community, and there were no major issues found - initially there were some changes requested, but they were rescinded after further investigation. Therefore, it has had some internal testing already. Before it was rescinded, one of the suggestions was to change the text to something like "Go to search results" - maybe we changing the text would be useful? Or "Jump to search results" or "Skip to search results"?

AshGDS:
Thanks @andysellick - I removed the UI bug fix, fixed the indentation, and responded to the user testing suggestion.

app/assets/javascripts/modules/mobile-filters-modal.js
andysellick: `Go to search results` sounds good, let's change it to that.

app/views/finders/_facet_collection.html.erb
andysellick: Ummm... still not quite right. The inner lines are partly indented by 2 spaces, partly by 4. Can you take another look?

AshGDS: @andysellick My bad, I've been using the `tab` key on Visual Studio Code to indent this, so I'm guessing it's judgement is wrong with `.erb` files.

app/views/finders/_facet_collection.html.erb
andysellick: Yes, I think my solution was a little too simple, but isn't the elsif unnecessary? The code is basically saying

```
if a_thing
  (render partial)
elsif not a_thing
  if another_thing
    (render partial)
  end
end
```

i.e. there are only two possible outcomes for the top level if statement - a_thing is either or not, so it's always going to run one of those statements. Doesn't that mean it could be

```
if a_thing
  (render partial)
elsif another_thing
  (render partial)
end
```

I concede that I still might be missing something, it's Friday afternoon, so happy for you to point out the flaw in my logic.

app/views/finders/_facet_collection.html.erb
AshGDS: Actually it's probably my fault. @andysellick Does it look OK now? If I got it wrong again, may be worth writing out the correct indentation 😅

app/views/finders/_facet_collection.html.erb
AshGDS: @andysellick That example makes sense - I think because your original suggestion uses `or` it would've changed the flow to:

```
    if a_thing or another_thing
        (render partial)
```

instead of 

```
    if a_thing
      (render partial)
    elsif another_thing
      (render partial)
    end
```

and no worries, I have Friday brain too

andysellick: Thanks for discussing this with me just now Ash. As discussed, I think we both understand this a bit more clearly but on the basis that this is a bit of code that's been moved rather than written as part of this PR, I think we should leave it as it is for now.

andysellick (approved):
All good, happy to approve. One thing before you merge, can any of the commits be squashed together?

AshGDS:
Thanks @andysellick - will do :+1:

PR #2994, opened 2023-02-03
Add option to remove top border on first list item

jon-kirwan:
## What
https://trello.com/c/dhg330gw/1785-remove-grey-top-border-on-the-top-item-on-the-list-on-specialist-finders

On specialist finder pages (or finder pages where no sort options are rendered) a double border is seen since no sort options are displayed (see _before_ screenshot below). The border on the first document list item should be removed. On the main search results page https://www.gov.uk/search/all?keywords=parliament&order=relevance the top border on the first document list item is necessary to visually separate the sort options from the document list.

For any specialist finder pages (or finder pages with no sort options) it should be true.

- Add option `remove_top_border_from_first_child` to remove the top border from the first item of the document list
- Amend spacing when there are no sort options for consistent item spacing

**Review URL(s)**

- https://finder-frontend-pr-2994.herokuapp.com/aaib-reports
- https://finder-frontend-pr-2994.herokuapp.com/search/all (unchanged)

## Why
On some finder pages e.g. [Air Accidents Investigation Branch reports](https://www.gov.uk/aaib-reports) a double border can be seen when no sort options are present.

The option `remove_top_border_from_first_child` will remove the top border from the first item only.

### DIFF
--- a/app/presenters/result_set_presenter.rb
+++ b/app/presenters/result_set_presenter.rb
@@ -50,6 +50,10 @@ class ResultSetPresenter
     presenter[:options].find { |o| o[:selected] }
   end
 
+  def has_sort_options
+    sort_presenter.to_hash.blank? ? true : false
+  end
+
 private
 
   attr_reader :metadata_presenter_class, :sort_presenter, :total, :documents, :facets, :content_item
--- a/app/views/finders/_search_results.html.erb
+++ b/app/views/finders/_search_results.html.erb
@@ -5,7 +5,8 @@
 <div class="finder-results js-finder-results">
   <%= render "govuk_publishing_components/components/document_list", {
     items: local_assigns[:document_list_component_data],
-    remove_underline: true
+    remove_underline: true,
+    remove_top_border_from_first_child: result_set_presenter.has_sort_options
   } %>
 </div>
 
--- a/app/views/finders/show.html.erb
+++ b/app/views/finders/show.html.erb
@@ -53,7 +53,7 @@
         <% end %>
         data-module="gem-track-click"
         >
-        <div id="js-search-results-info" data-module="remove-filter" class="result-info">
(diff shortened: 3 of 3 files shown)


app/presenters/result_set_presenter.rb
```diff
@@ -50,6 +50,10 @@ def sort_option
     presenter[:options].find { |o| o[:selected] }
   end
 
+  def has_sort_options
```
leenagupte, line 53: What does this method name mean?

app/views/finders/_search_results.html.erb
```diff
@@ -5,7 +5,8 @@
 <div class="finder-results js-finder-results">
   <%= render "govuk_publishing_components/components/document_list", {
     items: local_assigns[:document_list_component_data],
-    remove_underline: true
+    remove_underline: true,
+    remove_top_border_from_first_child: result_set_presenter.has_sort_options
```
leenagupte, line 9: Shouldn't this always be true?

leenagupte, line 9: Is there something where this is set to true for the 1st child?

jon-kirwan: Not with the main search results page https://www.gov.uk/search/all?keywords=parliament&order=relevance since the top border is necessary to visually separate the sort options from the document list and should be false.

For any specialist finder pages (or finder pages with no sort options) it should be true.

app/presenters/result_set_presenter.rb
leenagupte: Ah pk, after reading your response to my other comment, this makes more sense. It would really help if you add this reasoning to the commit message.

jon-kirwan: Thanks, yep that would be helpful. I've now updated the What section to try and better describe the issue.

leenagupte (approved):
Thanks for updating the commit message. 👍

PR #3021, opened 2023-03-16
Add new large variation to option-select component

MartinJJones:
## What

Add new large variation to option-select component. When `large` is set to `true`, the max height is increased to 600px, allowing more items to be visible.

## Why

This will allow for better use of space, especially on larger screens when only a couple of facets are used on the page.

Trello: https://trello.com/c/Urpavm6J/1886-increase-the-height-of-the-industries-facet-filter-box

---

### DIFF
--- a/app/assets/stylesheets/components/_option-select.scss
+++ b/app/assets/stylesheets/components/_option-select.scss
@@ -94,6 +94,10 @@
   }
 }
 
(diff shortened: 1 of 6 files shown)


jon-kirwan (approved):
Checked on Integration and looks all good to me!

PR #3028, opened 2023-03-27
Disable LTR for specialist finder path

KludgeKML:
Disables LTR ranking using the existing licence_transaction_path? method in the QueryBuilder.

Use hardcoded utility function introducted to catch boosts
and also turn off LTR ranking using ab_tests=disable:relevance
on that path, which improves search results on the licence
finder.

We're using the ab parameters relevance:disable to do this.
Since this isn't strictly an A/B thing it feels a bit wrong to
use this parameter, but it's all that's available to us at the
moment. We'll add a tech debt card to examine making this a
proper parameter.

https://trello.com/c/q3JBrj5U/1970-outcome-of-spike-switch-off-learn-to-rank

### DIFF
--- a/app/lib/search/query_builder.rb
+++ b/app/lib/search/query_builder.rb
@@ -234,6 +234,10 @@ module Search
     end
 
     def ab_query
+      # We're using the ab test relevance:disable params here to turn off LTR for this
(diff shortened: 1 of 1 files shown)


1pretz1 (approved):
LGTM

PR #3030, opened 2023-03-28
Add boost fields query for licence finder 

1pretz1:
This adds a new query param that tells Search API to weight matches
against licence_transaction_industry field higher than indexable_content
(body etc).

More info in the PR implementing the change [1].

[1]: https://github.com/alphagov/search-api/pull/2552

Depends on:

- https://github.com/alphagov/finder-frontend/pull/3030

Trello:
https://trello.com/c/VhsBND5i/1888-boost-industry-field

### DIFF
--- a/app/lib/search/query_builder.rb
+++ b/app/lib/search/query_builder.rb
@@ -31,6 +31,7 @@ module Search
         debug_query,
         ab_query,
         suggest_query,
+        boost_fields_query,
       ].reduce(&:merge)
 
       return [base_query] if filter_queries.empty?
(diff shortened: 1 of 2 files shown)


app/lib/search/query_builder.rb
```diff
@@ -240,6 +241,12 @@ def suggest_query
       { "suggest" => "spelling_with_highlighting" }
     end
 
+    def boost_fields_query
```
KludgeKML, line 244: Hey, do we want to take advantage of something like this here?:
https://github.com/alphagov/finder-frontend/pull/3028/files
https://github.com/alphagov/govuk-helm-charts/pull/1040
(we could add an ALLOW_BOOST_FIELDS or something like that)

KludgeKML (approved):
LGTM

app/lib/search/query_builder.rb
1pretz1: Comment resolved async. We'll probably move this config to the licence transaction [JSON file in Specialist Publisher](https://github.com/alphagov/specialist-publisher/blob/main/lib/documents/schemas/licence_transactions.json) after we've tested it as that seems a better place if this feature is wanted long term (more work to implement so not doing it now).

PR #3050, opened 2023-05-02
Add transformative redirects to finder frontend

callumknights:
This PR adds both the consultations and statistics transformative redirects that were previously handled by whitehall into finder-frontend.

These need to be handled differently to other redirects being moved from Whitehall as there is some setting of query string parameters currently done by whitehall, as well as transformation from legacy params to newer ones. 

This is done as part of the mission to remove any frontend rendering form whitehall, as these redirects are handled by whitehall frontend instances, they needed to be moved. 

Trello: https://trello.com/c/SmdAt9SX/523-migrate-redirect-logic-from-whitehall-to-finder-frontend, [Jira issue PP-716](https://gov-uk.atlassian.net/browse/PP-716)

### DIFF
--- a/app/controllers/redirection_controller.rb
+++ b/app/controllers/redirection_controller.rb
@@ -26,9 +26,13 @@ class RedirectionController < ApplicationController
                             .permit(:level_one_taxon, :level_two_taxon, departments: [], world_locations: [], content_store_document_type: [])
                             .transform_keys { |k| k == "departments" ? "organisations" : k }
                             .compact
-
     redirect_params[:content_store_document_type] = %w[open_consultations closed_consultations]
+
     redirect_to(finder_path("search/policy-papers-and-consultations", params: { order: "updated-newest" }.merge(redirect_params)))
+  end
+
+  def redirect_statistics_announcements
+    redirect_to(finder_path("search/research-and-statistics", params: statistics_announcements_topic_and_other_params))
   end
 
   def advanced_search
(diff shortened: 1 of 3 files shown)


app/controllers/redirection_controller.rb
```diff
@@ -16,6 +16,22 @@ def redirect_latest
...
+  def redirect_consultations
+    topics = {}
+    topics["level_one_taxon"] = params[:topics] if params[:topics]
+    topics["level_two_taxon"] = params[:subtaxons] if params[:subtaxons]
+
+    redirect_params = params.merge(topics)
+                            .slice(:departments, :level_one_taxon, :level_two_taxon, :world_locations, :content_store_document_type)
+                            .permit(:level_one_taxon, :level_two_taxon, departments: [],  world_locations: [], content_store_document_type: [])
+                            .transform_keys { |k| k == "departments" ? "organisations" : k }
+                            # .transform_values { |v| v == (["open_consultations"] || ["closed_consultations"] || ["policy_papers"]) ? %w[open_consultations closed_consultations] : v }
```
brucebolt: You probably want to delete this commented line from the first commit.

app/controllers/redirection_controller.rb
```diff
@@ -16,6 +16,22 @@ def redirect_latest
...
   end
 
+  def redirect_consultations
+    topics = {}
+    topics["level_one_taxon"] = params[:topics] if params[:topics]
+    topics["level_two_taxon"] = params[:subtaxons] if params[:subtaxons]
+
+    redirect_params = params.merge(topics)
+                            .slice(:departments, :level_one_taxon, :level_two_taxon, :world_locations, :content_store_document_type)
+                            .permit(:level_one_taxon, :level_two_taxon, departments: [],  world_locations: [], content_store_document_type: [])
```
brucebolt: There's two spaces between the end of departments array and world locations.  It's fixed in the next commit, but good to do that here too.

brucebolt (approved):
Looks good to me, especially since we've discussed a lot of this on Slack already 👍.

Just added a couple of comments about tidying up the first commit (since there's a bit of extra diff in both commits that's probably not needed).

app/controllers/redirection_controller.rb
callumknights: 👍

app/controllers/redirection_controller.rb
callumknights: 👍

PR #3053, opened 2023-05-10
AA test for Search Results

Rosa-Fox:
[Trello](https://trello.com/c/AH3QERus/613-deploy-the-50-50-aa-test-for-the-search-team)

The purpose of the A/A test is to test the % of allocation of traffic/sample population based on the current version of elastic (6.7) against itself. A (ES 6.7) vs A (ES6.7). The A/A is solely to identify how the approach allocates traffic. This will provide benchmark data to aid analysis in future AA tests.

This test is only applied to `/search/all` results, not to finder search results.

### Manual testing using Postman on /search/all for the different variants

Screenshot 2023-07-03 at 16 25 43
Screenshot 2023-07-03 at 16 25 19
Screenshot 2023-07-03 at 16 25 00

### A finder search results page, which the test is not applied to:

Screenshot 2023-07-04 at 12 26 26

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -1,5 +1,6 @@
 class FindersController < ApplicationController
   layout "finder_layout"
+  include AbTests::ElasticSearchAaTestable
 
   before_action do
     set_expiry(content_item)
@@ -7,6 +8,10 @@ class FindersController < ApplicationController
 
   def show
     slimmer_template "gem_layout_full_width" if i_am_a_topic_page_finder
+
+    if page_under_test?
+      set_requested_variant
+    end
 
     respond_to do |format|
       format.html do
@@ -58,7 +63,7 @@ private
 
   attr_reader :search_query
 
-  helper_method :facet_tags, :i_am_a_topic_page_finder, :result_set_presenter, :content_item, :signup_links, :filter_params, :facets
+  helper_method :facet_tags, :i_am_a_topic_page_finder, :result_set_presenter, :content_item, :signup_links, :filter_params, :facets, :page_under_test?
 
   def redirect_to_destination
     @redirect = content_item.redirect
--- a/app/views/finders/_show_header.html.erb
+++ b/app/views/finders/_show_header.html.erb
@@ -66,6 +66,12 @@
       <% end %>
     </div>
 
+    <% if page_under_test? %>
+      <% content_for :meta_tags do %>
+        <%= @requested_variant.analytics_meta_tag.html_safe %>
+      <% end %>
+    <% end %>
+
     <% if content_item.summary %>
       <div class="govuk-grid-column-two-thirds">
         <div class="metadata-summary ">
--- a/spec/controllers/finders_controller_spec.rb
+++ b/spec/controllers/finders_controller_spec.rb
@@ -40,6 +40,7 @@ describe FindersController, type: :controller do
   after { Rails.cache.clear }
 
   describe "GET show" do
+    render_views
     describe "a finder content item exists" do
       before do
         stub_content_store_has_item(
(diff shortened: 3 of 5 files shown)


app/controllers/finders_controller.rb
```diff
@@ -8,6 +8,16 @@ class FindersController < ApplicationController
   def show
     slimmer_template "gem_layout_full_width" if i_am_a_topic_page_finder
 
+    ab_test = GovukAbTesting::AbTest.new(
```
sihugh, line 11: The AB test will run on every page shown by this action, which might not be desired if K&C just want results from /search/all.  
You could add a `page_under_test?` method that tests whether the path is `/search/all`, and wrap this logic in that?

app/views/finders/_show_header.html.erb
```diff
@@ -66,6 +66,10 @@
       <% end %>
     </div>
 
+    <% content_for :meta_tags do %>
+      <%= @requested_variant.analytics_meta_tag.html_safe %>
```
sihugh, line 70: if you do add a `page_under_test?` method, you'd need to wrap this bit in it too

app/controllers/finders_controller.rb
```diff
@@ -8,6 +8,16 @@ class FindersController < ApplicationController
...
+      "EsSixPointSeven",
+      dimension: 1,
+      allowed_variants: %w[A B Z],
+      control_variant: "Z",
```
sihugh, line 15: 👍

sihugh (commented):
As Leena mentioned, convention is to put the controller parts of the config in a concern. This test doesn't have any logic though, so it's probably OK. You might want the practise!

app/controllers/finders_controller.rb
Rosa-Fox: Thanks, I checked with them and they only want it on /search/all... which makes sense as this is what they will want to test with the real AB test! Thanks for flagging. I have updated it to use the constraint.

app/controllers/finders_controller.rb
```diff
@@ -353,4 +359,8 @@ def filter_query_array(arr)
       arr.reject { |v| v == "all" }.compact.presence
     end
   end
+
+  def page_under_test?
```
leenagupte, line 363: I think this can be moved to the concern?

app/controllers/finders_controller.rb
```diff
@@ -8,6 +9,11 @@ class FindersController < ApplicationController
   def show
     slimmer_template "gem_layout_full_width" if i_am_a_topic_page_finder
 
+    if page_under_test?
+      @requested_variant = elastic_search_aa_test.requested_variant(request.headers)
+      @requested_variant.configure_response(response)
+    end
+
```
leenagupte, line 16: I think all of this can be moved to the concern too, to keep the controller tidy.

leenagupte (commented):
I think this looks ok. I've added a couple of suggestions for keeping the AB test code out of controller. I pretty much "borrowed" all suggestions from [Kelvin](https://github.com/alphagov/frontend/blob/main/app/controllers/concerns/bank_hol_ab_testable.rb) 😅

app/controllers/finders_controller.rb
Rosa-Fox: done!

app/controllers/finders_controller.rb
Rosa-Fox: done!

spec/controllers/finders_controller_spec.rb
```diff
@@ -108,6 +109,52 @@
...
+
+            get :show, params: { slug: "lunch-finder" }
+
+            expect(response.status).to eq(200)
+            expect(response.header["Vary"]).not_to eq("GOVUK-ABTest-EsSixPointSeven")
+            expect(response.body).not_to include("EsSixPointSeven:#{variant}")
+          end
+        end
+
+        it "should render the page without an AB test variant for search" do
```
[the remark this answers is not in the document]
Rosa-Fox: yeah, tbh this was more out of paranoia that the page would still render if something went wrong with setting the variant, but its probably not needed as it should work!

gclssvglx (approved):
Looks great - thanks Rosa 👍

PR #3057, opened 2023-05-11
Update results heading text for licence finder

jon-kirwan:
## What
Update results heading text for licence finder pages.

**Review URL(s)**

- https://finder-frontend-pr-2994.herokuapp.com/aaib-reports
- https://finder-frontend-pr-2994.herokuapp.com/search/all (unchanged)

## Why

Following a review with an accessibility specialist we found the results heading text should be clearer for users (particularly screen reader users).

### Before (Air Accidents Investigation Branch reports)
TBC

### After (Air Accidents Investigation Branch reports)
TBC


app/presenters/result_set_presenter.rb
```diff
@@ -21,7 +21,11 @@ def initialize(content_item, facets, results, filter_params, sort_presenter, met
   end
 
   def displayed_total
-    "#{number_with_delimiter(total)} #{pluralised_document_noun}"
+    heading_text = "#{number_with_delimiter(total)} #{pluralised_document_noun}"
+    # TODO: to be displayed for all finders pending content + design review. See: https://trello.com/c/bWnyVnhX/1983-update-x-licences-results-text
+    return "results: #{heading_text}" if content_item.document_noun == "licence"
```
maxgds, line 26: Let's avoid the bare string for "results" and put that into the translation files for consistency/good practice.

maxgds (changes requested):
See comment - re bare string.

app/presenters/result_set_presenter.rb
jon-kirwan: There's not too much in the way of translations for finder frontend but i've moved the whole sentence in to the search_result_presenter yaml file (with a few changes required by Rubocop). I've added as a separate commit here but i'll squash in to one commit when approved.

app/views/finders/show.html.erb
```diff
@@ -55,7 +55,7 @@
         >
         <div id="js-search-results-info" data-module="remove-filter" class="result-info <%= "govuk-!-margin-bottom-0" if result_set_presenter.has_sort_options %>">
           <div class="govuk-grid-row">
-            <div class="govuk-grid-column-one-third">
+            <div class="govuk-grid-column-one-half">
```
maxgds, line 58: This seems unrelated to the work described - nothing wrong with it, but could you include it in the PR description if it's intentional to include here.

maxgds (approved):
Code looks sound - just a request to update the PR description to include the column width change before merge.

PR #3071, opened 2023-05-25
Alter option select facet cache key to be sensitive to ordering.

KludgeKML:
- Making a digest of the map means that comparisons are sensitive to ordering of values in the hashes.

Currently reordering facets or the allowed values of facets currently won't bust the cache (which means that if we do a backend change of the order of facets we have to manually clear the cache to get the change visible). By altering the cache key to be dependent on ordering, we can make the cache key different, allowing it to refresh.

(Note that swapping the order of entire facets isn't cached, so already updates automatically)

https://trello.com/c/tQOVFnaT/2017-adjust-finder-frontend-caching-strategy-so-that-changing-order-or-name-of-facets-clears-cache

### DIFF
--- a/app/models/option_select_facet.rb
+++ b/app/models/option_select_facet.rb
@@ -51,7 +51,7 @@ class OptionSelectFacet < FilterableFacet
   end
 
   def cache_key
-    { selected: selected_values, allowed: allowed_values }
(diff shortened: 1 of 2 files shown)


1pretz1 (approved):
Nice work!

PR #3085, opened 2023-06-14
Licence stopwords

KludgeKML:
Refactors the existing stopword code, then uses that refactoring to add stopwords for the new licence finder.

https://trello.com/c/MEsGJsgP/2060-implement-stopwords-for-permit-licence-license-etc


app/lib/search/query_builder.rb
```diff
@@ -255,185 +440,8 @@ def licence_transaction_path?
...
-        yours
-        yourself
-        yourselves
-      ]
-
-      brexit_stopwords = %w[abroad brexit eu europe european exit leave union]
-
-      (generic_stopwords + brexit_stopwords).uniq
+    def brexit_path?
+      finder_content_item.base_path == "/find-eu-exit-guidance-business"
```
1pretz1, line 444: If the brexit stopwords were added just for one finder, do we even need them anymore? `/find-eu-exit-guidance-business` now redirects to a collection page https://www.gov.uk/government/collections/brexit-guidance.

1pretz1 (commented):
Nice work! I've left a comment on whether we even need the brexit stopwords anymore (assume we don't).

I'm also curious about the generic stopwords, which will only apply to the licence finder tool (if we remove brexit stuff). I think we should either bin them completely or add them to all finders. Binning them would be safer but we could be missing out on an opportunity. Maybe we should remove and track this somewhere so it could be looked into at a future point? (generic stopwords here could also be nullified by the new search team)

What do you think?

app/lib/search/query_builder.rb
KludgeKML: Have removed brexit stopwords. I've also cut out the generic stopwords - I think your second plan is the best, let's remove them for now and see what happens with the new search team. We can add back the generic stopword list if it turns out it's useful.

PR #3097, opened 2023-06-30
Remove subscribe to feed link for licence finder

1pretz1:
Collaborated with @MartinJJones for this piece of work 🙌 

The plan is to remove the feed links from all finders, unless we gain
evidence that it is needed. Until this decision is made, we don't want to
introduce the feature for the new licence finder.

We've also had to update the JS to keep the live search feature working,
as removing the link removed this functionality. We've now added JS tests 
to ensure the feature works when the feed link isn't present.

Trello:
https://trello.com/c/9RFdV1KK/2094-remove-feed-link-from-only-the-new-licence-finder

## Visual Changes

| Visible | Removed (Licence Finder) |
| --- | --- |
| subscribe-to-feed-footer | subscribe-to-feed-removed-top |
| subscribe-to-feed-head | subscribe-to-feed-footer-removed |

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -23,7 +23,9 @@
     this.$sortBlock = options.$results.querySelector('#js-sort-options')
     this.$paginationBlock = options.$results.querySelector('#js-pagination')
     this.action = this.$form.getAttribute('action') + '.json'
-    this.$atomAutodiscoveryLink = options.$atomAutodiscoveryLink
+    if (options.$atomAutodiscoveryLink) {
+      this.$atomAutodiscoveryLink = options.$atomAutodiscoveryLink
+    }
 
     this.baseTitle = document.querySelector("meta[name='govuk:base_title']")
     if (this.baseTitle) {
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -23,7 +23,9 @@
     this.$sortBlock = options.$results.querySelector('#js-sort-options')
     this.$paginationBlock = options.$results.querySelector('#js-pagination')
     this.action = this.$form.getAttribute('action') + '.json'
-    this.$atomAutodiscoveryLink = options.$atomAutodiscoveryLink
+    if (options.$atomAutodiscoveryLink) {
```
KludgeKML, line 26: (Just a request for clarification) where does options.$atomAutodiscoverLink get set?

maxgds:
Query: Has this been captured as tech debt so we don't forget to remove it later? Or is the removal likely to involve working in the same space anyway?

app/assets/javascripts/live_search.js
MartinJJones: Good question, it gets set in application.js - https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/application.js#L37

I've not come across this approach before in our codebase and there is a comment in this file saying this is not advisable - https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/application.js#L7-L8

It's likely possible to clean this up with some refactoring I imagine.

KludgeKML: Oh, I see - options is just a sort of hold-all for all the named parameters passed in when the GOVUK.LiveSearch is created. Cool.

KludgeKML (approved):
Looks okay to me (seconding Max's comment, though)

1pretz1:
When the decision on feed links is decided then the changes will take place in the same part of the codebase @maxgds. A tech debt card does sound like a good idea though so removing the feed link doesn't slip from our (or another teams) radar. 

We could either create a general tech debt card, or ensure that the team leading on the removal (Kuba's team?) has it tracked on their board?

jon-kirwan (approved):
All looks fine and works as expected for me 👍

PR #3108, opened 2023-07-11
GA4 Search Fixes

AshGDS:
This PR contains two fixes for the GA4 search tracking.

1. When tracking searches, it will now make sure the search term doesn't contain any extra spaces, and that the search term is lowercase.

2. When you initially load a search result, the page URL looks like `?keywords=hello+world`. However when you update a filter, the URL becomes `?keywords=hello%20world` because JavaScript is updating the URL using `encodeURI` which turns spaces into `%20`. This causes an issue for analysts as the URLs in the dashboard become inconsistent having a mixture of `%20` and `+`.  Therefore, in this PR I am manually changing any spaces to `+` before `encodeURI` runs, to get rid of any `%20` references in the `keywords` part of the URL.


JamesCGDS (approved):
Looks good! 👍

PR #3155, opened 2023-09-12
Remove `whenever` gem

csutter:
Scheduled tasks now run as Kubernetes cronjobs and this gem is no longer used. It also breaks the Rails console due to an incompatibility with Ruby 3.2, so this is a good opportunity to get rid of it!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,6 @@ gem "sassc-rails"
 gem "slimmer"
 gem "sprockets-rails"
 gem "uglifier"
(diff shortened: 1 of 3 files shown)


sihugh (approved):
Scheduled tasks are now [cronTasks](https://github.com/alphagov/govuk-helm-charts/blob/e97b9ad3753da83262bc21c572531d3d1988f6f8/charts/app-config/values-integration.yaml#L1113-L1116)

PR #3156, opened 2023-09-12
Allow manually calling `search-api-v2`

csutter:
As a first step to building the integration with `search-api-v2`, allow `finder-frontend` to call it if and only if an explicit `use_v2` query parameter is supplied.

- Bump `gds-api-adapters` to get new adapter for `search-api-v2`
- Add `Services#search_api_v2`
- Update `Search::Query` to call `search-api-v2` if `use_v2` parameter is present and truthy (accoring to `ActiveModel::Type::Boolean`)

### DIFF
--- a/app/lib/search/query.rb
+++ b/app/lib/search/query.rb
@@ -87,7 +87,11 @@ module Search
         override_sort_for_feed:,
       ).call
 
-      if queries.one?
+      if use_v2_api?
+        GovukStatsd.time("search_api_v2.finder_search") do
+          Services.search_api_v2.search(queries.first).to_hash
+        end
+      elsif queries.one?
         GovukStatsd.time("rummager.finder_search") do
           Services.rummager.search(queries.first).to_hash
         end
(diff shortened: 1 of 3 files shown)


app/lib/search/query.rb
```diff
@@ -99,5 +103,9 @@ def fetch_search_response(content_item)
         end
       end
     end
+
+    def use_v2_api?
+      ActiveModel::Type::Boolean.new.cast(filter_params["use_v2"])
```
leenagupte, line 108: This is nice!

Gemfile.lock
```diff
@@ -145,7 +145,7 @@ GEM
     factory_bot (6.3.0)
       activesupport (>= 5.0.0)
     ffi (1.15.5)
-    gds-api-adapters (90.0.0)
+    gds-api-adapters (91.0.0)
```
leenagupte, line 148: Not blocker: We'd normally put gem updates in a separate commit to increase readability and reduce noise in the commit.

leenagupte (approved):
👍

Gemfile.lock
csutter: Good shout – I can see Dependabot has already created #3153, I'll let you merge that and will rebase this PR!

PR #3166, opened 2023-09-21
Spike: Refactor option-select component to isolate it from GA4 functionality specific to finder-frontend

floehopper:
Trello: https://trello.com/c/qs60lCxl

This is a spike into moving the GA4-related code out of the `option-select` component prior to extracting it into the `govuk_publishing_components` gem. My motivation was that I felt the GA4-related code was quite specific to finder-frontend.

Broadly the approach I've tried to take is to move the setting of all GA4-related data attributes out into the Rails facet templates rather than relying on setting them in a combination of in the component templates, in the component JavaScript, and in some more "global" GA4 JavaScript. This branch is very much a spike so I've made no attempt at fixing any linting or tests and while I've tried hard to make the commit notes as detailed and useful as possible, the last few are marked as "WIP" because either I'm not very happy with them or there is still some work to do in them.

I think my conclusion is that, while this specific approach might not be the best, it might be worth going for some kind of middle ground where it's possible to enable GA4-related functionality within the components, but supply non-generic values as data attributes from the facet templates. I've tried to explain this in a bit more detail in the final commit note.

I'm very happy to chat about any of this if it would be useful! Otherwise I'll probably close the PR fairly soon.

/cc @andysellick

### DIFF
--- a/app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
+++ b/app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
@@ -147,7 +147,6 @@
     },
 
     addFilterButtonTracking: function (button, section) {
-      button.setAttribute('data-ga4-expandable', '')
       var ga4JSON = { event_name: 'select_content', type: 'finder', section: section }
 
       try {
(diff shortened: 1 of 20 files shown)


andysellick:
Relevant improvements from this PR have been implemented in this PR: https://github.com/alphagov/finder-frontend/pull/3224 so closing this one.

PR #3169, opened 2023-09-26
Add user research banner

catalinailie:
Add user research banner on HMRC contact page.

Before:
Screenshot 2023-09-26 at 10 03 01
After:
Screenshot 2023-09-26 at 10 02 49

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -17,6 +17,8 @@ class FindersController < ApplicationController
       format.html do
         raise UnsupportedContentItem unless content_item.is_finder?
 
+        @show_banner = content_item.base_path == "/government/organisations/hm-revenue-customs/contact"
+
         if legacy_params_present?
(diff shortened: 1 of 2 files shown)


app/controllers/finders_controller.rb
```diff
@@ -17,6 +19,8 @@ def show
       format.html do
         raise UnsupportedContentItem unless content_item.is_finder?
 
+        @show_banner = show_banner?(request.path)
```
leenagupte, line 22: The base path could be used here as the controller has access to the [content item](https://github.com/alphagov/finder-frontend/blob/7cd96b16dcbe1f8a27fdc79dcb2313bdd5e8dedd/app/controllers/application_controller.rb#L57).

leenagupte (approved):
This looks fine to me, but I think a frontend person who is more familiar with the component system should check that the stylesheets are being imported correctly.

Backend-wise, it looks fine. I'm not sure if a whole Helper class is needed to check the base path, however that's minor and it all seems to work, so 👍

MartinJJones (approved):
Changes look good to me 👍

PR #2286, opened 2020-11-02
Disable ecommerce tracking for more pages

benthorner:
These are currently leading to Smokey (JS) failures in Production.

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/finder-frontend), after merging.

PR #3174, opened 2023-09-29
Separate UA ecommerce tracking enabler from GA4 ecommerce tracking enabler

AshGDS:
## What
- Our GA4 ecommerce tracking was missing on these three finders: [1](https://www.gov.uk/world/organisations) [2](https://www.gov.uk/government/organisations/hm-revenue-customs/contact) [3](https://www.gov.uk/government/groups)
- This is because there is a boolean called `@include_ecommerce` in this repo, which determined wether UA ecommerce tracking was enabled or not.
- UA ecommerce tracking was specifically disabled on those three finders. This is because the payload they sent to UA was too large, resulting in a JavaScript error. See https://github.com/alphagov/finder-frontend/pull/2286. 
- The GA4 ecommerce tracking was also using this `@include_ecommerce` boolean to determine wether it was enabled or not.
- As we want to enable GA4 ecommerce tracking everywhere, but keep those three pages disabled for UA tracking, I have created a new boolean called `@include_ga4_ecommerce_tracking` and set it to `true`. When `true`, `data-ga4-ecommerce-path` is appended to the search result links, which allows the JS to use the links for the `ecommerce` array and the `select_item` events.
- This way, we can enable GA4 ecommerce tracking on these pages, while not interfering with the disabled UA ecommerce tracking.

## Why
- https://trello.com/c/TIFQm2EL/677-fix-hmrc-contacts-finder-ecommerce-tracking
- https://trello.com/c/wLz7uULw/679-fix-world-organisations-finder-ecommerce-tracking
- https://trello.com/c/jHYCtOA8/678-fix-groups-finder-ecommerce-tracking

## How to test
- Go to http://127.0.0.1:3062/government/groups or http://127.0.0.1:3062/world/organisations or http://127.0.0.1:3062/government/organisations/hm-revenue-customs/contact 
- You will see `data-ga4-ecommerce-path` on the links (and `data-ecommerce-index` - i've noted this [here](https://trello.com/c/gFiJmH3D/553-audit-and-ensure-that-every-data-attribute-relies-on-has-ga4-somewhere-in-its-name)) and no UA ecommerce attributes (e.g. `ecommerce_row`).
- You will see the `ecommerce` array populated in the ecommerce dataLayer object.
- If you click a search result and then stop the page loading, you'll see the `select_item` event in the dataLayer object.

### DIFF
--- a/app/presenters/search_result_presenter.rb
+++ b/app/presenters/search_result_presenter.rb
@@ -27,7 +27,7 @@ class SearchResultPresenter
         text: title,
         path: link,
         description: sanitize(summary_text),
-        data_attributes: ecommerce_data(link, title),
+        data_attributes: ga4_ecommerce_data(link).merge(ecommerce_data(link, title)),
       },
       metadata: structure_metadata,
       metadata_raw: metadata,
@@ -80,23 +80,29 @@ private
         GovukError.notify(MalformedPartError.new, extra: { part:, link: })
         next
       end
+      path = "#{link}/#{part[:slug]}"
       {
         link: {
           text: part[:title],
-          path: "#{link}/#{part[:slug]}",
+          path:,
           description: part[:body],
-          data_attributes: ecommerce_data("#{link}/#{part[:slug]}", part[:title], part_index: index),
+          data_attributes: ga4_ecommerce_data(path).merge(ecommerce_data(path, part[:title], part_index: index)),
         },
       }
     end
     structured_parts.compact
   end
 
+  def ga4_ecommerce_data(path)
+    {
+      ga4_ecommerce_path: path,
+    }
+  end
+
   def ecommerce_data(link, title, part_index: nil)
     return {} unless @include_ecommerce
 
     {
-      ga4_ecommerce_path: link,
       ecommerce_path: link,
       ecommerce_row: 1,
       ecommerce_index: document.index,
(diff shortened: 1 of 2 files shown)


app/presenters/result_set_presenter.rb
```diff
@@ -2,11 +2,11 @@ class ResultSetPresenter
   include ERB::Util
   include ActionView::Helpers::NumberHelper
 
-  attr_reader :pluralised_document_noun, :debug_score, :start_offset, :include_ecommerce
+  attr_reader :pluralised_document_noun, :debug_score, :start_offset, :include_ecommerce, :include_ga4_ecommerce_tracking
```
andysellick, line 5: If there's a requirement that we track/don't track some of the finders in GA4 then ignore me, but if we're going to track all of them then I wonder if we should just hard code that? Might make the code a bit simpler - wouldn't have to initialise and pass around variables, and the tests could be simpler.

AshGDS: Thanks Andy, I'll remove the GA4 variable

AshGDS:
Thanks @andysellick - this should be ready for another review :+1:

andysellick (approved):
This is quite complicated! But I think I get it, after reading your PR description and how to test. Looks good! All very useful, thank you.

PR #3193, opened 2023-10-23
Component auditing fixes

andysellick:

## What
Fixes some issues relating to component use in this application. See individual commits for details.

## Why
Issues flagged by our component auditing tools.

## Visual changes
First change shown below is the removal of the `margin: 0` on the intervention component.

Second is the appearance of an inverted header finder before and after. I'm not sure this option is in use anywhere (couldn't see a finder using it) but might as well make it work properly. 

There's a small issue introduced by this change, which is that the metadata component has extra spacing around it when inverted. This appears to be a quirk of the component design and something we could fix in the component, but for now I'm happy to replace a serious accessibility issue with a minor spacing issue. **EDIT** this has now been fixed in the component and applied in the final commit.

Before | After
------ | ------
Screenshot 2023-10-23 at 08 45 09 | Screenshot 2023-10-23 at 08 45 18
Screenshot 2023-10-23 at 09 02 19 | Screenshot 2023-10-23 at 09 15 14

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -13,6 +13,7 @@
 //= require govuk_publishing_components/components/checkboxes
 //= require govuk_publishing_components/components/error-summary
 //= require govuk_publishing_components/components/govspeak
+//= require govuk_publishing_components/components/intervention
 //= require govuk_publishing_components/components/metadata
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
LGTM :+1:

PR #3195, opened 2023-10-25
Duplicate ecommerce attributes, appending ga4 to the names

AshGDS:

## What
- Duplicate ecommerce attributes, appending ga4 to the names.
- The trackers in `govuk_publishing_components` will then use these instead of the UA attributes: https://github.com/alphagov/govuk_publishing_components/pull/3688

## Why
- This is to decouple our GA4 analytics code from universal analytics
- https://trello.com/c/gFiJmH3D/553-audit-and-ensure-that-every-data-attribute-relies-on-has-ga4-somewhere-in-its-name

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -146,6 +146,14 @@
       var consentCookie = GOVUK.getConsentCookie()
 
       if (consentCookie && consentCookie.settings) {
+        if (this.$resultsWrapper) {
+          this.$resultsWrapper.setAttribute('data-ga4-search-query', this.currentKeywords())
+          var sortedBy = this.$resultsWrapper.querySelector('.js-order-results')
+          if (sortedBy) {
+            this.$resultsWrapper.setAttribute('data-ga4-ecommerce-variant', sortedBy.options[sortedBy.selectedIndex].text)
+          }
+        }
+
         GOVUK.analyticsGa4.Ga4EcommerceTracker.init(referrer)
       } else {
         window.addEventListener('cookie-consent', function () {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -146,6 +146,17 @@
...
       if (consentCookie && consentCookie.settings) {
+        if (this.$resultsWrapper) {
+          this.$resultsWrapper.setAttribute('data-ga4-search-query', this.currentKeywords())
+          var sortedBy = this.$resultsWrapper.querySelector('.js-order-results')
+          if (sortedBy) {
+            this.$resultsWrapper.setAttribute('data-ga4-ecommerce-variant', sortedBy.options[sortedBy.selectedIndex].text)
+          }
+        }
+        if (this.$suggestionsBlock) {
+          this.$suggestionsBlock.setAttribute('data-ga4-search-query', this.currentKeywords())
```
andysellick, line 157: Interested to understand why we set `data-ga4-search-query` with the same value on both the suggestions block and the results wrapper. I think the results wrapper is always present, but the suggestions block might/might not be?

andysellick (commented):
Looks good, just one question.

app/assets/javascripts/live_search.js
AshGDS: @andysellick Good spot. I think it's because Universal Analytics ecommerce tracking is on the suggestions block <div>, and the results block <div>. In order for the analysts to know what search term is related to the suggestions <div> and the results <div>, they both have a `data-search-query` value. So they'll always be the same because it's just the keyword used for the search, but they're both different instances of the UA ecommerce tracker, so they each need their own data attribute.

I realised that we don't even use GA4 ecommerce tracking on the suggestions block. We use link tracking for the suggestions instead, and there's also no `data-ga4-ecommerce` attribute on the suggestions. So I'll remove these lines as they're not needed.

andysellick (approved):

PR #3207, opened 2023-11-09
Add DublinCore namespace to atom feed to make it validate

KludgeKML:
The atom builder in Rails will let you put in arbitrary keys, but atom readers in strict mode (such as that used in email-alert-api) will reject the xml as invalid. Rather than write and maintain our own namespace extension, replace the content_id tag with the Dublin Core Elements Identifier tag (dc:identifier) which is semantically appropriate, so that we're serving valid Atom.

https://trello.com/c/ncj8UGmZ/2246-add-content-id-information-to-medical-alerts-rss-feed


KludgeKML:
Closing as we're rethinking how we get these results.

PR #3224, opened 2023-11-29
Refactor GA4 code

andysellick:

## What
Refactor some of the GA4 code in `finder-frontend`, based on @floehopper's previous work in https://github.com/alphagov/finder-frontend/pull/3166. This moves the logic for generating indexes for GA4 tracking on search elements (the option select and expander components) from the JS into Ruby.

Now depends upon this change in the components gem: https://github.com/alphagov/govuk_publishing_components/pull/3750

## Why
We want our code to be clean and readable.

## Visual changes
None.

Trello card: https://trello.com/c/OdYXlYqQ/673-finder-frontend-code-optimisations

### DIFF
--- a/app/assets/javascripts/components/expander.js
+++ b/app/assets/javascripts/components/expander.js
@@ -74,7 +74,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
       try {
         buttonAttributes = JSON.parse(buttonAttributes)
         for (var rawKey in buttonAttributes) {
-          var key = rawKey.replace(/_/i, '-').toLowerCase()
+          var key = rawKey.replace(/_/g, '-').toLowerCase()
           var rawValue = buttonAttributes[rawKey]
           var value = typeof rawValue === 'object' ? JSON.stringify(rawValue) : rawValue
           $button.setAttribute('data-' + key, value)
--- a/app/assets/javascripts/components/expander.js
+++ b/app/assets/javascripts/components/expander.js
@@ -69,17 +69,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
     $button.setAttribute('aria-expanded', expanded)
     $button.setAttribute('aria-controls', this.$content.getAttribute('id'))
 
-    // GA4 Accordion tracking. Relies on the ga4-finder-tracker setting the index first, so we wrap this in a custom event.
-    window.addEventListener('ga4-filter-indexes-added', function () {
-      if (window.GOVUK.analyticsGa4) {
-        if (window.GOVUK.analyticsGa4.Ga4FinderTracker) {
-          window.GOVUK.analyticsGa4.Ga4FinderTracker.addFilterButtonTracking($button, this.$toggle.innerHTML)
+    var buttonAttributes = this.$module.getAttribute('data-button-data-attributes')
+    if (buttonAttributes) {
+      try {
+        buttonAttributes = JSON.parse(buttonAttributes)
+        for (var rawKey in buttonAttributes) {
+          var key = rawKey.replace(/_/i, '-').toLowerCase()
+          var rawValue = buttonAttributes[rawKey]
+          var value = typeof rawValue === 'object' ? JSON.stringify(rawValue) : rawValue
+          $button.setAttribute('data-' + key, value)
         }
+      } catch (e) {
+        console.error('Error with expander button data attributes, invalid JSON passed' + e.message, window.location)
       }
-    }.bind(this))
-
+    }
     $button.innerHTML = toggleHtml
-
     this.$toggle.parentNode.replaceChild($button, this.$toggle)
   }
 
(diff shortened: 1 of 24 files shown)


AshGDS (changes requested):
Looks good, though a small is change needed :+1:

app/views/finders/_radio_facet.html.erb
```diff
@@ -1,7 +1,7 @@
-<div class="facet-container" data-ga4-change-category="update-filter radio" data-ga4-section="" data-ga4-filter-parent>
+<%= tag.div(class: "facet-container", data: { "ga4-change-category": "update-filter radio", "ga4-filter-parent": radio_facet.ga4_section, "ga4-index": { index_section: index + 1, index_section_count: count } }) do %>
```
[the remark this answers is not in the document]
andysellick: Thanks @AshGDS, have restored.

AshGDS (dismissed):
LGTM :+1:

andysellick:
Thanks @AshGDS. I'm currently looking at expanding this PR further, which is going to involve some changes in the gem first, so I'm going to pause on this for the moment (you might want to withdrawn your approval).

andysellick:
@AshGDS have added some more changes to this PR. This should now fully remove the JS for event listening and setting indexes on the option select and expander component buttons. It's related to [this change in the gem](https://github.com/alphagov/govuk_publishing_components/pull/3750) for the option select, so will need to be tested together and I'll figure out how to deploy these two changes simultaneously.

Haven't updated the docs yet, that's another thing to look at, but should be enough to review for now.

AshGDS:
@andysellick Looks good to me :+1:

andysellick:
@AshGDS I've bumped the gem in this PR to include the needed changes to the option select component. If you have time can you give this a final once over to check I've not overlooked anything?

I haven't looked at documentation yet, I've just realised... I might do that in a separate PR for the sake of simplicity.

app/assets/javascripts/components/expander.js
```diff
@@ -69,17 +69,17 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
...
 
-    // GA4 Accordion tracking. Relies on the ga4-finder-tracker setting the index first, so we wrap this in a custom event.
-    window.addEventListener('ga4-filter-indexes-added', function () {
-      if (window.GOVUK.analyticsGa4) {
-        if (window.GOVUK.analyticsGa4.Ga4FinderTracker) {
-          window.GOVUK.analyticsGa4.Ga4FinderTracker.addFilterButtonTracking($button, this.$toggle.innerHTML)
-        }
+    var buttonAttributes = this.$module.getAttribute('data-button-data-attributes')
+    if (buttonAttributes) {
+      buttonAttributes = JSON.parse(buttonAttributes)
```
AshGDS, line 74: Does this need to be in a `try/catch`?

AshGDS (approved):
Looks good :+1:

app/assets/javascripts/components/expander.js
andysellick: Good shout, will match the same thing in the option select. Have added.

app/assets/javascripts/components/expander.js
```diff
@@ -69,17 +69,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none
...
-    window.addEventListener('ga4-filter-indexes-added', function () {
-      if (window.GOVUK.analyticsGa4) {
-        if (window.GOVUK.analyticsGa4.Ga4FinderTracker) {
-          window.GOVUK.analyticsGa4.Ga4FinderTracker.addFilterButtonTracking($button, this.$toggle.innerHTML)
+    var buttonAttributes = this.$module.getAttribute('data-button-data-attributes')
+    if (buttonAttributes) {
+      try {
+        buttonAttributes = JSON.parse(buttonAttributes)
+        for (var rawKey in buttonAttributes) {
+          var key = rawKey.replace(/_/i, '-').toLowerCase()
```
AshGDS, line 77: I just noticed something with this regex which also applies to the one in `govuk_publishing_components`

Would `/i` be irrelevant here as it's capturing a symbol rather than a character, so setting it to insensitive isn't necessary?

Also, should it have the `/g` flag instead so that it replaces every instance of an underscore in the key? With the current regex, I think it only works if the key only has one underscore?

AshGDS (commented):
Thanks for adding the `try/catch`. I may have noticed something regarding the regex

app/assets/javascripts/components/expander.js
andysellick: That's really helpful, thanks @AshGDS. Yes I think you're right, although for readability I've gone with the `replaceAll` function. I've also updated the tests to reflect this.

andysellick: Once I'm happy with this I'll make the same change on the option select component in the gem as well.

AshGDS: Thanks @andysellick - just to confirm, are we dropping IE11 support? If I recall correctly `replaceAll` doesn't work in IE11 (and it seems to not be working when I check in IE11 via BrowserStack). If we're not supporting IE11 anymore then happy to approve the change :+1:

andysellick: Ah thanks for spotting that, hadn't realised. Have reverted it back to `replace`.

andysellick:
@AshGDS going to hold off on merging this until tomorrow now. Feel free to take your time if you want to look over it again - it's already proved useful!

AshGDS (changes requested):
Thanks @andysellick :+1: Looks like there's one small thing to fix in the test

AshGDS (approved):
LGTM 👍

PR #3247, opened 2024-01-09
Only show parts for top three results

csutter:
There is a requirement for only the top three search results to show parts, which is currently handled by `search-api` only returning parts for the top three results. However, this should really be a frontend concern, and shouldn't be handled by the backend returning different result data depending on what the position of a result is across all results.

- Determine the `result_number` for each result in `ResultSetPresenter` and pass on to `SearchResultPresenter` (ideally this would use "rank" for better semantics but that's currently used by a debug parameter)
- Determine in `SearchResultPresenter` whether or not to include parts in the presentation depending on the result number

### DIFF
--- a/app/presenters/result_set_presenter.rb
+++ b/app/presenters/result_set_presenter.rb
@@ -87,6 +87,7 @@ private
       SearchResultPresenter.new(
         document:,
         rank: index + 1,
+        result_number: start_offset + index,
         metadata_presenter_class:,
         doc_count: documents.count,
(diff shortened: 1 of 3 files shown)


csutter:
See e.g. [this query](https://finder-frontend-pr-3247.herokuapp.com/search/all?keywords=service&content_purpose_supergroup%5B%5D=services&order=relevance) for parts being unaffected and continuing to show for top 3 results.

PR #3253, opened 2024-01-18
Add Prometheus metrics for search requests/duration

csutter:
- Add a `Metrics` convenience module for exporting Prometheus metrics
- Add a counter metric for number of search requests performed against a backend API, labelled by API version
- Add a histogram metric for duration of search requests performed against a backend API, labelled by API version
- Remove collection of legacy StatsD metrics for search request timing (these aren't really used for any useful purpose at the moment anyway, and we're moving away from StatsD across the GOV.UK estate)

---

Manually validated that the metrics are successfully exported:
image

### DIFF
--- a/app/lib/metrics.rb
+++ b/app/lib/metrics.rb
@@ -1,0 +1,45 @@
+module Metrics
+  CLIENT = PrometheusExporter::Client.default
+  COUNTERS = {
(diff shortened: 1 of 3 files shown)


sihugh (approved):
Lovely!

Sesame Street's Count von Count approving of this PR

PR #3255, opened 2024-01-19
Remove CSS reset

### LINKED ISSUE #2323: _reset.scss causes surprising behaviour for <strong>
Currently it's necessary to write: `<strong class="govuk-!-font-weight-bold">text</strong>` instead of just `<strong>` [[exampe](https://github.com/alphagov/finder-frontend/pull/2315/commits/ae42710fa2a40614c4485347de030c41c3d8f378)]. This is due to [a global reset rule](https://github.com/alphagov/finder-frontend/commit/ee71a6fce17cba1baf7d5812c7de491039b00adc#diff-f385dd830e5f8819e527a083c28c8ba8194e597c3859cf878548e7de39c90319R23), which means we have to add per-element overrides where the default style for the element conflicts with the imposed reset style.

Apparently we used to add `reset.scss` to projects in order to provide a foundation for cross-browser compatibility [[1](https://github.com/alphagov/finder-frontend/commit/ee71a6fce17cba1baf7d5812c7de491039b00adc#diff-f385dd830e5f8819e527a083c28c8ba8194e597c3859cf878548e7de39c90319)] [[2](https://github.com/alphagov/whitehall/commit/7718a08793aa89c50ef78e5a51de98ffc670ee1f#diff-4f0baac1d0bfdb5b0c85e03eccab791b49d4d25af181081feb593c4734b26da1)]. However, the approach seems to be unique to this app (and Whitehall), possibly because we now do it on a per-component basis [[example](https://github.com/alphagov/govuk_publishing_components/commit/19035c7df55db19d11373da00727e83801619aa6)]. It's unclear what effect removing the resets would have.

We should investigate where we're relying on these resets in this app, and apply them in exactly those places, so that the global resets on each element can be safely removed.

MartinJJones:
## What
- Remove `reset.scss`
- Set margins back to 0 for `app-c-expander__heading`

## Why
The CSS reset is no longer required in the finder-frontend application, removing the reset will ensure we have a consistent approach with other frontend rendering applications, it also fixes: https://github.com/alphagov/finder-frontend/issues/2323

[Trello card](https://trello.com/c/JHf43joM/2313-remove-eric-meyer-css-reset-from-finder-frontend-s), [Jira issue NAV-12196](https://gov-uk.atlassian.net/browse/NAV-12196)

## Visual Changes
The CSS reset in place was resetting the margins of the `dl` element in the [gem-c-metadata](https://components.publishing.service.gov.uk/component-guide/metadata/default/preview) component, after removing the reset, the default margins are used. This is consistent with other pages on GOV.UK where this component is used, for example - https://www.gov.uk/government/news/restoration-of-gloucester-flood-embankment-complete

| Before | After |
| --- | --- |
| metadata-list-before | metadata-list-after |

I could not see any other visual changes when testing and comparing the pages below:

- Email Sign up: https://www.gov.uk/administrative-appeals-tribunal-decisions/email-signup
- Search page: https://www.gov.uk/search
- https://www.gov.uk/search/all?order=updated-newest
- https://www.gov.uk/search/consultations
- https://www.gov.uk/search/statistics-announcements


davidtrussler (approved):
LGTM! 
Have compared the example pages listed and looks identical to me. Plus the `<dl>` change is better for consistency. 👍

PR #3256, opened 2024-01-19
[EMERGENCY ONLY] Disable v2 search AB test

csutter:
# ⚠️ Only merge this in case of emergency

**This commit removes usage of `search-api-v2` for the B variant of the "new search" AB test, which means all searches will go back to using v1.**

Merge this commit in case of an emergency where fixing the root issue isn't feasible.

### DIFF
--- a/app/lib/search/query.rb
+++ b/app/lib/search/query.rb
@@ -110,7 +110,7 @@ module Search
     end
 
     def use_v2_api?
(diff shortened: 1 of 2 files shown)


csutter:
The AB test has now ended and we can discard this PR.

#3298 is available to disable new search in case of emergency.

PR #3283, opened 2024-02-16
Accessibility fixes

JamesCGDS:
## What
This PR fixes a bug where focus is lost on the "Sort by" `select` element after using it to sort search results. This is because the DOM is updated with new elements when a user uses the "Sort by" filter. This includes inserting an entirely new "Sort by" `select` element and removing the previous `select` element which had received focus. I have applied a fix which retains the original `select` element and only updates the `option` elements within, therefore retaining the focus on the `select` element after it's been used.

This is achieved by using a regex on the new stringified HTML (`results.sort_options_markup`) that is received when `updateSortOptions()` is run following the application of filters to extract the `option` elements. The new `option` elements then replace the existing `option` elements in the "Sort by" `select` element. 

## Why
Identified as an issue as part of the WCAG audit of GOV.UK - [trello card](https://trello.com/c/fQ7cavwh/26-focus-gets-lost-when-sorting-search-results)

## Visual changes
None.

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -376,7 +376,7 @@
   }
 
   LiveSearch.prototype.removeSelectElement = function removeSelectElement (sortOptionsMarkup) {
-    var SELECT_TAG_REGEX = /<\s*\/?\s*select\b[^>]*>/g
+    var SELECT_TAG_REGEX = /<\/?select\b[^>]*>/g
     return sortOptionsMarkup.replace(SELECT_TAG_REGEX, '')
   }
 
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -359,7 +359,15 @@
 
   LiveSearch.prototype.updateSortOptions = function updateSortOptions (results, action) {
     if (action !== this.serializeState(this.state)) { return }
-    this.updateElement(this.$sortBlock, results.sort_options_markup)
+    var currentSortOptions = this.$sortBlock.querySelector('select')
+    var newSortOptions
+
+    if (results.sort_options_markup) {
+      var EXTRACT_OPTION_TAGS = /(?<=<select .+>)(.*)(?=<\/select>)/
+      newSortOptions = results.sort_options_markup.replace(newSortOptions, EXTRACT_OPTION_TAGS)
+    }
+
+    this.updateElement(currentSortOptions, newSortOptions)
     this.bindSortElements()
   }
 
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -365,12 +365,19 @@
     var newSortOptions
 
     if (results.sort_options_markup) {
-      var EXTRACT_OPTION_TAGS = /(?<=<select .+>)(.*)(?=<\/select>)/
-      newSortOptions = results.sort_options_markup.replace(newSortOptions, EXTRACT_OPTION_TAGS)
+      // The select element is removed from results.sort_options_markup in order to retain the original select element. Only the option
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/live_search.js
```diff
@@ -359,7 +361,15 @@
 
   LiveSearch.prototype.updateSortOptions = function updateSortOptions (results, action) {
     if (action !== this.serializeState(this.state)) { return }
-    this.updateElement(this.$sortBlock, results.sort_options_markup)
+    var currentSortOptions = this.$sortBlock.querySelector('select')
+    var newSortOptions
+
+    if (results.sort_options_markup) {
+      var EXTRACT_OPTION_TAGS = /(?<=<select .+>)(.*)(?=<\/select>)/
```
AshGDS, line 368: I don't think positive lookaheads work in IE11, if you're using regex to keep compatibility with this browser. https://caniuse.com/?search=exp - the JS for the page seems to break in IE11 on Browserstack.

Instead, I think you could remove `<select>` tags from the new markup string, and then replace the `innerHTML` of the existing `<select>` with the new markup string?

e.g. https://regex101.com/r/hQ5KdJ/1

```JS
var regexThatGrabsSelects = /<\s*\/?\s*select\b[^>]*>/g
newMarkup.replace(regex, "")
existingSelect.innerHTML = newMarkup
```

Also, I think there should be a test written that ensures the regex is working

AshGDS (changes requested):
Nice work, though it doesn't seem to be working in IE11

app/assets/javascripts/live_search.js
JamesCGDS: Excellent points/suggestions 👍 I've refactored the code a bit and put the regex stuff in a function so it's a bit easier to test. Thanks @AshGDS

JamesCGDS:
Thanks for the review @AshGDS - should be ready for a re-review now

app/assets/javascripts/live_search.js
```diff
@@ -359,10 +361,25 @@
...
+      // results.
+      newSortOptions = this.removeSelectElement(results.sort_options_markup)
+    }
+
+    this.updateElement(currentSortOptions, newSortOptions)
     this.bindSortElements()
   }
 
+  LiveSearch.prototype.removeSelectElement = function removeSelectElement (sortOptionsMarkup) {
+    var SELECT_TAG_REGEX = /<\s*\/?\s*select\b[^>]*>/g
```
AshGDS, line 379: I think we can actually simplify the regex a bit as it seems to be looking for spaces at the start of the `<select>` tag when it doesn't need to - my fault as I shared this regex initially. We may just be able to do:

`<\/?select\b[^>]*>` https://regex101.com/r/l4EPY2/1

app/assets/javascripts/live_search.js
```diff
@@ -359,10 +361,25 @@
...
+    var newSortOptions
+
+    if (results.sort_options_markup) {
+      // The select element is removed from results.sort_options_markup in order to retain the original select element. Only the option
+      // elements are updated with new markup. This is to prevent focus from being lost on the select element after the user sorts their
+      // results.
+      newSortOptions = this.removeSelectElement(results.sort_options_markup)
+    }
+
+    this.updateElement(currentSortOptions, newSortOptions)
```
AshGDS, line 374: Shouldn't all this code be inside the if statement on line 367? As all this code seems to only be relevant if `results.sort_options_markup` exists?

AshGDS (commented):
Looks good - just a couple more things to think about 👍

app/assets/javascripts/live_search.js
JamesCGDS: Cool, no worries - my fault also for not checking closely. All done 👍

AshGDS (approved):
LGTM 👍

PR #3295, opened 2024-02-27
Use v2 for all site search queries

csutter:
This switches the `/search/all` ("site search") finder over to use `search-api-v2` exclusively now that our AB test has concluded with a positive outcome.

- Instead of checking for AB test variant, check for site search finder base path to establish whether to use v2
- Add `use_v1` query parameter along the lines of the existing `use_v2` to force use of `search-api` in circumstances where `search-api-v2` would normally be used
- Fix stubbing of API endpoints in tests

### DIFF
--- a/app/lib/search/query.rb
+++ b/app/lib/search/query.rb
@@ -2,6 +2,8 @@
 # search results from rummager.
 module Search
   class Query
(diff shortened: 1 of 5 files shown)


leenagupte (approved):
👍 🚀 
So many stubs to update! I think there's job in the future for someone to rationalise all of that 😬

PR #3298, opened 2024-02-28
BREAK GLASS: Revert to always using Search API v1

csutter:
This is an emergency-only commit that can be merged in order to revert
all searches (in particular site search) back to exclusively using the
v1 `search-api`, for example if a major incident occurs in the short
term after launching.

### DIFF
--- a/spec/controllers/finders_controller_spec.rb
+++ b/spec/controllers/finders_controller_spec.rb
@@ -425,6 +425,7 @@ describe FindersController, type: :controller do
 
   describe "with legacy query parameters for publications" do
     before do
+      search_api_request(search_api_app: "search-api")
       search_api_request(search_api_app: "search-api-v2")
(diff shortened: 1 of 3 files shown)


spec/controllers/finders_controller_spec.rb
```diff
@@ -425,7 +425,7 @@
 
   describe "with legacy query parameters for publications" do
     before do
-      search_api_request
+      search_api_request(search_api_app: "search-api-v2")
```
leenagupte, line 428: I think the tests are failing because the stub for v1 `search_api_request` hasn't been added back.
I'd leave both in for the break glass to make switching easier

leenagupte (approved):

kevindew:
@csutter can we close this now? We get a reminder to review it each day

csutter:
Closing as this has been live for several months without incident now.

PR #3318, opened 2024-03-26
Force start Prometheus Exporter

leenagupte:
# What?
Suppresses Prometheus Exporter errors in development and test environments

## Before

Screenshot 2024-03-26 at 13 21 50

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -70,3 +70,6 @@ Rails.application.configure do
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
 end
+
+# TODO: remove this workaround once GovukPrometheusExporter initialisation is fixed in govuk_app_config.
+ENV["GOVUK_PROMETHEUS_EXPORTER"] = "force"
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM (Based on https://gds.slack.com/archives/CAB4Q3QBW/p1711386477944039)

sengi (commented):
I would definitely put a TODO on this because it's a temporary workaround at best and I think it's likely to lead to further confusion.

sengi:
I'm keen to fix the underlying problem in govuk_app_config, so something like `TODO: remove this workaround once GovukPrometheusExporter initialisation is fixed in govuk_app_config.`

sengi:
Also worth understanding that this doesn't really suppress the errors; it avoids them in the first place by starting that server that's supposed be listening on `localhost:9394`. So this is more accurately "force starting Prometheus exporter" rather than "suppress errors".

There are two underlying issues here: firstly govuk_app_config has some confusing logic about whether it starts the "exporter" and secondly finder-frontend seems to have grown some instrumentation code that behaves poorly when the exporter isn't reachable.

sengi (approved):
Just to be clear I'm not blocking this — but I do highly recommend adding the TODO so that others don't assume this workaround is normal or expected. It'd be unfortunate if this spreads to other apps.

leenagupte:
> Just to be clear I'm not blocking this — but I do highly recommend adding the TODO so that others don't assume this workaround is normal or expected. It'd be unfortunate if this spreads to other apps.

Thanks Chris, I'll add the comment and update the commit message.

PR #3331, opened 2024-04-15
Use `search-api` v1 for non-keyword searches

csutter:
We're moving away from our initial intent to use Search API v2 exclusively for all queries to "site search" (aka `/search/all`) for the short to medium term.

This is for two reasons:
- Our underlying search engine technology isn't designed (yet) to handle non-relevance based queries well (resulting in e.g. attempting to view _all_ documents matching certain filter criteria sorted by updated missing many documents)
- We get a lot of queries without keywords from automated processes, which costs money in a usage-based charging model

In the longer term, this should be resolved by a more flexible content API that allows retrieving doucments by certain criteria. Until then, we can continue to use v1 as it is still around and we know it's able to serve non-relevance based search scenarios well.

### DIFF
--- a/app/lib/search/query.rb
+++ b/app/lib/search/query.rb
@@ -116,6 +116,10 @@ module Search
       return true if ActiveModel::Type::Boolean.new.cast(filter_params["use_v2"])
       return false if filter_params["world_locations"].present?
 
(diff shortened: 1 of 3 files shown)


leenagupte (approved):
😢

PR #3334, opened 2024-04-23
Improve the spacing and alignment of search filters on mobile

matthillco:
## What

Improve the spacing and alignment of elements within the search filters component in the mobile view.
[Trello](https://trello.com/c/65zYsdtH/275-type-scale-finder-frontend-improve-spacing-and-alignment-in-search-filters-on-mobile), [Jira issue PNP-8551](https://gov-uk.atlassian.net/browse/PNP-8551)

## Why

This component is a bit messy with several elements not lining up and the submit button falls outside the container. With the forthcoming changes to the font sizes, the spacing looks even worse and must be fixed. The improvements will help make this more visually consistent.

## How

* Extra top padding added inside `facets__content` to ensure that all filter items have the same apparent vertical height
* Expander headings now use flexbox to align the icon and button text. This is more robust and ensures that the vertical  centring of the icon and text are respected for different browser zoom settings (for example when using the 'Zoom Text Only' option in Firefox).
* Consistent spacing around second bold results text
* Facets form now accounts for 2px shadow on the button

Note that the same tweak to the expander icons will be applied in the Options Select component in the gem, see this PR: https://github.com/alphagov/govuk_publishing_components/pull/4256

## BEFORE / AFTER

image

### DIFF
--- a/app/assets/stylesheets/components/_expander.scss
+++ b/app/assets/stylesheets/components/_expander.scss
@@ -22,7 +22,10 @@
 .govuk-frontend-supported {
   .app-c-expander__heading {
     position: relative;
-    padding: 10px 8px 5px 43px;
+    display: flex;
+    flex-direction: row-reverse;
+    justify-content: start;
+    align-items: center;
     margin: 0;
   }
 
@@ -45,11 +48,9 @@
 
 .app-c-expander__icon {
   display: none;
-  position: absolute;
-  top: 0;
-  left: 9px;
   width: 30px;
   height: 40px;
+  margin-inline: 9px 4px;
   fill: govuk-colour("black");
 }
 
--- a/app/assets/stylesheets/components/_mobile-filters.scss
+++ b/app/assets/stylesheets/components/_mobile-filters.scss
@@ -74,7 +74,8 @@
     .facets__box {
       border-left: 5px solid #b1b4b6;
       padding-left: govuk-spacing(2);
-      margin-bottom: govuk-spacing(3);
+      padding-bottom: 2px;
+      margin-bottom: govuk-spacing(4);
     }
 
     .facets__header {
@@ -86,10 +87,14 @@
       }
     }
 
+    .facets__content {
+      padding-top: govuk-spacing(2);
+    }
+
     .facets__footer {
       display: block;
       margin-top: govuk_spacing(3);
-      margin-bottom: govuk_spacing(3);
+      margin-bottom: 0;
     }
 
     .facets__return-link {
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -290,8 +290,11 @@ mark {
 }
 
 .result-info {
-  margin: 0 0 govuk-spacing(4) 0;
+  @include govuk-media-query($until: tablet) {
+    padding-bottom: govuk-spacing(2);
+  }
   border-bottom: solid 1px govuk-colour("black");
+  margin: 0 0 govuk-spacing(4) 0;
 }
 
 // NOTE: Reset from element level styles inherited from the deprecated `govuk-template`.
@@ -325,10 +328,7 @@ mark {
   cursor: pointer;
   -webkit-appearance: none;
   text-decoration: underline;
-
-  // we need consistent font-size across breakpoints
-  // so we cannot use govuk-font mixin
-  font-size: 16px;
+  @include govuk-font($size: 19);
   @include govuk-link-common;
   @include govuk-link-style-default;
 


app/assets/stylesheets/components/_mobile-filters.scss
```diff
@@ -109,6 +114,28 @@
     .facets__tags-block {
       display: block;
     }
+
+    // Temporary spacing fix because first item is shorter than the others
```
andysellick, line 118: It's unclear what temporary means here - were these changes originally meant to be deployed before the v5 upgrade, and can be removed once the typography changes are deployed?

app/assets/stylesheets/components/_mobile-filters.scss
```diff
@@ -109,6 +114,28 @@
...
+    }
+
+  }
+}
+
+// The vertical position of other expander icons was fixed in _expander.scss
+// as it's app specifc. However, we also have gem expanders mixed in here
+// so these need the same override, for all breakpoints:
+.js-enabled {
+  .facets__content .gem-c-option-select__icon {
```
andysellick, line 137: This should be fixed in the option select component in the gem.

andysellick (changes requested):
Spotted some things that need to be addressed before merging, happy to discuss.

app/assets/stylesheets/components/_mobile-filters.scss
```diff
@@ -84,12 +85,16 @@
         display: grid;
         grid-auto-flow: column;
       }
+
+      .gem-c-heading {
+        @include govuk-font(19);
```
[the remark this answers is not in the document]
matthillco: This rule is no longer needed with the new typography updates: it's now deleted.

app/assets/stylesheets/components/_mobile-filters.scss
```diff
@@ -109,6 +114,28 @@
     .facets__tags-block {
       display: block;
     }
+
+    // Temporary spacing fix because first item is shorter than the others
+    .facets__content > .app-c-expander:first-child {
+      padding-top: govuk-spacing(2);
```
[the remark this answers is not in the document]
matthillco: I've now moved the solution into the `facets__content` element which is a better place to have the missing padding. This is now all self contained.

app/assets/stylesheets/components/_mobile-filters.scss
matthillco: Yes, this has now been changed/removed, it's no longer needed in the previous form.

app/assets/stylesheets/components/_mobile-filters.scss
matthillco: This has been updated in the gem as requested, see here: https://github.com/alphagov/govuk_publishing_components/pull/4256

app/assets/stylesheets/components/_expander.scss
```diff
@@ -46,7 +46,7 @@
 .app-c-expander__icon {
   display: none;
   position: absolute;
-  top: 0;
+  top: 3px;
```
andysellick, line 49: Whilst this looks correct at a regular size, if you increase the font size it starts to look misaligned.

Screenshot 2024-09-25 at 10 41 02

(taken on Firefox with 'zoom text only' enabled)

I've tried setting the height to 100% rather than 40px, which seems to improve things:

Screenshot 2024-09-25 at 10 40 51

Outline added just for emphasis. Obviously would need updating in the option select component as well. Also might want to play with the `top: 3px` as well - this feels like a bit of a magic number and I'd prefer to avoid that.

app/assets/stylesheets/components/_mobile-filters.scss
```diff
@@ -74,7 +74,8 @@
     .facets__box {
       border-left: 5px solid #b1b4b6;
       padding-left: govuk-spacing(2);
-      margin-bottom: govuk-spacing(3);
+      padding-bottom: 2px;
+      margin-bottom: govuk-spacing(4);
```
andysellick, line 78: This feels a bit hacky but it seems like you're having to jump through this hoop due to an oddity with the button component - it's got a margin bottom of zero but a 2px box shadow, which means it always hangs down a little. That's an oddity we've introduced (the Design System button appears to have a default bottom margin of 32px, which would be the 2px plus the DS spacing scale). It might be worth investigating why our button component is like that? Personally I'd be tempted to set the default margin bottom to 2px so it doesn't hang down and we don't have this problem.

app/assets/stylesheets/components/_mobile-filters.scss
```diff
@@ -86,10 +87,14 @@
       }
     }
 
+    .facets__content {
+      padding-top: govuk-spacing(2);
```
andysellick, line 91: I think there's an alternative solution here. Each of the child elements (either an expander or an option select) has a slightly odd thing where they've got a padding and margin bottom of 10px. That means each one is reliant on the one above to position it correctly within the dividing lines, with the addition of this 10px at the very top to position the first one.

This feels a bit clunky to me - I think it would be cleaner if there was no margin and all these child elements had 10px of padding top and bottom. Thoughts?

app/assets/stylesheets/components/_expander.scss
matthillco: Yep, there's a lot of improvements that could be made here and I mentioned a full refactor in the other PR for the same tweak in the gem, citing magic numbers as part of the problem.

The best solution is to use flexbox and avoid all the outdated absolute positioning entirely, but I've had mixed reactions to using flexbox in other work I've done here, which I don't really understand (flexbox has been stable for years). I'll explore a flexbox solution instead, but I do hope it won't be squashed in preference to an outdated approach.

app/assets/stylesheets/components/_mobile-filters.scss
matthillco: Yep, it's hacky and I'm not too happy about it, but it seems the simplest option give these issues that are causing it:
- box shadows aren't computed as part of an elements size, so they don't take up any space. In this case, they fall outside the container, making the button appear misaligned with the border on the left. 
- govuk buttons have margins, but gem buttons don't, they're reset to 0 so the button is flush at the end of the container and thus the shadow falls outside.

The 2px of padding that I added is to force a space for the shadow, making the container just long enough so that the button appears to line up fully with the left hand border. 

Another solution would be to fiddle with a specific margin on this specific button that accounts for all these variables, but that seems like an even worse hack.

We could also remove the left hand grey border and padding from the `facets__box`. This border was clearly added to wrap everything and build a relationship with the Filter toggle at the top, so it is useful. The question then is whether it's more important to have this useful indicator or to have the button shadow line up with it. I know which I'd choose for the sake of simplicity, but it was a difference of opinion on that which made me do this hack in the first place.

Personally I'd solve the whole thing by rethinking the use of box-shadows to simulate borders,  but that's not my call.

This is a lot of thinking for such a tiny issue! I'm therefore inclined to leave this 'fix' alone.  🫣

app/assets/stylesheets/components/_mobile-filters.scss
matthillco: I would agree, however it's messy because some search results filters have weird things like this `facets-container` that contains a `govuk-form-group` as the first item:

image

If we used your suggestion, we'd still need to add an extra 10px of top padding to the `facets-container` otherwise the first radio button is flush with the first border. Given that my solution solves both problems, albeit in a slightly hacky way, I think it's the better option right now.

In all honesty this entire module needs a complete overhaul: it is incredibly confusing and comprises different components that don't play well together. Another one for the refactoring backlog...?

app/assets/stylesheets/components/_expander.scss
andysellick: Sure, either flexbox or the one line change I suggested is fine.

app/assets/stylesheets/components/_mobile-filters.scss
andysellick: Okay, let's leave as is for now. I've asked the Design System team about this bottom box shadow, interested to see what they say.

app/assets/stylesheets/components/_mobile-filters.scss
andysellick: Personally I'd go with that as a solution but as this whole section is going to be rebuilt shortly I'm happy to leave as-is.

app/assets/stylesheets/components/_expander.scss
matthillco: @andysellick I've replaced the outdated absolute positioning solution with flex-box, avoiding the need for most magic numbers. I've tested it across different browsers and different zoom levels (and the Firefox 'Zoom text only' setting) and it's robust. 

If you're happy with it, I'll roll out the same change to the [option selects in the other PR](https://github.com/alphagov/govuk_publishing_components/pull/4256) as they use the same layout rules and so can receive this same fix.

andysellick (approved):

app/assets/stylesheets/components/_expander.scss
andysellick: Sure go ahead.

PR #3376, opened 2024-06-10
Upgrade to govuk frontend 5.1.0

MartinJJones:
## What

- Move components that rely on govuk-frontend modules to separate `es6-components.js` file
- Remove Sass variables
- Update polyfill file paths

### Move components that rely on govuk-frontend modules to seperate `es6-components.js` file

In the event that a browser below the target for `govuk-frontend` loads a page with JS on it, attempting to parse the JS from `govuk-frontend` will cause an error. To avoid this from happening, JS that contains `govuk-frontend` JS has been moved to seperate file which will be loaded in a script tag with `type="module"`. This will prevent the JS from being parsed and so prevent the error

### Remove Sass variables

- The `$legacy` attribute in `govuk-colour` has been deprecated and using it will have no effect (other than generating warnings on pre-compilation)
- The Sass variables below have been removed, using them will have no effect:
  - `$govuk-compatibility-govuktemplate`
  - `$govuk-use-legacy-palette`
  - `$govuk-typography-use-rem` 
- The Sass variable `$govuk-new-link-styles` is now set to `true` by default, so it does not need to be set 

### Update polyfill file paths

Browser polyfills are no longer included in v5 of govuk-frontend, any Polyfills provided in v4 of govuk-frontend will be included in the new version of the govuk-publishing-components gem.

The intention is for this PR to include the bump to the govuk-frontend v5 version of the govuk-publishing-components gem, this will also fix the failing tests relating to the missing polyfill.

[Trello](https://trello.com/c/JgseDbQj/2514-upgrade-finder-frontend-to-run-on-v51-of-govuk-frontend), [Jira issue NAV-12430](https://gov-uk.atlassian.net/browse/NAV-12430)

---

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -189,7 +189,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (39.2.5)
+    govuk_publishing_components (40.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,6 +1,7 @@
 //= link_tree ../images
 //= link application.js
 //= link test-dependencies.js
+//= link es6-components.js
 //= link live_search.js
 
 //= link_tree ../builds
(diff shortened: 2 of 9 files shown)


andysellick (approved):
Changes look good. I note that the tests are failing, but only because this branch doesn't have the v5 upgrade as part of it yet, so that's fine.

PR #3405, opened 2024-07-09
Reintroduce GA4 pageview tracking code

AshGDS:
## What / Why
- Reintroduce some code that was accidentally removed during UA removal
- This caused updating sort options to not trigger a new pageview event
- This is because I removed `trackSearch`, which calls `trackingInit` (which I also removed), and `trackingInit` called`this.startEnhancedEcommerceTracking()` which was needed to reinitialise the GA4 Pageview tracker when the sort options update.  To see the code that was accidentally removed, see [here](https://github.com/alphagov/finder-frontend/pull/3396/files#diff-fe1355439297aa459f871621b4f95627ea330655c7c6eb08ef10bda7ab66c0feL262) and [here](https://github.com/alphagov/finder-frontend/pull/3396/files#diff-fe1355439297aa459f871621b4f95627ea330655c7c6eb08ef10bda7ab66c0feL450)
- Therefore I've added the necessary code back - I'm now just calling one function, `reinitialiseGa4Tracking`(as we don't need the equivalent of `trackSearch` anymore.)
- I've also namespaced those functions with Ga4 just so they aren't removed by accident in the future
- Also updates a couple cookie checks to `usage` instead of `settings`

### DIFF
--- a/app/assets/javascripts/live_search.js
+++ b/app/assets/javascripts/live_search.js
@@ -116,11 +116,16 @@
     }
   }
 
+  LiveSearch.prototype.trackingInit = function trackingInit () {
+    GOVUK.modules.start(this.$resultsWrapper)
+    this.startEnhancedEcommerceTracking()
+  }
+
   LiveSearch.prototype.Ga4EcommerceTracking = function (referrer) {
     if (GOVUK.analyticsGa4 && GOVUK.analyticsGa4.Ga4EcommerceTracker) {
       var consentCookie = GOVUK.getConsentCookie()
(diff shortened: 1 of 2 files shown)


AshGDS:
@andysellick The cookie being `.settings` by accident has surfaced another bug - if the user had accepted `settings` cookie but not `usage` cookies, the JS crashes when they try and search again as GA4 tracking tries to fire, but it doesn't exist

andysellick (approved):
Nice work 🚀

PR #3428, opened 2024-08-20
Move breadcrumbs out of the <main> landmark to fix a WCAG fail

hannalaakso:
## What
Move breadcrumbs and phase banner out of the `<main>` landmark, retaining the existing visual design. This change fixes a WCAG fail and skips the breadcrumbs and phase banner when the 'skip to main content' link is activated (which it should do as breadcrumbs are navigational content). 

This PR:
- Moves the phase banner and breadcrumbs out of `<main>`into a new partial that the finders layout now yields before rendering `<main>`. [The GOV.UK Design System documentation](https://design-system.service.gov.uk/styles/page-template/#options) calls this area of the page before the main container `before_content` so the name of the partial is aligned with that naming convention. (We yield the template in finder_layout rather than just render it, because the email alert subscriptions also use finder_layout and those pages don't include the phase banner or breadcrumbs.)
- Calls the `/before_content` partial inside `/show` (instead of `/show_header` which previously housed the breadcrumbs and phase banner) because the breadcrumbs and phase banner are now outside the main content and no longer inside the header content.
-  Supports an inverse design with a full-width blue banner on topic pages and simplifies the logic of enabling the design
- Removes redundant breadcrumbs call which was tech debt from the Brexit checkers

## Why
The breadcrumbs were within the `<main>` landmark which is also the target of the ‘skip to main content' link. There was no mechanism to allow users to bypass the breadcrumb links. Typically users should be able to use the ‘skip to main content’ link to bypass repeated blocks of content, like navigation links and breadcrumbs, that are repeated on multiple pages, and navigate directly to the main content of the page.

The breadcrumbs should not be in the main landmark and the 'skip to content' link should skip over them. This would also comply with [Design System guidance](https://design-system.service.gov.uk/components/breadcrumbs/#how-it-works).

WCAG reference: 2.4.1 Bypass Blocks (Level A)

## How the change was tested

The change was confirmed to allow the 'skip to main content' link to skip over the breadcrumbs. There were no visual changes to the page.

### Before

Screenshot 2024-09-10 at 10 12 04

### After

Screenshot 2024-09-10 at 10 12 04

### Before

Screenshot 2024-09-10 at 10 12 13

### After

Screenshot 2024-09-10 at 10 12 23

### Before

Screenshot 2024-09-10 at 10 25 49

### After

Screenshot 2024-09-10 at 10 25 58

## Pages checked

- /search/guidance-and-regulation
- /search/guidance-and-regulation?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- /aaib-reports
- /drug-device-alerts
- /government/case-studies
- /government/people

#### Windows 11	
✅ Edge  (latest versions)
✅ Google Chrome  (latest versions)
✅ IE11 (Windows 8)

#### macOS 14.5	
✅  Safari 17
✅  Google Chrome (latest version)

#### iOS	
✅  Safari on iPhone 14

#### Android	
✅  Chrome on Samsung Galaxy S23

### Screen readers
✅  JAWS 2023 with Chrome 
✅  NVDA 2024 with Firefox

Fixes https://trello.com/c/OVhV3mnw/2772-fix-breadcrumbs-wcag-fails-in-finder-frontend-m-l, [Jira issue NAV-15179](https://gov-uk.atlassian.net/browse/NAV-15179)#comment-66c488bd6d425a3f347b49bf

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -365,3 +365,10 @@ mark {
   padding-left: 13px;
   padding-right: 13px;
 }
+
+.app-before-content--inverse {
+  background-color: govuk-colour("blue");
+  // Stretch background colour over inner margins
+  overflow: hidden;
+  padding-top: govuk-spacing(3);
+}
--- a/app/helpers/topic_finder_helper.rb
+++ b/app/helpers/topic_finder_helper.rb
@@ -6,4 +6,8 @@ module TopicFinderHelper
   def topic_finder_parent(filter_params)
     Services.registries.all["full_topic_taxonomy"][filter_params["topic"]]
   end
+
+  def is_inverse?
+    topic_finder?(filter_params) && !content_item.all_content_finder?
+  end
 end
--- a/app/views/finders/_before_content.html.erb
+++ b/app/views/finders/_before_content.html.erb
@@ -1,0 +1,26 @@
+<% content_for :before_content do %>
+    <div class="app-before-content">
+        <div class="govuk-width-container">
+            <% if content_item.show_phase_banner? %>
+                <%= render 'govuk_publishing_components/components/phase_banner', {
+                    phase: content_item.phase, 
+                    message: sanitize(content_item.phase_message), 
+                    inverse: inverse
+                } %>
+            <% end %>
+
+            <% if @breadcrumbs.breadcrumbs %>
+                <%= render 'govuk_publishing_components/components/breadcrumbs', {
+                    breadcrumbs: @breadcrumbs.breadcrumbs,
+                    inverse: inverse,
+                    collapse_on_mobile: true
+                } %>
+            <% else %>
+                <%= render 'govuk_publishing_components/components/contextual_breadcrumbs', {     
+                    content_item: content_item.as_hash,
+                    inverse: inverse
+                } %>
+            <% end %>
+        </div>
+    </div>
+<% end %>
--- a/app/views/finders/_before_content.html.erb
+++ b/app/views/finders/_before_content.html.erb
@@ -1,24 +1,24 @@
 <% content_for :before_content do %>
-    <div class="app-before-content">
+    <div class="app-before-content<%= " app-before-content--inverse" if is_inverse? %>">
         <div class="govuk-width-container">
             <% if content_item.show_phase_banner? %>
                 <%= render 'govuk_publishing_components/components/phase_banner', {
                     phase: content_item.phase, 
                     message: sanitize(content_item.phase_message), 
-                    inverse: inverse
+                    inverse: is_inverse?
                 } %>
             <% end %>
 
             <% if @breadcrumbs.breadcrumbs %>
                 <%= render 'govuk_publishing_components/components/breadcrumbs', {
                     breadcrumbs: @breadcrumbs.breadcrumbs,
-                    inverse: inverse,
+                    inverse: is_inverse?,
                     collapse_on_mobile: true
                 } %>
             <% else %>
                 <%= render 'govuk_publishing_components/components/contextual_breadcrumbs', {     
                     content_item: content_item.as_hash,
-                    inverse: inverse
+                    inverse: is_inverse?
                 } %>
             <% end %>
         </div>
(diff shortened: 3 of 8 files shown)


features/breadcrumbs.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Breadcrumbs
+  In order to use the skip link,
+  As a user
+  I want the breadcrumbs to be outside the main content
+
+  Scenario: When there are breadcrumbs on the page
+    Given the page exists
+    When I visit the page
```
unoduetre, line 8: When I visit /search

features/breadcrumbs.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Breadcrumbs
+  In order to use the skip link,
+  As a user
+  I want the breadcrumbs to be outside the main content
+
+  Scenario: When there are breadcrumbs on the page
+    Given the page exists
```
unoduetre, line 7: Given there exists (something here) content item

features/step_definitions/breadcrumbs_steps.rb
```diff
@@ -0,0 +1,14 @@
+Given(/^the page exists$/) do
+  stub_content_store_has_item("/search", schema: "special_route")
+end
+
+When(/^I visit the page$/) do
```
unoduetre, line 5:
```ruby
When(/^I visit {string}$/) do |path|
  visit path
end
```

features/step_definitions/breadcrumbs_steps.rb
```diff
@@ -0,0 +1,14 @@
+Given(/^the page exists$/) do
```
unoduetre, line 1: Given(/^there exists (something here) content item$/) do

features/step_definitions/breadcrumbs_steps.rb
```diff
@@ -0,0 +1,14 @@
...
+  stub_content_store_has_item("/search", schema: "special_route")
```
unoduetre, line 2: This should work for the search page, but you mentioned you wanted to use a different page to check it, so you can check one with a banner.

Which page would you like to use?

unoduetre: EDIT: Or instead of doing all of that you can simply reuse an existing step, please, see a comment above.

features/breadcrumbs.feature
unoduetre: Or instead of doing all of that you can reuse an existing step:

    When I view the research and statistics finder with a topic param set

features/step_definitions/breadcrumbs_steps.rb
unoduetre: Or instead of doing all of that you can simply reuse an existing step, please, see a comment above.

features/breadcrumbs.feature
unoduetre: Or instead of doing all of that you can simply reuse an existing step, please, see a comment above.

features/step_definitions/breadcrumbs_steps.rb
unoduetre: Or instead of doing all of that you can simply reuse an existing step, please, see a comment above.

features/breadcrumbs.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Breadcrumbs
```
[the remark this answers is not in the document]
hannalaakso: Thanks Jess, I really like this approach and have implemented it ✨

app/assets/stylesheets/finder_frontend.scss
```diff
@@ -365,3 +365,9 @@ mark {
   padding-left: 13px;
   padding-right: 13px;
 }
+
+.app-before-content--inverse {
+  background-color: govuk-colour("blue");
+  padding-top: govuk-spacing(3);
+  overflow: hidden;
```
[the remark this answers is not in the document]
hannalaakso: Thanks Martin. I'm not sure if ` @include govuk-clearfix;` would have the desired effect, could you double check on this? I'm seeing the background colour not bleeding full height on a tablet: 
Screenshot 2024-09-11 at 17 01 13

Alternative to this fix would be to amend the contextual breacrumbs component so that classes could be passed to the breadcrumbs component within which could be used to amend the bottom margin of the component.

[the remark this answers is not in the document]
MartinJJones: Thanks Hanna, you're right, definitely an issue on tablet.

The fix looks good to me, as you say, ideally we would be able to customise the spacing in the breadcrumbs component which could be a nice improvement to make in the future as a separate piece of work.

MartinJJones (approved):
Nice work on fixing this, having the separate partial template is much cleaner as well 👍

KludgeKML (approved):
Nice change. Thanks!

PR #3438, opened 2024-09-06
New UI for all content finder

csutter:
> ⚠️ Not ready for review, draft PR to get a review app

As part of our efforts to improve the site search experience on GOV.UK, we have designed a new UI for the "all content" finder (`/search/all`).

This PR starts to implement this new design, behind a `ENABLE_NEW_SITE_SEARCH_UI` feature flag for the time being.


app/controllers/finders_controller.rb
```diff
@@ -13,8 +13,14 @@ def show
         raise UnsupportedContentItem unless content_item.is_finder?
 
         if legacy_params_present?
-          transform_legacy_announcement_params_and_redirect if content_item.base_path == "/search/news-and-communications"
-          transform_legacy_publication_params_and_redirect if content_item.base_path == "/search/all"
+          case content_item.base_path
```
leenagupte, line 16: This is nicer than the original inline "ifs".n ✨

features/site_search.feature
```diff
@@ -3,33 +3,29 @@ Feature: Site search
   As a user
   I want to be able to search a list of documents
 
-  Scenario: When no search terms are entered
+  Background:
     Given the search page exists
+    And the all content finder exists
+    And the new all content finder UI is disabled
```
leenagupte, line 9: Nice DRY-up ✨

spec/controllers/finders_controller_spec.rb
```diff
@@ -268,6 +268,14 @@
         expect(response).to render_template("finders/show")
       end
 
+      it "correctly renders the new template if the param is given" do
+        get :show, params: {
+          slug: "search/all", keywords: "hello", enable_new_all_content_finder_ui: "true"
```
leenagupte, line 273: Given that the ENV VAR is checked first, you'll to unset the stubbed constant first, to be sure the param is being tested.

i.e. remove this: 
``` ruby
stub_const("ENV", ENV.to_hash.merge("ENABLE_NEW_SITE_SEARCH_UI" => "true"))
```

app/controllers/finders_controller.rb
```diff
@@ -83,7 +83,7 @@ def show_template
   end
 
   def enable_new_all_content_finder_ui?
-    ActiveModel::Type::Boolean.new.cast(ENV["ENABLE_NEW_SITE_SEARCH_UI"]) ||
+    ActiveModel::Type::Boolean.new.cast(ENV["ENABLE_NEW_ALL_CONTENT_FINDER_UI"]) ||
```
leenagupte, line 86: Minor: This should probably be fixed up to previous commits.

leenagupte (commented):
I've left a couple of inline comments. The one that really needs looking over is the param test. Otherwise I think it looks good.

It'd be good for some with a keener frontend-eye to look over it. I think it's just reusing parts from existing views and isn't actually implementing anything new yet...

spec/controllers/finders_controller_spec.rb
csutter: Ah – this is outside the describe block that stubs the env var to `true`, but that raises a different point in that this set of contexts is hard to read I need to rearrange this whole block a bit better!

leenagupte (approved):
Looks good 👍 

Thanks for confirming elsewhere that the new view isn't adding anything extra in this PR.

PR #3441, opened 2024-09-10
Add spelling suggestion to new all content finder UI

csutter:
- Add the existing spelling suggestion partial (as we don't need to do anything special compared to before)
- Update Cucumber step to check for suggestion link as regex (because the other feature using it additionally has a specific sort order set)

image

### DIFF
--- a/app/views/finders/show_all_content_finder.html.erb
+++ b/app/views/finders/show_all_content_finder.html.erb
@@ -40,6 +40,10 @@
           } %>
         </div>
 
+        <div class="spelling-suggestions">
+          <%= render 'spelling_suggestion' %>
+        </div>
+
         <% if result_set_presenter.total_count.positive? %>
           <%= render "govuk_publishing_components/components/document_list", {
             disable_ga4: true,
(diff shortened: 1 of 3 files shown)


app/views/finders/show_all_content_finder.html.erb
```diff
@@ -40,6 +40,10 @@
           } %>
         </div>
 
+        <div class="spelling-suggestions">
```
leenagupte, line 43: `js-spelling-suggestions` class isn't needed here? I'm not able to check if the JS is being picked up regardless.

leenagupte (commented):
I have one inline comment about whether `js-spelling-suggestions` is needed.

app/views/finders/show_all_content_finder.html.erb
csutter: Good question: the `js-spelling-suggestions` ID is used by the existing finder UI to do its wonky JS-based page refresh.

As we will not be doing this for the new UI, at least not initially, any change of keywords/filters will trigger a full page refresh and not need any JS hacks to update parts of the page.

leenagupte (approved):
Thanks for responding. Looks good 👍

PR #3449, opened 2024-09-17
Filter section component

alexbowen:
This adds a new app-specific component that displays a details component to open/close a section of filters. The summary displays a section title (required) and optional status text if the filters within the section have been applied. This will be used for the new site search ("all content" finder) UI instead of the existing filters sidebar.

- Create `filter_section` component
- Add (empty for now) `filter_section` components to the (work in progress and hidden behind feature flag) `show_all_content_finder` view template

## Review app
- [Component guide preview](https://finder-frontend-pr-3449.herokuapp.com/component-guide/filter_section)
- [In context on new search page](https://finder-frontend-pr-3449.herokuapp.com/search/all)

### Closed
Screenshot 2024-09-18 at 15 22 21

### Hover
Screenshot 2024-09-18 at 15 22 57

### Focus
Screenshot 2024-09-18 at 15 22 31

### In context on new UI
Screenshot 2024-09-18 at 15 27 37

## Designs
https://www.figma.com/design/H99sZyzK3aT9Tet0r6AtOC/Search-Designs?node-id=1154-41574&node-type=canvas&t=aJqXa88NDTPd7HYg-0

### DIFF
--- a/app/assets/stylesheets/components/_filter-section.scss
+++ b/app/assets/stylesheets/components/_filter-section.scss
@@ -1,0 +1,75 @@
+@import "govuk_publishing_components/individual_component_support";
+@import "govuk_publishing_components/components/mixins/prefixed-transform";
+
+.app-c-filter-section {
+  border-bottom: 1px solid $govuk-border-colour;
+
+  &:last-child {
+    border-bottom: 0;
+    margin-bottom: 0;
+    padding-bottom: 0;
+  }
+
+  &:first-child {
+    margin-top: 0;
+  }
+
+  &[open] .app-c-filter-section__summary {
+    &::before {
+      @include prefixed-transform($translateY: 30%, $rotate: -45deg);
+    }
+  }
+}
+
+.app-c-filter-section__summary {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  cursor: pointer;
+  color: $govuk-text-colour;
+
+  &:hover .app-c-filter-section__summary-heading {
+    @include govuk-link-decoration;
+    @include govuk-link-hover-decoration;
+  }
+
+  &:focus .app-c-filter-section__summary-heading {
+    background-color: $govuk-focus-colour;
+    @include govuk-focused-text;
+  }
+
+  &::before {
+    border-style: solid;
+    border-width: 2px 2px 0 0;
+    content: '';
+    display: inline-block;
+    height: govuk-spacing(2);
+    left: govuk-spacing(1);
+    position: relative;
+    vertical-align: middle;
+    width: govuk-spacing(2);
+    border-color: $govuk-text-colour;
+    clip-path: unset;
+    @include prefixed-transform($translateY: -25%, $rotate: 135deg);
+  }
+
+  &:focus-visible {
+    outline: none;
+  }
+}
+
+.app-c-filter-section__summary-heading {
+  margin-left: govuk-spacing(3);
+  @include govuk-font(19, $weight: bold);
+}
+
+.app-c-filter-section__summary-status {
+  padding-right: govuk-spacing(1);
+  flex-grow: 1;
+  text-align: right;
+  @include govuk-font(16, $weight: regular);
+}
+
+.app-c-filter-section__content {
+  padding: govuk-spacing(2) 0;
+}
--- a/app/views/components/_filter_section.html.erb
+++ b/app/views/components/_filter_section.html.erb
@@ -1,0 +1,32 @@
+<% add_app_component_stylesheet("filter-section") %>
+<%
+  raise ArgumentError, "heading_text is required" unless local_assigns[:heading_text]
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  open ||= nil
+  status_text ||= ""
+  heading_text ||= ""
+  heading_level ||= 2
+  summary_aria_attributes ||= {}
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("app-c-filter-section")
+  component_helper.set_open(open)
+%>
+
+<%= tag.details(**component_helper.all_attributes) do %>
+  <%= tag.summary class: "app-c-filter-section__summary", aria: summary_aria_attributes do %>
+      <%= content_tag("h#{heading_level}", class: "app-c-filter-section__summary-heading") do %>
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/components/_filter-section.scss
```diff
@@ -0,0 +1,56 @@
...
+    border-bottom: 0;
+    margin-bottom: 0; 
+  }
+
+  &:first-child {
+    margin-top: 0; 
+  }
+
+  &[open] .app-c-filter-section__summary::before {
+    transform: translateY(70%) rotate(-45deg);
```
csutter, line 17: Might be worth using [`prefixed-transform` mixin](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/mixins/_prefixed-transform.scss#L1) from Publishing Components

app/views/components/docs/filter_panel.yml
```diff
@@ -31,6 +31,19 @@ examples:
         <p class="govuk-body">
           I am open by default!
         </p>
+  with_filter_section:
+    description: |
+      Pass filter section component as a block
+    data:
+      result_count: 1
+      button_text: View filter section
+      block: |
```
csutter, line 40: Oooh cool, good idea to demo this here I think 🙌

spec/components/filter_section_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  it "set section title text with hidden context for accessibility" do
+    render_component({ title: "section title" })
+
+    assert_select ".app-c-filter-section__summary-title", text: "filter by section title"
+    assert_select ".app-c-filter-section__summary-title .govuk-visually-hidden", text: "filter by"
+  end
+
+  it "set section status text" do
```
csutter, line 36: Is it worth testing that the status text message doesn't get rendered if it isn't given?

app/assets/stylesheets/components/_filter-section.scss
```diff
@@ -0,0 +1,56 @@
...
+    left: 0;
+    position: relative;
+    vertical-align: middle;
+    width: govuk-spacing(2);
+    border-color: $govuk-text-colour;
+    clip-path: unset;
+    transform: translateY(30%) rotate(135deg);
+  }
+
+  &-content {
```
csutter, line 39: There's a preference here to have BEM blocks and modifiers all separately in the SCSS file (so they're easier to search for in the code) instead of using parent modifier like this 💡

app/assets/stylesheets/components/_filter-section.scss
```diff
@@ -0,0 +1,56 @@
...
+    margin-top: 0; 
+  }
+
+  &[open] .app-c-filter-section__summary::before {
+    transform: translateY(70%) rotate(-45deg);
+  }
+}
+
+.app-c-filter-section__summary {
+  display: flex;
```
csutter, line 22: Do we need the `summary-content` block to be separate from the summary itself? Or can we just combine them?

app/views/components/_filter_section.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("app-c-filter-section")
+  component_helper.add_class(shared_helper.get_margin_bottom)
+  component_helper.set_open(open)
+%>
+
+<%= tag.details(**component_helper.all_attributes) do %>
+  <%= tag.summary class: "app-c-filter-section__summary", aria: summary_aria_attributes do %>
+    <div class="app-c-filter-section__summary-content">
+      <%= sanitize("<span class=\"app-c-filter-section__summary-title\"><span class=\"govuk-visually-hidden\">filter by </span>#{title}</span><span class=\"app-c-filter-section__summary-status\">#{status_text}</span>") %>
```
csutter, line 21: Don't think this needs to be output as a string. Also, should we not include the status span entirely if none is given?

```erb
<span class="app-c-filter-section__summary-title">
  <span class="govuk-visually-hidden">Filter by</span>
    <%= title %>
  </span>

  <% if status_text.present? %>
    <span class="app-c-filter-section__summary-status">
      <%= status_text %>
    </span>
  <% end %>
</span>
```

app/assets/stylesheets/components/_filter-section.scss
csutter: Also this probably needs a `cursor: pointer`. And should we think about having a hover state?

app/views/components/_filter_section.html.erb
alexbowen: markup was slightly wrong with closing tags, but in essence changed it as per your suggestion

app/assets/stylesheets/components/_filter-section.scss
```diff
@@ -0,0 +1,80 @@
+@import "govuk_publishing_components/individual_component_support";
+@import "govuk_publishing_components/components/mixins/prefixed-transform";
+
+@mixin filter-section-heading-underline {
+  text-decoration: underline;
+  text-underline-offset: 6px;
+  text-decoration-thickness: 4px;
```
csutter, line 7: Is there a reason we want these specific values?

I think `$govuk-link-underline-offset` and `$govuk-link-hover-underline-thickness` might be the way to go for consistency, and because they're more flexible with changing text sizes.

csutter: Or maybe just use
```scss
@mixin govuk-link-decoration
@mixin govuk-link-hover-decoration
```
... to get everything in one package, and also the `text-decoration-skip-ink` property

app/views/components/_filter_section.html.erb
```diff
@@ -0,0 +1,33 @@
+<% add_app_component_stylesheet("filter-section") %>
+<%
+  raise ArgumentError, "heading_text is required" unless local_assigns[:heading_text]
+  raise ArgumentError, "heading_level for a section should not be heading level 1" if local_assigns[:heading_level] == 1
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  open ||= nil
+  status_text ||= ""
+  heading_text ||= ""
+  heading_level ||= "2"
```
csutter, line 9: This should be `2` (integer) for consistency with similar components

alexbowen: good spot

csutter (approved):
🎉

PR #3450, opened 2024-09-17
New all content finder UI tweaks

csutter:
A couple of assorted UI tweaks to the new all content finder UI:
- Use `prefixed-transform` mixin instead of raw transform to improve support for legacy browsers
- Allow filter panel to accept verbatim result count text instead of generating result count text in the component based on a number, and only render heading if text is passed in
- Pass in result presenter's preformatted result text in all content finder view
- Add basic Cucumber step to validate result count is present


alexbowen (approved):
lgtm!

PR #3452, opened 2024-09-18
updated code to handle unfiltered parameters

sairamya93:


sairamya93:
Trello Card: https://trello.com/c/bP7hTCJg/343-finder-frontend-does-not-gracefully-handle-unpermitted-parameters

URL with Unfiltered parameters - Before

Screenshot 2024-09-19 at 12 04 18

URL with Unfiltered parameters - After handling the exception

Screenshot 2024-09-19 at 12 06 58

spec/controllers/finders_controller_spec.rb
```diff
@@ -96,6 +96,12 @@
         get :show, params: { slug: "lunch-finder" }
         expect(response.status).to eq(406)
       end
+
```
KludgeKML, line 99: Can we update this test case, because the fix works, but I don't think this test case covers it.

KludgeKML (approved):
LGTM!

PR #3455, opened 2024-09-18
alignment of filter section/panel styles

alexbowen:
This addresses some design inconsitencies between the filter panel and sections:

- chevrons on panel and sections for filters were inconsistent sizes. Creates a shared mixin to address inconsistency
- chevron is consistently applied to before pseudo class on element where is is required
- correct margin between panel header and content as per designs
- align css class names as per GDS model
- align the panel toggle button hover/focus styling with the panel section summary elements
- remove uneccessary nesting to be inline with GDS conventions

## before

Screenshot 2024-09-19 at 10 53 11

## after
Screenshot 2024-09-19 at 10 52 15

### DIFF
--- a/app/assets/stylesheets/components/_filter-panel.scss
+++ b/app/assets/stylesheets/components/_filter-panel.scss
@@ -1,5 +1,5 @@
 @import "govuk_publishing_components/individual_component_support";
-@import "govuk_publishing_components/components/mixins/prefixed-transform";
+@import "mixins/chevron";
 
 .app-c-filter-panel__content {
   background-color:  govuk-colour("light-grey");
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/components/_filter-panel.scss
```diff
@@ -15,6 +15,7 @@
   gap: govuk-spacing(2);
 
   .gem-c-heading {
```
andysellick, line 17: We shouldn't be styling anything with a `.gem-c-` prefix in applications. This breaks our model of component isolation. If changes are needed they should be made directly to the component in `govuk_publishing_components` - or applied to surrounding elements (the margin-left could easily be done as padding left on the wrapper for this heading component).

alexbowen: in actual fact dont need the style at all, the two elements used to b the other way round, think thats why it was there

app/assets/stylesheets/components/_filter-panel.scss
```diff
@@ -14,7 +14,7 @@
   align-items: center;
   gap: govuk-spacing(2);
 
-  .gem-c-heading {
+  .js-app-c-filter-panel__count {
```
csutter, line 17: The `js-` prefix is only for elements that are used as a target in Javascript. Don't think this is? Should probably just be `.app-c-filter-panel__count`.

app/views/components/_filter_panel.html.erb
```diff
@@ -30,6 +29,7 @@
         text: result_text,
         heading_level: 2,
         font_size: "s",
+        classes: "js-app-c-filter-panel__count"
```
csutter, line 32: See `js` prefix comment above.

app/assets/stylesheets/components/_mixins.scss
```diff
@@ -0,0 +1,22 @@
...
+    width: govuk-spacing(2);
+    border-color: initial;
+    clip-path: unset;
+    @include prefixed-transform($translateY: -25%, $rotate: 135deg);
+
+    @if $open {
+      @include prefixed-transform($translateY: 30%, $rotate: -45deg);
+    }
+  }
+}
```
csutter, line 22: Looking at patterns elsewhere, I think this file would be better off being named `.../stylesheets/components/mixins/_chevron.scss`

csutter (commented):
Looks good, couple of comments and then we should be good to go.

Can you update the title/description to make it clear it's not just the chevrons but the filter panel and sections overall (as we're also changing some other style aspects like the hover states)?

app/assets/stylesheets/components/_filter-panel.scss
csutter: Maybe instead of using the heading component, we can just manually add a heading in given that we're overriding the styling anyway?

app/assets/stylesheets/components/_filter-panel.scss
```diff
@@ -14,44 +14,40 @@
   align-items: center;
   gap: govuk-spacing(2);
 
-  .gem-c-heading {
-    font-weight: normal;
+  .app-c-filter-panel__count {
```
andysellick, line 17: Does this need to be nested beneath `.app-c-filter-panel__content`? We try to avoid nesting if possible, particularly since we're using BEM here.

csutter (approved):

PR #3456, opened 2024-09-19
Fix sort order behaviour bug when JS disabled

csutter:
When a finder has both relevance and popularity options, `SortPresenter` will check for the presence of keywords and force the relevance option even if popularity was explicitly requested. It doesn't do the converse though: force the popularity option when relevance was explicitly requested but no keyword(s) given.

This is papered over with Javascript for those users who have it enabled, but since our upcoming search UI does not have dynamic behaviour, it has surfaced this bug.

- Add tests for overriding explicitly provided sort order
- Fix missing override of relevance sort order when no keyword(s) given
- Tidy up `SortPresenter#user_selected_option` a little bit

## How to reproduce
- Disable Javascript
- Go to "all content" finder
- Note that "Most viewed" is the selected default sort option (correctly so)
- Enter a keyword and submit
- Note that the sort order is overridden to "Relevance" (correctly so)
- Remove the keyword and submit again
- **"Relevance" incorrectly remains selected, despite being correctly disabled**:

Screenshot 2024-09-19 at 11 29 08

### DIFF
--- a/app/presenters/sort_presenter.rb
+++ b/app/presenters/sort_presenter.rb
@@ -62,8 +62,12 @@ private
 
   def user_selected_option
     selected = sort_options.find { |option| option_value(option) == user_selected_order }
-    if selected == popularity_option && !relevance_option.nil? && keywords.present?
+    return selected if relevance_option.nil? || popularity_option.nil?
+
+    if selected == popularity_option && keywords.present?
       relevance_option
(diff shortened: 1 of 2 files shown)


leenagupte (approved):
Thanks for fixing this! ⭐ 

I loved the explanation in the PR about what was wrong and how to see it for yourself. 🪇 🥇

PR #3459, opened 2024-09-19
Filter section: make a11y heading prefix configurable

csutter:
This shouldn't always be set, or hardcoded to "filter by" - for example, we will have a "Sort by" section that shouldn't have a prefix at all.

### DIFF
--- a/app/views/components/_filter_section.html.erb
+++ b/app/views/components/_filter_section.html.erb
@@ -4,7 +4,7 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   open ||= nil
   status_text ||= ""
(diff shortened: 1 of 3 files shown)


app/views/components/_filter_section.html.erb
```diff
@@ -4,7 +4,7 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   open ||= nil
   status_text ||= ""
-  heading_text ||= ""
+  visually_hidden_heading_prefix ||= ""
```
csutter, line 7: Note: We already check on line 3 whether `heading_text` is given, so we don't need to assign it anything here.

alexbowen (approved):

PR #3465, opened 2024-09-20
Update all content finder fixture

csutter:
This doesn't accurately represent the facets of the all content finder anymore, limiting our ability to write reasonable and representative Cucumber tests for the new UI.

- Update `all_content.json` with the current state of the all content finder
- Add basic `DocumentHelper#stub_topical_events_api_request` to stub requests from the topical events registry triggered by the newly added hidden topical event facet, and use it where needed
- Remove stubbing of no longer used people registry from all content finder step
- Remove now incorrect default sort order from request stubs


leenagupte (approved):
⭐

PR #3471, opened 2024-09-23
All content finder: Initial filter summary implementation

csutter:
This removes the current hardcoded demo options on the new "all content" finder UI's filter summary, and replaces them with actual real data.

As a first step, this only shows the sort option as the other facets are not implemented yet. As we implement them, they too will be added to show filter tags here.

---

image

---

> [!NOTE]
> The view being updated by this PR is behind a feature flag and not yet visible to users.


csutter:
Closing draft as now implemented differently in #3484.

PR #3478, opened 2024-09-25
Change facet visibility terminology

csutter:
Currently, facets implement a `Facet#has_ga4_section?` method that maps 1:1 onto whether a facet is visible to the user, or hidden (because every visible facet should be tracked in GA4).

This change:
- Adds an explicit and more obvious `Facet#user_visible?` method instead, and removes `#has_ga4_section?`
- Renames `FacetsIterator#each_with_index_and_count` to `#each_with_visible_index_and_count` to more accurately describe what it does
- Adds tests to the currently untested `FacetIterator` and visibility methods

### DIFF
--- a/app/lib/facets_iterator.rb
+++ b/app/lib/facets_iterator.rb
@@ -6,7 +6,7 @@ class FacetsIterator
     @user_visible_facets = @facets.select(&:user_visible?)
   end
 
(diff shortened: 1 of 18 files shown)


leenagupte (approved):
This is a great name change. Does what it says on the tin ✨

PR #3479, opened 2024-09-26
New all content finder UI: Add ability to sort

csutter:
This adds the ability to sort results to the new "all content" finder UI.

This is done in a slightly clever way by adding a "virtual" facet (in the sense that it doesn't exist in the underlying finder content item) that wraps the `SortPresenter` and prepending it to the "real" facets if the new UI is active.

This is done for two reasons:
- The new design has sorting as part of the filters (rather than an independent UI element), which means it needs to act like it's part of the facets for analytics tracking purposes – but making sort act like a facet also allows us to reuse a lot of code dedicated to rendering facets
- It allows us to do the minimum amount of work necessary, without having to engage in a major refactor of `SortPresenter` and its overwhelming tech debt that risks breaking existing functionality

## Review app
TBC

## UI changes
image


app/controllers/finders_controller.rb
```diff
@@ -124,9 +124,15 @@ def results
   end
 
   def all_facets
-    return @facets if defined?(@facets)
-
-    FacetsBuilder.new(content_item:, search_results:, value_hash: filter_params).facets
+    @all_facets ||= FacetsBuilder.new(
```
KludgeKML, line 127: Quick double-check - I assume it's safe to memoize this (ie we're expecting this method to always return the same values no matter how many times it's called during any one call to the action that references it?)

app/models/sort_facet.rb
```diff
@@ -10,6 +12,14 @@ def name
   end
   alias_method :ga4_section, :name
 
+  def key
+    KEY
+  end
+
+  def options
```
KludgeKML, line 19: I'm not super-enthused about the idea of the model knowing about the presenter - I think it might make more sense if the presenter is something the view creates/calls passing the model to it.

app/controllers/finders_controller.rb
csutter: I think so: I think [this change](https://github.com/alphagov/finder-frontend/commit/51ef79134a1bcab9bfa7de2b14aaaa07fb64e27d) accidentally removed the memoizing but it should be there.

KludgeKML: Okay, cool.

app/models/sort_facet.rb
csutter: Yeah, me neither. Irritatingly most of the existing facet logic contains a lot of presentation concerns, so I felt less bad about doing this here.

The partial that renders this facet does implicitly have access to the sort presenter (by virtue of the controller action setting it up as an instance variable), so if you think it would be better to just use it from the view instead of passing it through here I'm happy to do that?

KludgeKML: Ah, a classic quandary: "should I follow the broader idiom, or should I follow the idiom of this specific app?".

I think in this case if it doesn't add in a lot of extra code it would be better to use it from the view, and you can remove the model's requirement to know anything about the presenter.

KludgeKML (approved):
LGTM

PR #3481, opened 2024-09-26
Tidy up `DocumentHelper`

csutter:
We shouldn't have to keep stubbing Search API v1 and v2 separately. This adds a helper method that stubs them both and cleans up repetitive code in each of the stub methods.

### DIFF
--- a/features/support/document_helper.rb
+++ b/features/support/document_helper.rb
@@ -22,163 +22,127 @@ module DocumentHelper
     stub_content_store_has_item("/", "links" => { "level_one_taxons" => [] })
   end
 
-  def stub_search_api_request
-    stub_request(:get, SEARCH_ENDPOINT)
-      .with(query: hash_including(rummager_all_documents_params))
-      .to_return(body: all_documents_json)
+  def stub_response(query, body, including_v2: false)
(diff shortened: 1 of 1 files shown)


KludgeKML:
Nice. One question - is it possible that we will get a false positive test response due to this? (ie now that we're always stubbing both V1 and V2 requests, do we have tests that are currently working only on V1 where the test might accidentally use V2 and would miss a regression in V1). Or are the tests always very specific about what api they're using, and they'll still break in the right places?

csutter:
> Nice. One question - is it possible that we will get a false positive test response due to this? (ie now that we're always stubbing both V1 and V2 requests, do we have tests that are currently working only on V1 where the test might accidentally use V2 and would miss a regression in V1). Or are the tests always very specific about what api they're using, and they'll still break in the right places?

Good question. This whole kerfuffle comes out of the fact that the "all content" finder can still use v1 as an emergency fallback (with a feature flag), and we want to keep that behaviour for now, so anywhere v2 can be called needs to be able to stub v1 (but not vice versa).

You're right that there's a risk that now that we now stub v2 for a whole bunch of requests that it technically doesn't support (but may do so in the future), we might give someone false confidence that it does 😬 

Have amended with a parameter to explicitly enable stubbing v2!

PR #3483, opened 2024-09-26
New all content finder UI: Implement facets

csutter:
This implements the remaining facets for the new "all content" finder UI.

For the most part, this reuses markup from the existing facets, but without the expandable sections or the JS support.

## Review app
https://finder-frontend-pr-3483.herokuapp.com/search/all

## UI changes
image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -44,3 +44,16 @@ if ($form && $results) {
     $atomAutodiscoveryLink: $atomAutodiscoveryLink
   })
 }
+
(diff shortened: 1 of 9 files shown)


csutter:
> I'm a bit unsure about "Enter a real date" as the feedback for invalid date formats, but I assume a content person has signed off on that (or will look at it and revise later), so LGTM.

Yes - this probably needs tweaking! But it's there already 🫠 : https://github.com/alphagov/finder-frontend/blob/main/app/validators/date_validator.rb#L48

PR #3490, opened 2024-10-04
All content finder: Implement hidden clearable facet

csutter:
This implements _persisting_ the hidden clearable facet for the new "all content" finder UI across searches (they already work, but are lost when modifying the user-changeable filters or search terms).

These are facets that are able to be used with the finder as query parameters, and if used are shown to the user, but do not have UI to manually select. We can have a much simpler implementation than the existing finder UI because we do not need to consider Javascript.

## Testing on review app
- Visit [a page with hidden clearable manual, organisation, and topical events filters](https://finder-frontend-pr-3490.herokuapp.com/search/all?keywords=test&order=most-viewed&level_one_taxon=&manual%5B%5D=%2Fguidance%2Fmot-inspection-manual-for-private-passenger-and-light-commercial-vehicles&organisations%5B%5D=hm-revenue-customs&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&topical_events%5B%5D=election-2024) (note that obviously there's no content matching all three of these filters)
- Change any user-facing filter and/or change the search term
- Note that the hidden clearable filters persist

### DIFF
--- a/app/models/hidden_clearable_facet.rb
+++ b/app/models/hidden_clearable_facet.rb
@@ -1,4 +1,6 @@
 class HiddenClearableFacet < FilterableFacet
+  attr_reader :value
+
   def initialize(facet, value_hash)
     @value = Array(value_hash)
     super(facet)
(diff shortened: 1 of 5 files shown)


leenagupte (approved):
✨

PR #3495, opened 2024-10-07
New "all content" finder UI tweaks

csutter:
This implements a number of small UI tweaks following a design review:
- Apply button should not have `name` attr (this causes the button's text to be added to the query parameters, which we don't need and clutters up the analytics)
- Improve spacing of page heading by moving overall UI into a container and adjusting margins
- Fix garbled disclosure marker on filter section `details` element on Webkit and older Blink browsers (by hiding the `::marker` pseudo-element)
- Fix incorrect "Filter and sort" button underline on Webkit (caused by Webkit not respecting `text-decoration-thickness` on `button` elements, and fixed by applying it to an extra inner `span` instead)
- Tweak vertical alignment of result count in filter panel summary
- Improve display of spelling suggestion (by inlining only the bits we need from the existing old finder partial, and tweaking whitespace)
- Remove standalone page heading and use non-inline label search field with a heading-wrapped label
- Increase gap between filter panel action buttons
- Increase padding of filter sections
- Make text in filter summary more consistently sized (19px)
- Tweak spacing of tags in filter summary and increase spacing underneath
- Remove duplicate width container causing display issues at some breakpoints
- Only use two thirds layout on desktop breakpoint and above (remaining full width on tablet)
- Tidy and fix broken form markup (tag wasn't closed), and use ERB to generate the form so the view code looks more manageable
- Use document list component's new `equal_item_spacing` option to tighten up display of results
- Remove bottom border on summary component and stop removing the top border on the first document list element instead

**Review app: https://finder-frontend-pr-3495.herokuapp.com/search/all**

### DIFF
--- a/app/views/finders/show_all_content_finder.html.erb
+++ b/app/views/finders/show_all_content_finder.html.erb
@@ -26,13 +26,6 @@
     <div class="govuk-width-container">
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-          <%= render "govuk_publishing_components/components/heading", {
-            text: sanitize("Search <span class='govuk-visually-hidden'>all content on</span> GOV.UK"),
-            heading_level: 1,
-            font_size: "xl",
-            margin_bottom: 4,
-          } %>
-
           <div id="keywords" role="search" aria-label="Sitewide" data-ga4-change-category="update-keyword text">
             <%= render "govuk_publishing_components/components/search", {
               id: "finder-keyword-search",
@@ -40,6 +33,14 @@
               type: 'search',
               value: result_set_presenter.user_supplied_keywords,
               disable_corrections: true,
+              label_size: "xl",
+              label_text: safe_join([
+                "Search ",
+                tag.span("all content on", class: "govuk-visually-hidden"),
+                " GOV.UK",
+              ]),
+              wrap_label_in_a_heading: true,
+              heading_level: 1,
               margin_bottom: 4,
             } %>
           </div>
--- a/app/views/finders/show_all_content_finder.html.erb
+++ b/app/views/finders/show_all_content_finder.html.erb
@@ -96,6 +96,7 @@
           <% if result_set_presenter.total_count.positive? %>
             <%= render "govuk_publishing_components/components/document_list", {
               remove_top_border_from_first_child: true,
+              margin_bottom: 5,
               disable_ga4: true,
               items: result_set_presenter.search_results_content[:document_list_component_data],
             } %>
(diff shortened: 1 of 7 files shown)


app/views/finders/show_all_content_finder.html.erb
```diff
@@ -43,9 +44,24 @@
...
           </div>
 
-          <div class="spelling-suggestions">
-            <%= render 'spelling_suggestion' %>
-          </div>
+          <% if @spelling_suggestion_presenter.suggestions.any? %>
+            <% suggestion = @spelling_suggestion_presenter.suggestions.first %>
+
+            <div class="app-all-content-finder__spelling-suggestions">
+              Did you mean <%= link_to sanitize(suggestion[:highlighted]), suggestion[:link],
```
kevindew, line 53: Is the source of highlighted quite trusted? Just wanted to check given sanitize is quite broad in what it accepts and I'm wondering if you only care about allowing a single element (mark) in and no attributes e.g. `sanitize(suggestion[:highlighted], tags: %w[mark], attributes: [])`

app/views/finders/show_all_content_finder.html.erb
```diff
@@ -26,20 +26,17 @@
...
-
           <div id="keywords" role="search" aria-label="Sitewide" data-ga4-change-category="update-keyword text">
             <%= render "govuk_publishing_components/components/search", {
               id: "finder-keyword-search",
               name: "keywords",
               type: 'search',
               value: result_set_presenter.user_supplied_keywords,
               disable_corrections: true,
+              label_size: "xl",
+              label_text: sanitize("Search <span class='govuk-visually-hidden'>all content on</span> GOV.UK"),
```
kevindew: Are we just using sanitize to appease an error here? Since it has no external input `.html_safe` would feel more idiomatic - though since that is often a bit smelly `safe_join` could be best:

```
safe_join(["Search ", tag.span("all content on", class: "govuk-visually-hidden"), " GOV.UK"])
```

app/views/finders/show_all_content_finder.html.erb
```diff
@@ -107,5 +115,5 @@
         <%= render "govuk_publishing_components/components/previous_and_next_navigation", @pagination.next_and_prev_links %>
       </div>
     </div>
-  </div>
+  <% end %>
```
kevindew, line 119: Nice spot 👍

kevindew (commented):
Nice one, looks good to me. Just a couple of things related to HTML escaping.

app/views/finders/show_all_content_finder.html.erb
csutter: Ah yes, this is just inlined from the old partial but no reason we can't improve things here 👍

kevindew (approved):
Thanks

PR #3498, opened 2024-10-10
All content finder: Use full size result description

csutter:
This enables the `full_size_description` option on the document list component for search results to match the new search UI design, if the results are presented for the all content finder.

### Before
image

### After
image


leenagupte (approved):
🎉

PR #3499, opened 2024-10-10
All content finder: JS module & analytics improvements

csutter:
## JS module
The all content finder currently has some ad hoc Javascript in application.js to handle setting up the taxonomy select and rudimentary analytics tracking.

This has worked well for prototyping, but lacks tests and will become messy when we need to add additional code imminently for further GA4 tracking.

This PR adds an `all-content-finder` Javascript module and moves the existing logic into it, adding tests for good measure.

## Analytics improvements
This PR also improves the GA4 tracking on the all content finder JS module to start tracking as soon as a user consents to usage cookies (not just from the next page navigation/reload), and sets up eCommerce tracking which was missing until now.


app/assets/javascripts/modules/all-content-finder.js
```diff
@@ -0,0 +1,53 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  // Enhances the all content finder with dynamic behaviour and analytics tracking
+  class AllContentFinder {
```
kevindew, line 8: I did wonder about whether to question ES6 or not here, but I notice we're using Element.closest later which is modern JS without a polyfill so I guess we're already there with modern JS.

app/assets/javascripts/modules/all-content-finder.js
```diff
@@ -0,0 +1,53 @@
...
+    constructor (module) {
+      this.module = module
+
+      this.$form = module.querySelector('.js-all-content-finder-form')
+      this.$taxonomySelect = module.querySelector('.js-all-content-finder-taxonomy-select')
```
kevindew, line 13: If you're following the convention that dom elements need a $ prefix then this.module should have one too.

app/assets/javascripts/modules/all-content-finder.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    init () {
+      this.setupTaxonomySelect()
+
+      if (this.userHasConsentedToAnalytics()) {
+        this.setupAnalyticsTracking()
+      } else {
+        // Allow tracking of events as soon as user consents, not just at next page reload
+        window.addEventListener('cookie-consent', this.setupAnalyticsTracking.bind(this))
```
kevindew, line 23:
```suggestion
        window.addEventListener('cookie-consent', () => { this.setupAnalyticsTracking() })
```

Seems a bit more idiomatic ES6 to use an arrow function here.

spec/javascripts/modules/all-content-finder.spec.js
```diff
@@ -0,0 +1,151 @@
...
+        window.dispatchEvent(new Event('cookie-consent'))
+
+        setTimeout(() => {
+          const event = new Event('change', { bubbles: true })
+          const input = fixture.querySelector('#foo')
+          input.dispatchEvent(event)
+
+          expect(GOVUK.analyticsGa4.Ga4FinderTracker.trackChangeEvent).toHaveBeenCalledWith(input, 'FooCategory')
+          done()
+        }, 1) // allow for async event to propagate
```
kevindew, line 138: Are you sure you need this? I was under the impression dispatchEvent was synchronous and not sure there's any async code (same for below)

kevindew (commented):
Looks great - just a few minor comments.

app/assets/javascripts/modules/all-content-finder.js
csutter: Yeah – figured we might as well stick with ES6 for this as a "new" piece of functionality. Nothing in this is (or will be as far as I can tell) on the critical path UX-wise, it's just progressive enhancement and analytics so I don't feel too bad about legacy browsers.

spec/javascripts/modules/all-content-finder.spec.js
csutter: Very good question. I had the tests crash and thought this fixed it, but maybe it was ~random solar flares~ a transient local issue because it works today 🤷‍♂️

kevindew (approved):
Nice one 👍

PR #3500, opened 2024-10-10
All content finder: More small styling fixes

csutter:
- Fix missing border when filter summary is shown
- Fix slightly wonky padding when filter
- Fix misaligned spelling suggestions (and move from `div` to `p` with appropriate style)

## Before
image

## After
image


leenagupte (approved):
Looks great! 🎉

PR #3502, opened 2024-10-11
Resolve Sass deprecation warnings

csutter:
- Move vanilla declarations above Sass `@include`s
- Wrap division in a `calc()`


csutter:
Shouldn't break anything anyway, but have tested on review app and it's no more broken than it already is.

matthillco (approved):
Looks good to me. Thanks for doing this 👏

PR #3503, opened 2024-10-11
All content finder: Clean up templates

csutter:
The component and facet templates have a fair bit of unidiomatic code in them, including:
- Using symbols with dashes in them for data attributes, instead of letting Rails automatically dasherise more idiomatic underscored symbols
- Setting data attributes directly on top level of tags instead of `data` hash
- Verbose manual tags with basic string content where tag helper can be more succinct

This tidies these templates up a bit to make them more readable, consistent, and maintainable.

> [!NOTE]
> The diff for this PR is best viewed with whitespace changes hidden (`?w=1`)

### DIFF
--- a/app/views/components/_filter_panel.html.erb
+++ b/app/views/components/_filter_panel.html.erb
@@ -34,7 +34,7 @@
           text: button_text,
           index_section: 0,
           index_section_count: section_count
-        }.to_json
+        }
       }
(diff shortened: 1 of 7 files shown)


leenagupte (approved):
Thanks for the `?w=1` tip. Very nice!
The changes look good too 😅

PR #3505, opened 2024-10-14
Improve handling of facet sections for GA4

csutter:
The current logic for dealing with section indices and counts is quite
messy, with the `FacetsIterator` yielding several values that are then
passed through several layers of views and partials.

This introduces a `FacetPresenter` that holds section index/count values
(appropriately using 1-based indexing) and makes the iterator yield that
instead.

It then changes all facet partials (both current UI and new all content
finder UI) to use `FacetPresenter` instead of manipulating and passing
index and count values through several layers of view code.

Where appropriate, it also cleans up the old partials to use idiomatic
modern Ruby syntax, better spacing, and Rails-y underscored data
attribute values.

### DIFF
--- a/app/lib/facets_iterator.rb
+++ b/app/lib/facets_iterator.rb
@@ -1,6 +1,5 @@
 class FacetsIterator
   include Enumerable
-  delegate :each, to: :@facets
 
   def initialize(facets)
     @facets = facets
@@ -11,9 +10,9 @@ class FacetsIterator
     @user_visible_facets.count
   end
 
-  def each_with_visible_index_and_count
+  def each
     @facets.each do |facet|
-      yield facet, @user_visible_facets.index(facet), user_visible_count
+      yield FacetPresenter.new(facet, @user_visible_facets.index(facet), user_visible_count)
     end
   end
 end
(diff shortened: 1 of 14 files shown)


app/presenters/facet_presenter.rb
```diff
@@ -0,0 +1,11 @@
+class FacetPresenter < SimpleDelegator
+  def initialize(facet, section_index, section_count)
+    super(facet)
+
+    # Facet index is 1-based for GOV.UK analytics
+    @section_index = section_index + 1 if section_index
+    @section_count = section_count
+  end
+
+  attr_reader :section_index, :section_count
```
kevindew, line 10: Would suggest moving this to top of the file to be conventional with rest of this apps codebase.

app/lib/facets_iterator.rb
```diff
@@ -11,9 +11,9 @@ def user_visible_count
     @user_visible_facets.count
   end
 
-  def each_with_visible_index_and_count
+  def each
```
kevindew, line 14: There's a delegator at the top of this file that is now no longer needed.

kevindew (approved):
Looks good to me, nice improvement. Just 2 minor things but I don't think they warrant another review unless they're more to a quick tweak.

PR #3508, opened 2024-10-15
FacetPresenter: Move common attributes to presenter

csutter:
All of the new facet partials have several shared attributes that are better off living in the presenter rather than being repeated several times over in the partials.

- Add `FacetPresenter#section_attributes` with common attributes and use in facet partials
- Remove unnecessary (unused) HTML ID on sections
- Tweak `FacetPresenter` spec to use different values for section index and count to avoid confusing the values in the future


kevindew (approved):
Nice refactoring 👍

PR #3509, opened 2024-10-15
All content finder: Assorted GA4 tracking fixes

csutter:
This fixes a number of paper cuts with GA4 tracking on the new all content finder UI, including:
* Missing result count on eCommerce tracking
* Filter removal through tags in filter summary including the visually hidden text
* "Clear all filters" button missing section index/count
* "Apply filters"/"Clear all filters" buttons including the wrong text (panel open button's text instead of their own)

### DIFF
--- a/app/presenters/filters_presenter.rb
+++ b/app/presenters/filters_presenter.rb
@@ -13,6 +13,7 @@ class FiltersPresenter
       {
         label: filter[:name],
         value: filter[:label],
+        displayed_text: "#{filter[:name]}: #{filter[:label]}",
         remove_href: finder_url_builder.url_except(filter[:query_params]),
         visually_hidden_prefix: "Remove filter",
       }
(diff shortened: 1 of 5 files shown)


kevindew (commented):
Looks good to me. I was surprised by the first commit so I've reached out to you on Slack to check the id behaviour is correct, as it seems so weird to require such a generic id - however I'm assuming it is right and a quirk in our system so pre-emptively approving

kevindew (approved):

csutter:
> Looks good to me. I was surprised by the first commit so I've reached out to you on Slack to check the id behaviour is correct, as it seems so weird to require such a generic id - however I'm assuming it is right and a quirk in our system so pre-emptively approving

Just noting down our conclusion from Slack: this comes from [publishing components analytics code](https://github.com/alphagov/govuk_publishing_components/blob/9ea8d10f6eca54b4ef584f247b6f598365b2af7e/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js#L48) and is too deeply ingrained there to be worth fixing now. 😞

PR #3510, opened 2024-10-15
All content finder: Assorted component tidying

csutter:
This is an assortment of tech debt fixes and improvements for the filter-related components on the new all content finder UI:

* Hardcode the "filter by" prefix in the filter section component (as this is always either "filter by" or `nil` and we have no need to change it)
* Simplify the filter summary component by removing various unnecessary nil checks and providing default values
* Use consistent "reset" instead of "clear all" terminology across all components


kevindew (approved):
Looks good to me - nice job

PR #3515, opened 2024-10-17
All content finder: Avoid cluttered URLs on submit

csutter:
This adds some progressive enhancement to remove empty fields and the default sort order from the search form's `FormData` on submission.

This leads to shorter, more aesthetically pleasing URLs and makes life easier for performance analysts. In the event that the user doesn't have Javascript enabled, the regular HTML form submission behaviour takes over and the form will continue to work as intended (and they won't fire analytics events anyway).

Currently, when submitting an empty search form, you end up on:
```
/search/all?keywords=&order=most-viewed&level_one_taxon=&public_timestamp%5Bfrom%5D%5Bday%5D=&public_timestamp%5Bfrom%5D%5Bmonth%5D=&public_timestamp%5Bfrom%5D%5Byear%5D=&public_timestamp%5Bto%5D%5Bday%5D=&public_timestamp%5Bto%5D%5Bmonth%5D=&public_timestamp%5Bto%5D%5Byear%5D=
```

After, you get:
```
/search/all?
```

### DIFF
--- a/app/assets/javascripts/modules/all-content-finder.js
+++ b/app/assets/javascripts/modules/all-content-finder.js
@@ -10,10 +10,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$module = $module
       this.$form = $module.querySelector('.js-all-content-finder-form')
       this.$taxonomySelect = $module.querySelector('.js-all-content-finder-taxonomy-select')
+      this.defaultSortOrders = ['relevance', 'most-viewed']
     }
 
     init () {
       this.setupTaxonomySelect()
+      this.setupFormDataCleaner()
 
       if (this.userHasConsentedToAnalytics()) {
         this.setupAnalyticsTracking()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/all-content-finder.js
```diff
@@ -10,10 +10,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$module = $module
       this.$form = $module.querySelector('.js-all-content-finder-form')
       this.$taxonomySelect = $module.querySelector('.js-all-content-finder-taxonomy-select')
+      this.default_sort_orders = ['relevance', 'most-viewed']
```
csutter, line 13: I was going to use a class field for this but Standard isn't a fan!

app/assets/javascripts/modules/all-content-finder.js
```diff
@@ -34,6 +36,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$taxonomySelect.addEventListener('change', () => taxonomySelect.update())
     }
 
+    // A best effort, progressive enhancement to keep URLs aesthetically pleasing and easier to
+    // manage in terms of analytics by removing empty fields and default sort order from the form
+    // data on submission (so they don't clutter up the query parameters)
+    setupFormDataCleaner () {
+      this.$form.addEventListener('formdata', (e) => {
```
kevindew, line 43: I had no idea about this formdata event - very cool!

app/assets/javascripts/modules/all-content-finder.js
kevindew: I think you should capitalise this to represent it as a constant (or if that's not appealing change it to camel case - as the snake case looks inconsistent)

kevindew (commented):
Real nice feature 👍 

I've spotted a small thing to fix, only other minor concern is whether the iteration through the formData is a bit cryptic but could only think of clunky alternatives so think it's good.

app/assets/javascripts/modules/all-content-finder.js
csutter: Gah - I had it capitalised when I tried to use it as a class field, but then regressed into Ruby naming when that didn't work! 🤦

kevindew (approved):
Nice one 👍

PR #3516, opened 2024-10-18
`DateParser`: Allow nil input

csutter:
There are edge cases where a date parameter is provided but left empty, e.g. `public_timestamp%5Bto%5D`. In those cases, rather than an empty string (which would be the case for `public_timestamp%5Bto%5D=`), Rails will return nil for the param's value, breaking `DateParser`.

### DIFF
--- a/app/parsers/date_parser.rb
+++ b/app/parsers/date_parser.rb
@@ -9,7 +9,9 @@ class DateParser
       DateStringParser.new(date_param).parse
     when Hash
       DateHashParser.new(date_param).parse
-    else raise ArgumentError, "date_param must be a String or a Hash"
+    when nil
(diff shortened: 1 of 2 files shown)


app/parsers/date_parser.rb
```diff
@@ -9,6 +9,8 @@ def parse
       DateStringParser.new(date_param).parse
     when Hash
       DateHashParser.new(date_param).parse
+    when nil
+      nil
     else raise ArgumentError, "date_param must be a String or a Hash"
```
kevindew, line 14: Should this be updated now that we allow nil?

kevindew (approved):
Nice one

PR #3518, opened 2024-10-18
Taxon facet: Move from "Sub-topic" to "Subtopic"

csutter:
> [!WARNING]
> This shouldn't be merged until just before the new UI is released (feature flag is switched), otherwise it might impact analytics.

This is our preferred spelling and should be reflected across the entire UI.

### DIFF
--- a/app/models/taxon_facet.rb
+++ b/app/models/taxon_facet.rb
@@ -15,10 +15,10 @@ class TaxonFacet < FilterableFacet
     level_one_taxons.unshift(default_topic_value)
   end
 
-  def sub_topics
-    return [default_sub_topic_value] unless level_two_taxons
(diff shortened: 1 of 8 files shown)


kevindew (commented):
Controversial change! What's wrong with Sub-taxon? :trollface: 

I did a search and found one usage in application code: https://github.com/alphagov/finder-frontend/blob/55171844755ec386a571bdfd41724000749295bb/app/models/taxon_facet.rb#L133-L135 then there's a few in tests.

If you're being a completist too you might want to change the variable/method names from sub_topic to subtopic

csutter:
> I did a search and found one usage in application code
Serves me right for case-sensitive searching 🤦 Fixed!

PR #3520, opened 2024-10-18
Change date validation error message

csutter:
"Enter a real date" isn't very GOV.UK - this should just be "enter a date".


kevindew (approved):
Nice catch. I do like that it'll happily accept "thirteen autumns ago" as a "real date" 😁

PR #3521, opened 2024-10-21
All content finder: GA4 improvements

csutter:
This PR improves the GA4 tracking for finders, mostly for the benefit of the new all content finder UI (but some changes also apply to the old UI).

In particular, it:
- Changes the finder tracker to include section details (index/count/link) on removal and not just on adding
- Adds a new `date` update type for the finder tracker to handle the three-field date input on the new UI
- Enables tracking date input errors on the new UI using `ga4-auto-tracker`

### DIFF
--- a/app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
+++ b/app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
@@ -96,16 +96,16 @@
             schema.section = section.getAttribute('data-ga4-section')
           }
 
+          var index = this.getSectionIndex(filterParent)
           if (wasFilterRemoved) {
             schema.action = 'remove'
             schema.text = elementType === 'text' ? undefined : elementValue
           } else {
             schema.action = elementType === 'text' ? 'search' : 'select'
-            var index = this.getSectionIndex(filterParent)
-            schema.index_link = index.index_link || undefined
-            schema.index_section = index.index_section || undefined
-            schema.index_section_count = index.index_section_count || undefined
           }
+          schema.index_link = index.index_link || undefined
+          schema.index_section = index.index_section || undefined
+          schema.index_section_count = index.index_section_count || undefined
           break
 
         case 'update-keyword':
--- a/app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
+++ b/app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
@@ -24,6 +24,9 @@
       data.event_name = 'select_content'
 
       var elementInfo = this.getElementInfo(eventTarget, elementType, section)
+      if (!elementInfo) {
+        return
+      }
       var elementValue = elementInfo.elementValue
       data.text = elementValue
       var wasFilterRemoved = elementInfo.wasFilterRemoved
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -83,6 +86,27 @@
...
           break
+
+        case 'date': {
+          // The GOV.UK Design System date input consists of three grouped but separate fields (day,
+          // month, year). We want to fire a single event when all three fields are filled in to
+          // avoid firing excessive events.
+          const inputs = Array.from(
+            eventTarget.closest('.govuk-date-input').querySelectorAll('input')
+          )
+          const allInputsSet = inputs.every((input) => input.value)
```
kevindew, line 97:
```suggestion
          const allInputsSet = inputs.every(input => input.value)
```

Nit: parenthesis aren't required for one argument - I've had a few reviews by frontenders here that are keen that we don't use these parentheses unnecessarily, so sharing the nag

kevindew (commented):
Looks good, just a few minor things

app/assets/javascripts/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -83,6 +86,27 @@
             wasFilterRemoved = true
           }
           break
+
+        case 'date': {
```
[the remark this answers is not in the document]
csutter: Little bit eccentric yeah – Standard won't let you define new variables inside a case without an explicit scope (as it would leak otherwise), so the `const inputs` below would fail linting.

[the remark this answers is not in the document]
kevindew: oh that's a bit annoying! thanks for explaining - fine to leave in.

docs/analytics-ga4/ga4-finder-tracker.md
```diff
@@ -71,10 +71,13 @@ If the filter `<select>` element changes and is set to their _first option_, we
 
 Radio buttons have `data-ga4-change-category="update-filter radio"` on them. If the changed radio button is not the _first radio button_ (i.e. the default selection), then, the `update-filter` event will build an "Filter added" GTM Object. If the changed radio button is the _first radio button_, it will build a "Filter removed" GTM Object. This is because the first radio button in a list is typically the default "All XYZ" filter state. The value we grab for a radio button is the user friendly label associated with the `<input>`.
 
-### Date filters (`<input type="text">`)
+### Date filters with single input field (`<input type="text">`)
```
kevindew, line 74: I've no qualms with you using "Legacy" on the description of these if that is helpful in explaining them - they'll always be legacy from this point until they're removed.

kevindew (approved):
Nice one, thanks - feel free to change the title in the doc if you prefer it to have "Legacy" in it.

docs/analytics-ga4/ga4-finder-tracker.md
csutter: Decided to forgo "legacy" as we are still using them in some contexts and we don't know if/when we'll get rid of them!

PR #3522, opened 2024-10-21
All content finder: Enable parent breadcrumb

csutter:
This implements the parent breadcrumb feature analogous to the existing finders: if specific breadcrumbs are set, they are rendered instead of the regular contextual breadcrumbs.

### DIFF
--- a/app/views/finders/show_all_content_finder.html.erb
+++ b/app/views/finders/show_all_content_finder.html.erb
@@ -14,7 +14,16 @@
 <% end %>
 <% content_for :meta_title, content_item.title %>
 
-<%= render 'govuk_publishing_components/components/contextual_breadcrumbs', content_item: content_item.as_hash %>
+<% if @breadcrumbs.breadcrumbs %>
+  <%= render 'govuk_publishing_components/components/breadcrumbs', {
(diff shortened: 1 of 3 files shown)


kevindew (approved):

app/views/finders/show_all_content_finder.html.erb
```diff
@@ -14,7 +14,16 @@
 <% end %>
 <% content_for :meta_title, content_item.title %>
 
-<%= render 'govuk_publishing_components/components/contextual_breadcrumbs', content_item: content_item.as_hash %>
+<% if @breadcrumbs.breadcrumbs %>
```
kevindew, line 17: Just a thought after approving. Is @breadcrumbs always an object? Just wondered if this could return nil and you get a NoMethodError

csutter: Yes - it comes from [`FindersController`](https://github.com/alphagov/finder-frontend/blob/main/app/controllers/finders_controller.rb#L162) and is always set. Just the return value of `#breadcrumbs` may be nil which is a bit awkward, but I didn't want to start refactoring that as I'd probably go down a bit of a rabbithole!

PR #3530, opened 2024-10-23
All content finder: Send `search` event on submit

csutter:
This currently happens whenever the user changes the text in the keywords field, whether they then submit the form or not (because they legacy UI would run a search with JS on a change event).

This removes the change tracker from the keywords field, choosing instead to send a `search` event on form submit - but only if the user has actually changed their keywords since the last search/page load (which is what we want to track with this event).


kevindew (approved):
Looks good, just a small suggestion to avoid lots of indentation

PR #3533, opened 2024-10-24
SortFacet: Add `#sentence_fragment`

csutter:
This is required for compatibility with the other ("real") facets when rendering the page as JSON or Atom.

We can safely return nil though as it's just called but not actually used in the response.


kevindew (approved):
May not actually need a test

PR #3538, opened 2024-10-28
Manuals registry: Include HMRC manuals

csutter:
For some reason, these aren't fetched, which means that searches filtered by HMRC manuals don't show a filter tag.

## Before
image

## After
image


kevindew (approved):
Nice catch

PR #3545, opened 2024-11-01
Use new Publishing Components GA4 search tracker

csutter:
This allows us to remove the custom search tracking from the all content finder Javascript module.


leenagupte (approved):
I love a bit of JS deletion 🎉

PR #3548, opened 2024-11-04
Clean up Cucumber feature test suite

csutter:
The feature test suite has a number of steps and helper code that aren't actually used anymore.

This PR removes those, and also renames `site_search_steps` to `all_content_finder_steps` to be in line with the feature name.


leenagupte (approved):
I love a good deletion. 🧹 🥇

PR #3553, opened 2024-11-07
All content finder: Minor design/content tweaks

csutter:
This PR updates the all content finder to:
- use the `short_name` for displaying date facet names (slightly counterintuitively, the `short_name` for this field is actually _longer_ than the `name` 😵‍💫)
- slightly tweaks the spacing between the chevron and text on the "Filter and sort" panel button

### Before
image

### After
image

### DIFF
--- a/app/models/date_facet.rb
+++ b/app/models/date_facet.rb
@@ -41,7 +41,7 @@ class DateFacet < FilterableFacet
     present_values.map do |type, date|
       preposition = preposition_mappings[type]
       {
-        name: "#{name} #{preposition}",
+        name: "#{short_name} #{preposition}",
         label: date.date.strftime("%e %B %Y"),
         query_params: { key => { type => date.original_input } },
       }
--- a/features/fixtures/all_content.json
+++ b/features/fixtures/all_content.json
@@ -8,6 +8,15 @@
     "document_noun": "result",
     "facets": [
       {
+        "display_as_result_metadata": true,
+        "filterable": true,
+        "key": "public_timestamp",
+        "name": "Date",
+        "preposition": "Updated",
+        "short_name": "Updated",
+        "type": "date"
+      },
+      {
         "display_as_result_metadata": false,
         "filter_key": "all_part_of_taxonomy_tree",
         "filterable": true,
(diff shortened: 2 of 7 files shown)


app/models/date_facet.rb
```diff
@@ -41,7 +41,7 @@ def applied_filters
     present_values.map do |type, date|
       preposition = preposition_mappings[type]
       {
-        name: "#{name} #{preposition}",
+        name: "#{short_name} #{preposition}",
```
leenagupte, line 44: Why does the `short_name` need to be updated here and in the view? I assumed that seeing `short_name` in the "active filters" section on the page was just presentational?  Or is `name` the key and they have to match?

features/fixtures/all_content.json
```diff
@@ -141,30 +141,30 @@
...
         "api_path": "/api/content/search/all",
-        "api_url": "https://www.integration.publishing.service.gov.uk/api/content/search/all",
+        "api_url": "https://www.gov.uk/api/content/search/all",
         "base_path": "/search/all",
         "content_id": "dd395436-9b40-41f3-8157-740a453ac972",
         "document_type": "finder",
         "links": {},
         "locale": "en",
-        "public_updated_at": "2024-09-20T09:17:24Z",
+        "public_updated_at": "2024-10-24T11:19:57Z",
```
leenagupte, line 150: It's odd that the dates needed to be updated.

leenagupte (commented):
I looks good 🎉 . I just have one inline question.

app/models/date_facet.rb
csutter: Good question – it's because this and the view control separate things.

This one in `#applied_filters` is for the filter tags:
image

... and the one in the view is for the input field legend:
image

leenagupte: Thank you. I missed that both are being updated.

leenagupte (approved):

features/fixtures/all_content.json
csutter: That's just a side effect of importing the live content item – I think it's probably helpful to leave the date changes in even though they don't matter, because that way in the future someone might be able to tell when the fixture was last updated!

PR #3554, opened 2024-11-07
Amend Github PR template(s)

csutter:
- Remove separate, Dependabot-only `docs/pull_request_template.md`
- Prune example search pages for sense check and add to regular template


KludgeKML (approved):
LGTM!

PR #3555, opened 2024-11-07
Set up `SearchAutocomplete` AB test

csutter:
This sets up an AB test for using the new `search_with_autocomplete` component on the all content finder.

> [!NOTE]
> This AB test is live, but currently at 100% `Z` (baseline). We'll be using the AB test as a feature flag for final testing in production before sending any traffic to the control (`A`) and experiment (`B`) variants.

- Add GOV.UK Publishing Components Javascript for the `search_with_autocomplete` component
- Set up the usual AB testing helper module
- Use component if and only if AB test in experiment variant (`B`)


leenagupte (approved):
Looks good 🎉

PR #3561, opened 2024-11-14
Add and use API endpoint for search suggestions

csutter:
This returns suggestions from the (internal) Search API v2, for
consumption by the autocomplete component soon to be used across GOV.UK for site search.

- Update `gds-api-adapters` to get access to new `autocomplete` method
  for Search API v2
- Add basic API endpoint controller and route proxying to Search API v2
- Reconfigure autocomplete setup to use v2 `source_key` and the new
  endpoint's path (with an override option for local development)
- Update `govuk_publishing_components` for autocomplete tracking
  improvements

### DIFF
--- a/app/controllers/api/autocompletes_controller.rb
+++ b/app/controllers/api/autocompletes_controller.rb
@@ -1,0 +1,14 @@
+class Api::AutocompletesController < ApplicationController
+  def index
+    render json: autocomplete_response
+  end
+
+private
+
+  def autocomplete_response
+    Services
+      .search_api_v2
+      .autocomplete(params.require(:q))
+      .to_hash
+  end
+end
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -62,7 +62,10 @@ class FindersController < ApplicationController
     @pagination = pagination_presenter
     @spelling_suggestion_presenter = spelling_suggestion_presenter
     @ab_test_search_component = use_autocomplete? ? "search_with_autocomplete" : "search"
-    @autocomplete_url = ENV.fetch("SEARCH_AUTOCOMPLETE_API_URL", "")
+    @autocomplete_source_url = ENV.fetch(
+      "SEARCH_AUTOCOMPLETE_API_URL",
+      helpers.absolute_url_for(api_search_autocomplete_path(format: :json)),
+    )
   end
 
 private
--- a/spec/requests/api/autocomplete_spec.rb
+++ b/spec/requests/api/autocomplete_spec.rb
@@ -1,0 +1,26 @@
+require "spec_helper"
+
+RSpec.describe "autocomplete API", type: :request do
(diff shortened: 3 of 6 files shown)


app/controllers/finders_controller.rb
```diff
@@ -62,7 +62,10 @@ def show_page_variables
     @pagination = pagination_presenter
     @spelling_suggestion_presenter = spelling_suggestion_presenter
     @ab_test_search_component = use_autocomplete? ? "search_with_autocomplete" : "search"
-    @autocomplete_url = ENV.fetch("SEARCH_AUTOCOMPLETE_API_URL", "")
+    @autocomplete_source_url = ENV.fetch(
+      "SEARCH_AUTOCOMPLETE_API_URL",
+      URI.join(Plek.new.website_root, api_search_autocomplete_path(format: :json)).to_s,
```
kevindew, line 67: As this route is generated at request time I imagine we don't actually need to use Plek and can instead do: `api_search_autocomplete_url(format: :json)` which I expect would give it increased environment portability.

Having done a search through the code it looks like people have been scared off of using it before (scarred by broken proxy headers?) though as I can't find `_url` usage, but I did find a helper you could use: https://github.com/alphagov/finder-frontend/blob/d185ae1ec9421a755dbc0b0167cc9fddd2e98974/app/helpers/application_helper.rb#L2-L4

app/controllers/api/autocompletes_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  def index
+    render json: autocomplete_response
+  end
+
+private
+
+  def autocomplete_response
+    Services
+      .search_api_v2
+      .autocomplete(params.require(:q))
```
kevindew, line 11: It feels like we should add a request test that asserts that we get a bad_request response if there isn't a q parameter.

spec/requests/api/autocomplete_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  before do
+    allow(Services).to receive(:search_api_v2).and_return(search_api_v2)
+  end
+
+  it "returns suggestions from Search API v2" do
+    get "/api/search/autocomplete?q=loving+him+was"
+
+    expect(search_api_v2).to have_received(:autocomplete).with("loving him was")
+    expect(response).to be_successful
+    expect(JSON.parse(response.body)).to eq("suggestions" => %w[blue grey red])
```
kevindew, line 18:
```suggestion
    expect(JSON.parse(response.body)).to eq("suggestions" => suggestions)
```

since we've already got the variable defined.

kevindew (commented):
Looks great, just a couple of small suggestions

app/controllers/finders_controller.rb
csutter: There is that, however we'd have to amend the autocomplete component to accept relative URLs (as Javascript's `URL` bizarrely can't deal with them by default).

I'd be happy with that, it's just a tradeoff to make!

kevindew (approved):
Nice one

PR #3564, opened 2024-11-14
Autocomplete: Use URL helper instead of `absolute_url_for`

csutter:
This uses Rails's standard URL helper instead of the custom one in the app to establish the absolute URL for the autocomplete endpoint.


kevindew (approved):
Nice one 👍 

Based off what we found I'll just leave a comment for prosperity that if we find there's a problem with this (edge case we didn't consider with proxy headers) we could resolve the quoted string problem with: `absolute_url_for(api_search_autocomplete_path(format: :json)).to_s`

PR #3571, opened 2024-11-22
Add CORS configuration for autocomplete API

csutter:
Update from @kevindew 

Since the original was written it was also discovered that this
configuration would be needed for more than just development
environments and would actually be needed in production. The situation
where this is needed is CSV previews [2] which are GOV.UK pages with the
layout_super_navigation_header component hosted on the
assets.publishing.service.gov.uk.

In order to demonstrate CORS taking effect requests need to be provided
with an origin header e.g:

```
➜  ~ curl -Is -H "Origin: https://www.gov.uk/" \
  [http://127.0.0.1:3062/api/search/autocomplete.json\?q\=test](http://127.0.0.1:3062/api/search/autocomplete.json/?q\=test) | grep access-control
access-control-allow-origin: https://www.gov.uk/
access-control-allow-methods: GET
access-control-expose-headers:
access-control-max-age: 7200
```

An absence of any access-control-* headers indicates a CORS fail and in
a browser a request will be blocked e.g:

```
➜  ~ curl -Is -H "Origin: https://example.com/" \
  [http://127.0.0.1:3062/api/search/autocomplete.json\?q\=test](http://127.0.0.1:3062/api/search/autocomplete.json/?q\=test) | grep access-control
```

[1]: https://github.com/alphagov/govuk_publishing_components/blob/171e814b327bcfa0f2437fff0514ff086e31c96b/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb#L334
[2]: https://assets.publishing.service.gov.uk/media/663ca4da8603389a07a6d2f8/Malpractice_in_VTQ_-_Example_CSV_File.csv/preview
---

Allow the autocomplete API to be accessed from any GOV.UK domain, including non-production ones. This lets us use the API in local     development "live" stacks as well as the GOV.UK Publishing Components     guide.


spec/requests/api/autocomplete_spec.rb
```diff
@@ -23,4 +24,31 @@
...
+        get "/api/search/autocomplete", params:, headers: { Origin: allowed_host }
+
+        expect(response.headers.to_h).to include({
+          "access-control-allow-origin" => allowed_host,
+          "access-control-allow-methods" => "GET",
+        })
+      end
+    end
+
+    it "returns CORS headers when there is a format extension on the path" do
```
jackbot, line 40: The spec description says "a format extension on the path" but I'm not seeing one in the action `get` request, am I missing something here?

config/initializers/cors.rb
```diff
@@ -0,0 +1,14 @@
+# Be sure to restart your server when you modify this file.
+
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  # Allow the autocomplete API to be accessed from any GOV.UK domain, including non-production ones.
+  # This allows autocomplete to be available on CSV previews, which are hosted on
+  # assets.publishing.service.gov.uk. This allows allows for local development usage.
```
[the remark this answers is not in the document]
kevindew: Oops thanks. Amended.

spec/requests/api/autocomplete_spec.rb
kevindew: Whoops, have sorted now

jackbot (approved):

PR #3574, opened 2024-11-25
Add GOV.UK Chat promo banner to business finance support finder

davidgisbey:
## Description

We want to bring more users to the GOV.UK Chat app, so we'd like to render a promo component on one specific page: https://www.gov.uk/business-finance-support.

This follows a similar approach to that taken for [Government Frontend](https://github.com/alphagov/government-frontend/pull/3303) and [Collections](https://github.com/alphagov/collections/pull/3742) of adding a helper, modifying views, adding an env var to govuk-helm-charts and having appropriate tests.

This follows the pattern established in [Collections](https://github.com/alphagov/collections/pull/3742) and [Government Frontend](https://github.com/alphagov/government-frontend/pull/3303)

## Screenshots 

image

image

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -13,7 +13,7 @@ $app-colour-true-black: #000000;
   max-width: 100%;
 }
 
-.related-links {
+.title-side {
   margin-top: govuk-spacing(8);
   margin-bottom: govuk-spacing(8);
 }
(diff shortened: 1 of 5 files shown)


davidgisbey:
After [some discussion](https://gds.slack.com/archives/C06HN22CHUK/p1732613214506099) we're not going to add this to finder_frontend for now as it's only 40k page views out of 4 million and the design would need some thought.

We will address this if we want to add it to more finders down the line

PR #3588, opened 2024-12-03
Exclude top margin on intervention on taxon page with inverse design 

hannalaakso:
## What
Fix a bug with an overlapping margin caused by the user research banner on taxon pages when the page has the relevant query string eg. /search/guidance-and-regulation?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0 This fix simply excludes the top margin on the inverse design (instead of doing something more complicated like adding a wrapper for the intervention banner which could have the blue background colour which would enable the top margin also on the inverse design). This is because the taxon inverse design got added for some historical page migration and we'll probably look to remove it at some point (but as of a couple of months ago, we still weren't confident we can do it).

### Before

Screenshot 2024-12-03 at 11 33 48

### After

Screenshot 2024-12-03 at 11 32 08

### DIFF
--- a/app/views/finders/_show_header.html.erb
+++ b/app/views/finders/_show_header.html.erb
@@ -1,14 +1,14 @@
 <div class="govuk-width-container">
-  <div class="govuk-!-margin-top-4">
-    <% if content_item.has_user_research_banner? %>
-        <%= render "govuk_publishing_components/components/intervention", {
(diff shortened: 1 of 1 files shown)


hannako (approved):
Thank you Hanna!

MartinJJones (approved):

PR #3623, opened 2025-01-08
Adapt CORS policy for autocomplete_spec

csutter:
This updates the CORS policy to be more relaxed about GOV.UK subdomains, as well as allow Heroku preview apps.

This adds a negligible risk that third-party Heroku hosted apps (as well as non-GDS operated GOV.UK subdomains) could in theory use the autocomplete endpoint from Javascript, but we don't consider this to have enough impact or likelihood to be worth mitigating against.


kevindew (approved):
Looks good 👍

PR #3624, opened 2025-01-08
Remove "disable autocomplete" feature flag

csutter:
We are centralising the enabling/disabling of autocomplete in Search API v2 so we don't need individual feature flags across several apps.

- Use search_with_autocomplete component directly
- Remove ApplicationHelper#search_component

Depends on: https://github.com/alphagov/search-api-v2/pull/375/


kevindew (approved):
Nice one

PR #3627, opened 2025-01-15
Derive search result metadata labels from allowed values on facet where present

ryanb-gds:
For many document types, the values indexed in Elasticsearch for its metadata fields are the kebab case value from the select option value in the Publishing app.

These values are not suitable for presenting on GOV.UK, so there is a configuration in each content type defined in Search API called `expanded_search_result_fields`. This configuration is used to replace the kebab case value with a hash containing the original value and the human friendly label it relates to in the Search API query results.

Maintaining this configuration is painful, particularly for specialist document types, because it means that every time a department wants to add, change or remove an option on a facet (a frequent occurrence), we have to update the configuration in both the Publishing tool and in Search API.

However, given the label provided in the expanded search result fields configuration never varies* from the label present in the `allowed_values` attribute on the facet configuration, we may as well fetch the label from there and remove the `expanded_search_result_fields` from Search API, which will reduce the effort required from Publishing teams to maintain Specialist Finders.

This commit extracts the appropriate metadata labels from the `allowed_values` configuration on the facet if one is provided. It will work both for search results that have had the field expansion applied, and those that haven't, so should be safe to release before the removal of the configuration. It also preserves the existing behaviour for facets that do not set `allowed_values`.

\* as far as we are aware having checked the finder schemas - it would be strange for users if it did

https://trello.com/c/PZmDZD6l

- https://finder-frontend-pr-3627.herokuapp.com/search/all
- https://finder-frontend-pr-3627.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-3627.herokuapp.com/drug-device-alerts

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -68,10 +68,10 @@ private
     metadata_facets(facets).select { |f| f.type == "date" }.map(&:key)
   end
 
-  def text_metadata_keys(facets)
-    keys = metadata_facets(facets).select { |f| f.type == "text" }.map(&:key)
-    keys.reject do |key|
-      key == "organisations" && is_mainstream_content?
+  def text_metadata_facets(facets)
+    facets = metadata_facets(facets).select { |f| f.type == "text" }
+    facets.reject do |facet|
+      facet.key == "organisations" && is_mainstream_content?
     end
   end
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM, but happy to wait for a finder-frontend maintainer to approve 👍

spec/models/document_spec.rb
```diff
@@ -65,6 +65,26 @@
         [FactoryBot.build(:option_select_facet, key: "a_filter_key")]
       end
 
+      describe "The document is not tagged to any values of the facet filter key" do
```
KludgeKML, line 68: This feels more like it should be structured like:
```
describe "#metadata" do
   context "when the document is not tagged to any values of the facet filter key" do.... (etc)
```

app/models/document.rb
```diff
@@ -100,38 +100,47 @@ def build_date_metadata(key)
...
-
-  def text_labels_for(key)
-    labels = Array(document_hash.fetch(key, []))
-               .map { |label| get_metadata_label(key, label) }
-               .select(&:present?)
-
-    if key == "organisations"
-      labels = labels.sort_by do |label|
-        label.sub("Closed organisation: ", "ZZ").upcase
+    metadata_labels = {}.tap do |result|
```
KludgeKML, line 103: Hmm. I have a little feel about this, which is that this construct is worse than just:
```
   metadata_labels = {}
   text_medadata_facets.each do |facet|
     ...etc...
     metadata_labels[facet.key] = if facet.allowed_values.empty?
     ...etc...
```
I may be wrong, but the tap doesn't really add anything here?

ryanb-gds: You're quite right. Updated

spec/models/document_spec.rb
ryanb-gds: I initially followed the convention of the existing tests here, so I've added a second commit where I apply this naming style for the whole file. Hope that works

KludgeKML: Might be a good idea for us to look at the tests (but that's a long way down our priority list at the moment). Thanks, anyway!

KludgeKML (approved):
LGTM!

PR #3629, opened 2025-01-16
Revert "Derive search result metadata labels from allowed values on facet where present"

ryanb-gds:
Reverts alphagov/finder-frontend#3627 as it caused an error in production.

Link to Sentry error: https://govuk.sentry.io/issues/6223168052/events/dce67ad450c648b9b91ce29a331cfb38/


KludgeKML (approved):
LGTM

PR #3630, opened 2025-01-16
Reinstate metadata labels from allowed values

ryanb-gds:
This reinstates the reverted PR #3627, which did not cover the case where the document value for the facet was a single string and the facet had a set of allowed values, causing the Service Standard finder to crash. This PR includes a fix for that scenario.

To test the effectiveness of the fix visit [https://finder-frontend-pr-3630.herokuapp.com/service-standard-reports](https://finder-frontend-pr-3630.herokuapp.com/service-standard-reports) 

## Some search page examples to sense check for regressions:
- https://finder-frontend-pr-3630.herokuapp.com/search/all
- https://finder-frontend-pr-3630.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-3630.herokuapp.com/drug-device-alerts

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -68,10 +68,10 @@ private
     metadata_facets(facets).select { |f| f.type == "date" }.map(&:key)
   end
 
-  def text_metadata_keys(facets)
-    keys = metadata_facets(facets).select { |f| f.type == "text" }.map(&:key)
-    keys.reject do |key|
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM

PR #3664, opened 2025-02-11
Allow previewing serving configurations on Search API v2

csutter:
Search API v2 now supports getting results through different serving configs [1]. This allows us to (among other things) have a "preview" serving config on which we can test out different boosting options.

We considered having a preview feature in Search Admin, but having the ability here in Finder Frontend instead allows administrative users to test different configurations more comprehensively, for example by using filters and sorting, without having to reimplement half of Finder Frontend in another app.

Having some (non-sensitive) debugging functionality in Finder Frontend is also an existing pattern (see `debug` parameter for Search API v1 searches).

- Pass through the API used to `Search::QueryBuilder`
- Send `serving_config` parameter to Search API v2 if that is currently in use and the `debug_serving_config` parameter is in the request
- Add a banner to the UI to show when a non-default serving configuration is applied, and allow removing it
- Add a hidden field for the overridden serving config if given, so it persists across searches

[1]: https://github.com/alphagov/search-api-v2/pull/385

## Screenshots

image

### DIFF
--- a/app/lib/search/query.rb
+++ b/app/lib/search/query.rb
@@ -86,6 +86,7 @@ module Search
         finder_content_item: content_item,
         params: filter_params,
         override_sort_for_feed:,
+        use_v2_api: use_v2_api?,
       ).call
 
       if use_v2_api?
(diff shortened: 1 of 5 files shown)


leenagupte (approved):
Looks good! 🎉 

I like the use of the `notice`. Hopefully that'll make it easier to spot any typos in the serving config name 😅

PR #3679, opened 2025-02-25
Add support for Nested facets

lauraghiorghisor-tw:
These changes implement a custom NestedFacet, which allows us to filter based on a main facet - subfacet relationship. 

We do send some additional data in the raw facets, in order to be able to represent the relationship between the facets, but otherwise the behaviour is contained in the NestedFacet model, and the respective nested facets are stand-alone facets.

JS from the prototype here https://github.com/alphagov/nested-facet-prototype by @gtvj. 

[Trello](https://trello.com/c/skH2B2fH/3373-nested-facets-finder-frontend)

### No JS
- The two selects, main and sub, render the entirety of the possible options. The subfacet labels are dasherised to help distinguish between duplicate values.
- !Note the full width of the select and left alignment of the facet names against the other facet types.
> Screenshot 2025-02-26 at 16 37 17

> Screenshot 2025-02-25 at 14 04 23

> Screenshot 2025-02-25 at 14 04 42

### JS enabled

- Compared to the other facets (slightly out of place in that it does not use the accordion pattern).

> Screenshot 2025-02-26 at 16 37 42

- The subfacet options are related to the selection in the mainfacet only. The labels are not dasherised.

> Screenshot 2025-02-25 at 14 10 31

### Filtering - no search results found
- The metadata tags at the top are correctly rendering the prepositions

> Screenshot 2025-02-25 at 14 12 07

### Filtering - with a search result
- The metadata tags for the document are correctly rendering the short names of the facets
- The URL shows the correct query: `http://finder-frontend.dev.gov.uk/trademark-decisions?trademark_decision_grounds_section=section-5-1-5-2-and-5-3-earlier-trade-marks&trademark_decision_grounds_sub_section=section-5-1-5-2-and-5-3-earlier-trade-marks-not-applicable`

> Screenshot 2025-02-25 at 14 12 17

For reference, a document created by Specialist Publisher would be tagged like this:
> Screenshot 2025-02-25 at 14 09 25

### DIFF
--- a/app/assets/javascripts/modules/nested-facets.js
+++ b/app/assets/javascripts/modules/nested-facets.js
@@ -1,0 +1,77 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  function NestedFacets (module) {
+    this.module = module
+    this.mainFacet = document.querySelector('#' + this.module.getAttribute('data-main-facet-id'))
+    this.subFacet = document.querySelector('#' + this.module.getAttribute('data-sub-facet-id'))
+    this.options = this.instantiateOptions()
+  }
+
+  NestedFacets.prototype.init = function () {
+    this.updateFacets()
+    this.module.addEventListener('change', () => {
+      this.updateFacets()
+    })
+  }
+
+  NestedFacets.prototype.updateFacets = function () {
+    this.setSubFacetSelectDisabledState()
+    this.resetSubFacetValue()
+    this.populateSubFacets()
+  }
+
+  NestedFacets.prototype.setSubFacetSelectDisabledState = function () {
+    const mainFacetSelected = !!this.mainFacet.value
+    if (mainFacetSelected) {
+      this.subFacet.removeAttribute('disabled')
+    } else {
+      this.subFacet.setAttribute('disabled', true)
+    }
+  }
+
+  NestedFacets.prototype.populateSubFacets = function () {
+    const subFacetOptionsForSelectedMain = this.options[this.mainFacet.value]
+    const subFacet = this.subFacet
+    const subFacetOptions = subFacet.querySelectorAll('option')
+
+    for (let o = 0; o < subFacetOptions.length; o++) {
+      if (subFacetOptions[o].value) {
+        subFacetOptions[o].parentNode.removeChild(subFacetOptions[o])
+      }
+    }
+    if (subFacetOptionsForSelectedMain) {
+      for (let i = 0; i < subFacetOptionsForSelectedMain.length; i++) {
+        subFacet.appendChild(subFacetOptionsForSelectedMain[i])
+      }
+    }
+  }
+
+  NestedFacets.prototype.instantiateOptions = function () {
+    const options = {}
+    const optionElements = this.subFacet.querySelectorAll('option')
+
+    for (let o = 0; o < optionElements.length; o++) {
+      const mainFacetValue = optionElements[o].getAttribute('data-main-facet-value')
+
+      options[mainFacetValue] = options[mainFacetValue] || []
+      options[mainFacetValue].push(optionElements[o])
+    }
+    return options
+  }
+
+  NestedFacets.prototype.resetSubFacetValue = function () {
+    const selected = this.subFacet.options[this.subFacet.selectedIndex]
+    const mainFacetValue = this.mainFacet.value
+    const isOrphanedChild = selected && selected.getAttribute('data-main-facet-value') !== mainFacetValue
+
+    if (isOrphanedChild) {
+      this.subFacet.value = ''
+    }
+  }
+
+  Modules.NestedFacets = NestedFacets
+})(window.GOVUK.Modules)
--- a/app/assets/javascripts/modules/nested-facets.js
+++ b/app/assets/javascripts/modules/nested-facets.js
@@ -57,6 +57,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     for (let o = 0; o < optionElements.length; o++) {
       const mainFacetValue = optionElements[o].getAttribute('data-main-facet-value')
 
+      const mainFacetLabelPrefix = optionElements[o].getAttribute('data-main-facet-label') + ' - '
+      optionElements[o].text = optionElements[o].text.replace(mainFacetLabelPrefix, '')
+
       options[mainFacetValue] = options[mainFacetValue] || []
       options[mainFacetValue].push(optionElements[o])
     }
--- a/app/lib/facets_builder.rb
+++ b/app/lib/facets_builder.rb
@@ -23,7 +23,9 @@ private
   def build_facet(facet_hash)
     if facet_hash["filterable"]
       case facet_hash["type"]
-      when "text", "content_id"
+      when "text"
+        facet_hash["nested_facet"] ? NestedFacet.new(facet_hash, value_hash[facet_hash["key"]]) : OptionSelectFacet.new(facet_hash, value_hash[facet_hash["key"]])
+      when "content_id"
         OptionSelectFacet.new(facet_hash, value_hash[facet_hash["key"]])
       when "topical"
         TopicalFacet.new(facet_hash, value_hash[facet_hash["key"]])
--- a/app/models/nested_facet.rb
+++ b/app/models/nested_facet.rb
@@ -1,0 +1,12 @@
+class NestedFacet < OptionSelectFacet
+  attr_accessor :sub_facet_key
+
+  def initialize(facet, values)
+    @sub_facet_key = facet["sub_facet_key"]
+    super
+  end
+
+  def is_main_facet?
+    sub_facet_key.present?
+  end
+end
--- a/app/models/nested_facet.rb
+++ b/app/models/nested_facet.rb
@@ -6,7 +6,22 @@ class NestedFacet < OptionSelectFacet
     super
   end
 
+  def facet_options
+    default_selection_options = [{ text: "All #{pluralized_facet_short_name}", value: "" }]
+    allowed_values.inject(default_selection_options) do |options, allowed_value|
+      option = { text: allowed_value["label"], value: allowed_value["value"] }
+      option.merge!(data_attributes: { main_facet_value: allowed_value["main_facet_value"] }) unless is_main_facet?
+      options << option
+    end
+  end
+
   def is_main_facet?
     sub_facet_key.present?
+  end
+
+private
+
+  def pluralized_facet_short_name
+    (short_name || name).downcase.pluralize
   end
 end
--- a/app/models/nested_facet.rb
+++ b/app/models/nested_facet.rb
@@ -12,6 +12,7 @@ class NestedFacet < OptionSelectFacet
       option = {
         text: facet_text(allowed_value),
         value: allowed_value["value"],
+        selected: selected_values.include?(allowed_value),
       }
       option.merge!(data_attributes: { main_facet_value: allowed_value["main_facet_value"], main_facet_label: allowed_value["main_facet_label"] }) unless is_main_facet?
       options << option
--- a/app/models/nested_facet.rb
+++ b/app/models/nested_facet.rb
@@ -9,8 +9,11 @@ class NestedFacet < OptionSelectFacet
   def facet_options
     default_selection_options = [{ text: "All #{pluralized_facet_short_name}", value: "" }]
     allowed_values.inject(default_selection_options) do |options, allowed_value|
-      option = { text: allowed_value["label"], value: allowed_value["value"] }
-      option.merge!(data_attributes: { main_facet_value: allowed_value["main_facet_value"] }) unless is_main_facet?
+      option = {
+        text: facet_text(allowed_value),
+        value: allowed_value["value"],
+      }
+      option.merge!(data_attributes: { main_facet_value: allowed_value["main_facet_value"], main_facet_label: allowed_value["main_facet_label"] }) unless is_main_facet?
       options << option
     end
   end
@@ -20,6 +23,10 @@ class NestedFacet < OptionSelectFacet
   end
 
 private
+
+  def facet_text(value)
+    value["main_facet_label"] ? "#{value['main_facet_label']} - #{value['label']}" : value["label"]
+  end
 
   def pluralized_facet_short_name
     (short_name || name).downcase.pluralize
--- a/app/views/finders/_nested_facet.html.erb
+++ b/app/views/finders/_nested_facet.html.erb
@@ -10,7 +10,7 @@
 <% end %>
 
 <% if nested_facet.is_main_facet? %>
-  <div class="govuk-!-margin-top-3">
+  <div class="govuk-!-margin-top-3" data-module="nested-facets"  data-main-facet-id="<%= nested_facet.key %>" data-sub-facet-id="<%= nested_facet.sub_facet_key %>">
     <%= content %>
   </div>
 <% else %>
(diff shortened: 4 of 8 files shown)


app/models/nested_facet.rb
```diff
@@ -0,0 +1,27 @@
+class NestedFacet < OptionSelectFacet
```
lauraghiorghisor-tw, line 1: I did debate whether to inherit from something else, but this one seems to give us the behaviour we want in the facet tags (showing `preposition + tag` at the top of the search results).

app/lib/facets_builder.rb
```diff
@@ -23,7 +23,9 @@ def facet_hashes
   def build_facet(facet_hash)
     if facet_hash["filterable"]
       case facet_hash["type"]
-      when "text", "content_id"
+      when "text"
```
lauraghiorghisor-tw, line 26: I am not sure what the `content_id` type is about, but for our use case, we only ever care about type text.

app/views/finders/_nested_facet.erb
```diff
@@ -0,0 +1,17 @@
+<% add_gem_component_stylesheet("select") %>
+
+<% if nested_facet.is_main_facet %>
```
lauraghiorghisor-tw, line 3: Not the greatest way to wrap this in order to load the js module, suggestions welcome.

app/models/nested_facet.rb
```diff
@@ -3,7 +3,21 @@ class NestedFacet < OptionSelectFacet
 
   def initialize(facet, values)
     @sub_facet_key = facet["sub_facet_key"]
-    @is_main_facet = !!facet["sub_facet_key"]
+    @is_main_facet = !facet["sub_facet_key"].nil?
     super
   end
+
+  def facet_options
```
ChrisBAshton, line 10: Would be good to explain in the commit message what's happening in this commit. Quite a lot to follow otherwise.

app/views/finders/_nested_facet.erb
```diff
@@ -6,3 +6,7 @@
   full_width: true,
   options: nested_facet.facet_options
 } %>
+
+<% unless nested_facet.is_main_facet %>
```
ChrisBAshton, line 12: Any reason this shouldn't be fixed up to the previous commit?

Also commit message could explain why this addition is necessary.

app/views/finders/_nested_facet.erb
ChrisBAshton: Could make it more like

```erb
<% if nested_facet.is_main_facet %>
    <div class="js-nested-facets" data-module="nested-facets" data-attributes="<%= { main_facet_id: nested_facet.key, sub_facet_id: nested_facet.sub_facet_key }.to_json %>">
<% else %>
    <div>
```

Then you don't have to repeat the conditional later on for the closing `</div>`.

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,91 @@
+window.GOVUK = window.GOVUK || {}
```
ChrisBAshton, line 1: Typo in commit message "Add javasript to render nested facet values"

spec/lib/facets_builder_spec.rb
```diff
@@ -11,6 +11,52 @@
...
+  let(:hash_including_nested_facets) do
+    [
+      {
+        "filterable": true,
+        "type": "text",
+        "name": "Facet Name",
+        "key": "facet_key",
+        "preposition": "with",
+        "sub_facet_key": "test_sub_facet_key",
+        "sub_facet_name": "Test Sub Facet Name",
```
KludgeKML, line 23: Should these match the name and key of the sub-facet?

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,91 @@
...
+        subFacet.appendChild(facet[i])
+      }
+    }
+  }
+
+  NestedFacets.prototype.instantiateOptions = function instantiateOptions () {
+    const options = {}
+    const optionElements = this.$subFacet().querySelectorAll('option')
+
+    for (let o = 0; o < optionElements.length; o++) {
```
ChrisBAshton, line 68: I think using [JS .map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) is preferred (shorter, easier to read and reason about than an iterative for-loop).

Otherwise a quick sense check of the JS, looks OK to me 👍

ChrisBAshton (commented):
Looks good to me!

I'll let someone with more experience of finder-frontend (and GOVUK JS conventions) do the approval.

app/models/nested_facet.rb
```diff
@@ -0,0 +1,9 @@
+class NestedFacet < OptionSelectFacet
+  attr_accessor :is_main_facet, :sub_facet_key
+
+  def initialize(facet, values)
+    @sub_facet_key = facet["sub_facet_key"]
+    @is_main_facet = !!facet["sub_facet_key"]
```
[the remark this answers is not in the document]
KludgeKML: We could also rename `@is_main_facet` to `@main_facet?` potentially - that's more idiomatic.

app/models/nested_facet.rb
```diff
@@ -3,7 +3,21 @@ class NestedFacet < OptionSelectFacet
 
   def initialize(facet, values)
     @sub_facet_key = facet["sub_facet_key"]
-    @is_main_facet = !!facet["sub_facet_key"]
+    @is_main_facet = !facet["sub_facet_key"].nil?
```
KludgeKML, line 6: Ah, you've done this here, rather than squashing this change into the original commit.

app/models/nested_facet.rb
KludgeKML: Seconded.

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,94 @@
...
+(function (Modules) {
+  'use strict'
+
+  function NestedFacets (module) {
+    this.module = module
+    this.options = this.instantiateOptions()
+  }
+
+  NestedFacets.prototype.init = function () {
+    if (this.module) {
```
andysellick, line 13: There shouldn't be a need to check for the existence of `this.module` - if the module has been initialised, something will have been passed to it.

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,94 @@
...
+      this.update()
+      this.module.addEventListener('change', () => {
+        this.update()
+      })
+    }
+  }
+
+  NestedFacets.prototype.update = function updateFacets () {
```
andysellick, line 21: I don't think you need to name these functions twice (unless you're doing something specific with that). So all of these prototype calls can be of the form

```JavaScript
NestedFacets.prototype.update = function () {
```

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,94 @@
...
+    this.populateSubFacets()
+  }
+
+  NestedFacets.prototype.$mainFacets = function $mainFacets () {
+    const mainFacetId = JSON.parse(this.module.dataset.attributes).main_facet_id
+    return document.querySelector('#' + mainFacetId)
+  }
+
+  NestedFacets.prototype.$subFacet = function $subFacet () {
+    const subFacetId = JSON.parse(this.module.dataset.attributes).sub_facet_id
```
andysellick, line 33: Is it ever possible that `sub_facet_id` won't exist?

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,94 @@
...
+    return document.querySelector('#' + subFacetId)
+  }
+
+  NestedFacets.prototype.setSubFacetSelectDisabledState =
+    function setSubFacetSelectDisabledState () {
+      const selectedMainFacet = !!this.$mainFacets().value
+      if (!selectedMainFacet) {
```
andysellick, line 40: Would this be more readable if the if statement were reversed? i.e.

```JavaScript
const selectedMainFacet = !!this.$mainFacets().value
if (selectedMainFacet) {
  this.$subFacet().removeAttribute('disabled')
} else {
...
```

app/views/finders/_nested_facet.erb
andysellick, line 1: Does this file need to be named `_nested_facet.html.erb`? I'm not sure what difference it makes but that seems to be our convention.

andysellick (commented):
Have had a quick look, sorry rushing to fit this in between meetings. Happy to look in a bit more depth later this week.

spec/lib/facets_builder_spec.rb
lauraghiorghisor-tw: They should, yes. The test in this case only covers shallow expectations, but it's good to have the whole chunk of data for "documentation" purposes - and the keys should match.

app/views/finders/_nested_facet.erb
lauraghiorghisor-tw: It is purely for style - adding a separator only under the sub facet - the select govuk component does not come with one, unlike the other components used for the other facets. 
Will improve commit ✅

app/models/nested_facet.rb
lauraghiorghisor-tw: My bad, late rubocop run. Will rebase.

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  NestedFacets.prototype.update = function updateFacets () {
+    this.setSubFacetSelectDisabledState()
+    this.resetSubFacetValue()
+    this.populateSubFacets()
+  }
+
+  NestedFacets.prototype.$mainFacets = function $mainFacets () {
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Renamed ✅

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,94 @@
...
+
+  NestedFacets.prototype.setSubFacetSelectDisabledState =
+    function setSubFacetSelectDisabledState () {
+      const selectedMainFacet = !!this.$mainFacets().value
+      if (!selectedMainFacet) {
+        this.$subFacet().setAttribute('disabled', true)
+      } else {
+        this.$subFacet().removeAttribute('disabled')
+      }
+    }
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: I reran linter, seems alright?

app/assets/javascripts/modules/nested-facets.js
lauraghiorghisor-tw: I believe a map or forEach would be messed up by the child removal, which is why we're looping through the indices. Take it up with @gtvj 😅

app/assets/javascripts/modules/nested-facets.js
andysellick: Great, just wanted to check it wouldn't error.

app/assets/javascripts/modules/nested-facets.js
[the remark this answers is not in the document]
andysellick: Looks fine now 👍

app/models/nested_facet.rb
ChrisBAshton: Seems a sensible choice 🤷 I wouldn't have blinked twice if you hadn't called it out!

app/models/nested_facet.rb
```diff
@@ -6,7 +6,21 @@ def initialize(facet, values)
     super
   end
 
+  def facet_options
+    allowed_values.inject([{ text: "All #{pluralized_facet_short_name}", value: "" }]) do |options, allowed_value|
```
ChrisBAshton, line 10: I'd suggest pulling this out into a well named variable, e.g.

```suggestion
    all_option = { text: "All #{pluralized_facet_short_name}", value: "" }
    allowed_values.inject([all_option]) do |options, allowed_value|
```

app/views/finders/_nested_facet.html.erb
```diff
@@ -1,16 +1,16 @@
...
-      label: nested_facet.name,
-      full_width: true,
-      options: nested_facet.facet_options
-    } %>
+  <%= render "govuk_publishing_components/components/select", {
+    id: nested_facet.key,
+    label: nested_facet.name,
+    full_width: true,
+    options: nested_facet.facet_options
+  } %>
```
ChrisBAshton, line 9: Nitpick: Can this indentation change be fixed up to the commit where the code was introduced?

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -65,6 +65,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     for (let o = 0; o < optionElements.length; o++) {
       const mainFacetValue = optionElements[o].getAttribute('data-main-facet-value')
 
+      const mainFacetLabel = optionElements[o].getAttribute('data-main-facet-label') + ' - '
```
ChrisBAshton, line 68: Nitpick: I'd suggest a name of `mainFacetLabelPrefix` instead (better implies that there is a " - " appended to the end of it)

spec/models/nested_facet_spec.rb
```diff
@@ -131,5 +138,46 @@
         )
       end
     end
+
+    context "when there is a selection" do
```
ChrisBAshton, line 142: Thanks for the detailed explanation in the commit message 👍 

This is a good unit test. It feels like we may also want an integration test (perhaps somewhere like [finders_controller_spec](https://github.com/alphagov/finder-frontend/blob/main/spec/controllers/finders_controller_spec.rb)), to properly verify that a change in URL parameter is reflected in which options have `selected="selected"` set?

ChrisBAshton (approved):
Looks good ⭐ Some comments below, but none of them blocking.

app/views/finders/_nested_facet.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<% end %>
+
+<% if nested_facet.is_main_facet? %>
+  <div class="govuk-!-margin-top-3" data-module="nested-facets" data-attributes="<%= { main_facet_id: nested_facet.key, sub_facet_id: nested_facet.sub_facet_key }.to_json %>">
+    <%= content %>
+  </div>
+<% else %>
+  <div class="gem-c-option-select govuk-!-padding-bottom-0">
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: We needed a separator, thought I'd grab it from the class (all other facets seem to get it for free from the component library) - we can do a `<hr class="govuk-section-break govuk-section-break--visible">` instead 🤔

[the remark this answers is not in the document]
andysellick: I think that'd be better, yes 👍

app/assets/javascripts/modules/nested-facets.js
```diff
@@ -0,0 +1,88 @@
...
+
+  NestedFacets.prototype.updateFacets = function () {
+    this.setSubFacetSelectDisabledState()
+    this.resetSubFacetValue()
+    this.populateSubFacets()
+  }
+
+  NestedFacets.prototype.mainFacet = function () {
+    const mainFacetId = JSON.parse(this.module.dataset.attributes).main_facet_id
+    return document.querySelector('#' + mainFacetId)
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: I have tried it out, I believe I'll need to keep those two as functions, to get a fresh read of the element wherever we use the `mainFacet`, its value, and `subFacet`.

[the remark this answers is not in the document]
andysellick: Oh it needs to be read multiple times, not just on initialisation? Okay, fair enough then.

PR #3691, opened 2025-03-11
Set up site search freshness boost AB test

csutter:
, [Jira issue SCH-458](https://gov-uk.atlassian.net/browse/SCH-458)This sets up the upcoming `SearchFreshnessBoost` AB test for site search, which uses a different Vertex AI Search serving config through Search API v2.

For the `B` variant, this uses the `variant_search` serving config. Otherwise we'll continue to use the `default_search` serving config, but do so explicitly (rather than implicitly as up until now).

see also:
- [Trello card](https://trello.com/c/7vbzHU0K)
- https://github.com/alphagov/search-api-v2/pull/397
- https://github.com/alphagov/govuk-fastly/pull/142
- https://github.com/alphagov/govuk-infrastructure/pull/1798


leenagupte (approved):
Nice!

PR #3693, opened 2025-03-12
Remodel NestedFacet to behave more like a TaxonFacet

lauraghiorghisor-tw:
## Remodel NestedFacet to behave more like a TaxonFacet

This PR attempts to make the newly added `NestedFacet` a bit more consistent with the pre-existent, and somewhat similar, `TaxonFacet`. There are two behaviour points that are ultimately the goal:
- getting the applied filters/metadata tags at the top of the search results to behave similarly to taxon, i.e. visually represent that there is a logical unit of a nested facet rather than two distinct facets
- wrapping the nested facet in an expander/accordion behaviour, similar to how other facets behave.

From a technical point of view, this is acknowledging that Taxon was already there doing ✨ _similar but slightly different things_ ✨, and it's worth exploring it properly in the context of the nested facet implementation - as @andysellick rightly pointed out when reviewing the [previous PR here](https://github.com/alphagov/finder-frontend/pull/3679).

In the previous implementation, the `NestedFacet` was written up as two distinct facets of type "text", with some additional fields to keep track of which is the "main" and which is the "sub". The scope of that change was a lot smaller than if we implemented a custom facet, with most of the changes in the originating finder repo, "Specialist Publisher".

The view ( I mean .erb) that served each facet in the previous implementation did not lend itself to the accordion/expander behaviour (basically: I could not wrap the two selects in the expander div in the no-js implementation 😅), so I looked into making a single nested facet again. Taxon is a good example to follow, as it, too, has two selects/facets under the hood, but behaves like a "single" facet.

Nonetheless, there are differences between the two: `TaxonFacet` populates its options in a unique way, whilst `NestedFacet` will get its options from Content Store, and ultimately from the schema defined in Specialist Publisher. There are also differences in how the two are implemented in Search API. For `NestedFacet`, we have opted to maintain two facets both in Search API and Publishing API schemas, whilst `TaxonFacet` seemingly has its own custom rules in Search API, and likely send a unified query (from what I can tell).

The approach for this PR is to build a single `NestedFacet` from the raw "nested" data facet we get from Content Store, and in doing so, replicate desired behaviours from `TaxonFacet`, and follow its implementation as far as we can. To this end, the new implementation uses a new "type" - "nested" - to make it easier to carve out the custom logic further down the line.

Whilst further refactoring is possible - abstract out common behaviours between the two classes, or use inheritance - I believe it is too early to make that call. Further steps in the `NestedFacets` work will include expanding the selection to multiple options, which will diverge from taxon behaviour. The test coverage of nested behaviours in this and upcoming commits should also support this future behaviour expansion.

These changes are accompanied by corresponding changes in [Specialist Publisher](https://github.com/alphagov/specialist-publisher/pull/3049) and [Publishing API](https://github.com/alphagov/publishing-api/pull/3203).

## Screenshots

- Before (left) and after (right)
> Screenshot 2025-03-12 at 12 22 16

- For reference, this is a taxon example:
- Link [here](https://www.gov.uk/search/news-and-communications?level_one_taxon=ba951b09-5146-43be-87af-44075eac3ae9&level_two_taxon=58643944-9cc3-4f61-8805-4c1d090bf21d&order=updated-newest).
> Screenshot 2025-03-12 at 16 24 33

| Taxon | Nested |
|--------|--------|
| Collapsed with selection | |
| Screenshot 2025-03-12 at 16 30 06 | Screenshot 2025-03-12 at 09 28 33 | 
| Collapsed with no selection | |
| Screenshot 2025-03-12 at 16 39 32 | Screenshot 2025-03-12 at 09 29 20 |
| Expanded | |
| Screenshot 2025-03-12 at 16 25 34 | Screenshot 2025-03-12 at 09 28 44 |
| All selected | |
| Screenshot 2025-03-12 at 16 24 47 | Screenshot 2025-03-12 at 16 42 40 |
| Applied filters | |
| Screenshot 2025-03-12 at 16 43 40 | Screenshot 2025-03-12 at 16 43 54 |

## Links

[Trello](https://trello.com/c/EeZOA3iW/3534-spike-replicating-taxons-approach-for-nested-facets)
[Previous PR](https://github.com/alphagov/finder-frontend/pull/3679)
[Draft finder link where previous changes are visible](https://draft-origin.publishing.service.gov.uk/trademark-decisions)
Sister PRs: [Specialist Publisher](https://github.com/alphagov/specialist-publisher/pull/3049) and [Publishing API](https://github.com/alphagov/publishing-api/pull/3203)

### DIFF
--- a/app/lib/facets_builder.rb
+++ b/app/lib/facets_builder.rb
@@ -23,9 +23,7 @@ private
   def build_facet(facet_hash)
     if facet_hash["filterable"]
       case facet_hash["type"]
-      when "text"
-        facet_hash["nested_facet"] ? NestedFacet.new(facet_hash, value_hash[facet_hash["key"]]) : OptionSelectFacet.new(facet_hash, value_hash[facet_hash["key"]])
-      when "content_id"
+      when "text", "content_id"
         OptionSelectFacet.new(facet_hash, value_hash[facet_hash["key"]])
       when "topical"
         TopicalFacet.new(facet_hash, value_hash[facet_hash["key"]])
@@ -45,6 +43,8 @@ private
         RadioFacetForMultipleFilters.new(facet_hash, value_hash[facet_hash["key"]], ::Filters::ResearchAndStatsHashes.new.call)
       when "official_documents"
         RadioFacetForMultipleFilters.new(facet_hash, value_hash[facet_hash["key"]], ::Filters::OfficialDocumentsHashes.new.call)
+      when "nested"
+        NestedFacet.new(facet_hash, value_hash.slice(facet_hash["key"], facet_hash["sub_facet_key"]))
       else
         raise ArgumentError, "Unknown filterable facet type: #{facet_hash['type']}"
       end
@@ -66,15 +66,15 @@ private
 
   def allowed_values_for_facet_details(facet_key, facet_details)
     facet_details.fetch("options", {})
-      .map { |f| f.fetch("value", {}) }
-      .map { |value| present_facet_option(value, facet_key) }
-      .reject { |f| f["label"].blank? || f["value"].blank? }
+                 .map { |f| f.fetch("value", {}) }
+                 .map { |value| present_facet_option(value, facet_key) }
+                 .reject { |f| f["label"].blank? || f["value"].blank? }
   end
 
   def allowed_values_from_registry(facet_key)
     registries.all[facet_key].values
-      .map { |_, results| present_facet_option(results, facet_key) }
-      .reject { |f| f["label"].blank? || f["value"].blank? }
+              .map { |_, results| present_facet_option(results, facet_key) }
+              .reject { |f| f["label"].blank? || f["value"].blank? }
   end
 
   def present_facet_option(value, facet_key)
--- a/app/lib/filters/nested_filter.rb
+++ b/app/lib/filters/nested_filter.rb
@@ -1,0 +1,32 @@
+module Filters
+  class NestedFilter < Filter
+    def initialize(facet, params)
+      super
+
+      return if params.blank?
+
+      @main_facet_key = params["main_facet_key"]
+      @sub_facet_key = params["sub_facet_key"]
+      @main_facet_value = params[@main_facet_key]
+      @sub_facet_value = params[@sub_facet_key]
+    end
+
+    def query_hash
+      result = {}
+      result.merge!(@main_facet_key => [@main_facet_value]) if @main_facet_value
+      result.merge!(@sub_facet_key => [@sub_facet_value]) if @sub_facet_value
+
+      result
+    end
+
+  private
+
+    def value
+      @value ||= begin
+        return nil unless @main_facet_value || @sub_facet_value
+
+        [@main_facet_value, @sub_facet_value].compact
+      end
+    end
+  end
+end
--- a/app/lib/search/query_builder.rb
+++ b/app/lib/search/query_builder.rb
@@ -112,9 +112,13 @@ module Search
     end
 
     def metadata_fields
-      raw_facets.map do |f|
-        unfilterise(f["filter_key"] || f["key"])
-      end
+      raw_facets.map { |f|
+        if f["sub_facet_key"]
+          [unfilterise(f["filter_key"] || f["key"]), unfilterise(f["sub_facet_key"])]
+        else
+          unfilterise(f["filter_key"] || f["key"])
+        end
+      }.flatten
     end
 
     def raw_facets
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -69,7 +69,7 @@ private
   end
 
   def text_metadata_facets(facets)
-    facets = metadata_facets(facets).select { |f| f.type == "text" }
+    facets = metadata_facets(facets).select { |f| %w[text nested].include?(f.type) }
     facets.reject do |facet|
       facet.key == "organisations" && is_mainstream_content?
     end
@@ -102,16 +102,19 @@ private
   def text_metadata(facets)
     metadata_labels = {}
     text_metadata_facets(facets).each do |facet|
-      document_values = document_hash[facet.key]
-      next if document_values.nil?
-
-      metadata_labels[facet.key] = if facet.allowed_values.empty?
-                                     metadata_labels_from_document_values(document_values, facet.key)
-                                   else
-                                     metadata_labels_from_allowed_values(facet.allowed_values, document_values, facet.key)
-                                   end
+      metadata_labels[facet.key] = extract_facet_key_value(facet.allowed_values, facet.key, document_hash[facet.key]) if document_hash[facet.key].present?
+      metadata_labels[facet.sub_facet_key] = extract_facet_key_value(facet.sub_facets_allowed_values, facet.sub_facet_key, document_hash[facet.sub_facet_key]) if facet.respond_to?(:sub_facet_key) && document_hash[facet.sub_facet_key].present?
     end
+
     metadata_labels.map(&method(:build_text_metadata)).select(&method(:metadata_value_present?))
+  end
+
+  def extract_facet_key_value(allowed_values, facet_key, document_values)
+    if allowed_values.empty?
+      metadata_labels_from_document_values(document_values, facet_key)
+    else
+      metadata_labels_from_allowed_values(allowed_values, document_values, facet_key)
+    end
   end
 
   def build_text_metadata(facet_key, values)
@@ -160,6 +163,9 @@ private
   end
 
   def label_for_metadata_key(facets, key)
+    facet = metadata_facets(facets).find { |f| f.try(:sub_facet_key) == key }
+    return facet.sub_facet_name if facet
+
     facet = metadata_facets(facets).find { |f| f.key == key }
 
     if format == "oim_project"
--- a/app/models/nested_facet.rb
+++ b/app/models/nested_facet.rb
@@ -17,6 +17,7 @@ class NestedFacet < FilterableFacet
       option = {
         text: allowed_value["label"],
         value: allowed_value["value"],
+        selected: @value_hash[key] == allowed_value["value"],
       }
       options << option
     end
@@ -33,6 +34,7 @@ class NestedFacet < FilterableFacet
       option = {
         text: facet_text(sub_facet_value),
         value: sub_facet_value["value"],
+        selected: @value_hash[sub_facet_key] == sub_facet_value["value"],
       }
       option.merge!(data_attributes: { main_facet_value: sub_facet_value["main_facet_value"], main_facet_label: sub_facet_value["main_facet_label"] })
 
--- a/app/models/nested_facet.rb
+++ b/app/models/nested_facet.rb
@@ -42,7 +42,74 @@ class NestedFacet < FilterableFacet
     end
   end
 
+  def applied_filters
+    return [] unless has_filters?
+
+    main_facet_filter = {
+      name: name,
+      label: selected_main_facet_value[:text],
+      query_params: {
+        main_facet_key => selected_main_facet_value[:value],
+        sub_facet_key => selected_sub_facet_value&.fetch(:value),
+      }.compact,
+    }
+
+    if selected_sub_facet_value
+      sub_facet_filter = {
+        name: sub_facet_name,
+        label: selected_sub_facet_value[:text],
+        query_params: { sub_facet_key => selected_sub_facet_value[:value] },
+      }
+    end
+
+    [main_facet_filter, sub_facet_filter].compact
+  end
+
+  def has_filters?
+    selected_main_facet_value.present?
+  end
+
+  def sentence_fragment
+    return nil if selected_main_facet_value.nil?
+
+    {
+      "type" => "nested",
+      "preposition" => preposition,
+      "values" => value_fragments,
+      "word_connectors" => and_word_connectors,
+    }
+  end
+
 private
+
+  def value_fragments
+    [
+      value_fragment(selected_main_facet_value, key),
+      value_fragment(selected_sub_facet_value, sub_facet_key),
+    ].compact
+  end
+
+  def value_fragment(value, key)
+    return nil if value.nil?
+
+    {
+      "label" => value[:text],
+      "parameter_key" => key,
+      "value" => value[:value],
+    }
+  end
+
+  def selected_main_facet_value
+    @selected_main_facet_value ||= main_facet_options.find do |v|
+      v[:value] == @value_hash[key]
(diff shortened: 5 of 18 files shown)


app/models/nested_facet.rb
```diff
@@ -80,6 +80,13 @@ def sentence_fragment
     }
   end
 
+  def query_params
```
lauraghiorghisor-tw, line 83: No idea if anything uses this or why needed. See commit message.

app/lib/search/query_builder.rb
```diff
@@ -107,9 +107,9 @@ def base_return_fields
     end
 
     def metadata_fields
-      raw_facets.map do |f|
-        unfilterise(f["filter_key"] || f["key"])
-      end
+      raw_facets.map { |f|
+        f["sub_facet_key"] ? [unfilterise(f["filter_key"] || f["key"]), unfilterise(f["sub_facet_key"])] : unfilterise(f["filter_key"] || f["key"])
```
ryanb-gds, line 111: Can I suggest we go for normal conditions over ternary statements here for better readability?

ryanb-gds (commented):
I don't really know what I'm looking at here so will leave to others to approve but I was just being nosy and wanted to say that I think this is a really nice PR. Excellent commit messages and seemingly as elegant/minimal code as possible code to solve a fiddly problem.

spec/controllers/finders_controller_spec.rb
```diff
@@ -142,7 +141,7 @@
             )
             .to_return(status: 200, body: rummager_response, headers: {})
 
-          expect(NestedFacet).to receive(:new).with(
+          expect(OptionSelectFacet).to receive(:new).with(
```
lauraghiorghisor-tw, line 144: This was written to have some coverage over the params values being sent to the facets builder (they are eventually used to set the selected value on the drop down options) - whilst originally written with a nested flag thrown in there, I think it has the same value being more generic.

app/lib/filters/nested_filter.rb
```diff
@@ -0,0 +1,28 @@
...
+      @main_facet_value = params[@main_facet_key]
+      @sub_facet_value = params[@sub_facet_key]
+
+      super
+    end
+
+    def query_hash
+      return {} unless @main_facet_value && @sub_facet_value
+
+      { @main_facet_key => [@main_facet_value], @sub_facet_key => [@sub_facet_value] }
```
ChrisBAshton, line 15: This would probably be more readable spread over multiple lines. (I'd be happy with a one-liner if these weren't array values)

app/lib/filters/nested_filter.rb
ChrisBAshton, line 1: > It appears that the `TaxonFilter` sends a load different from the one we need, so we define our own custom Filter.

It would be good to expand on this in the commit message a bit, just to hammer home why we're having to roll out something bespoke.

app/models/nested_facet.rb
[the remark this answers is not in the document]
ChrisBAshton: 👀 interesting...

From the commit message, sounds like the `query_params` methods for other facets can likely be deleted (in another PR, outside of scope). I'd probably drop this commit in the meantime: we shouldn't add code that doesn't seem to do anything, even if it's just for consistency.

ChrisBAshton (commented):
Approach LGTM 👍 nice one for figuring all this out.

app/models/nested_facet.rb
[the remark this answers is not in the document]
lauraghiorghisor-tw: Waiting to see if anyone has any idea - will drop it otherwise ;)

lauraghiorghisor-tw:
> EDIT: the screenshot here:

Ah, yes, I did notice that. But checked it against a taxon finder and it was weird there too, so I embraced it. I added screenshots and link to the taxon finder for reference. Thanks for catching that ;)

app/views/finders/_nested_facet.html.erb
```diff
@@ -1,15 +1,19 @@
-<div data-module="nested-facets" data-main-facet-id="<%= nested_facet.main_facet_key %>" data-sub-facet-id="<%= nested_facet.sub_facet_key %>">
-  <%= render "govuk_publishing_components/components/select", {
-    id: nested_facet.key,
-    label: nested_facet.name,
-    full_width: true,
-    options: nested_facet.main_facet_options
-  } %>
+<%= render "components/expander", {
```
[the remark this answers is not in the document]
andysellick: You might need to check with a performance analyst on what tracking might be required here.

lauraghiorghisor-tw:
@KludgeKML Would you be able to have a look?

app/lib/facets_builder.rb
```diff
@@ -45,6 +43,8 @@ def build_facet(facet_hash)
         RadioFacetForMultipleFilters.new(facet_hash, value_hash[facet_hash["key"]], ::Filters::ResearchAndStatsHashes.new.call)
       when "official_documents"
         RadioFacetForMultipleFilters.new(facet_hash, value_hash[facet_hash["key"]], ::Filters::OfficialDocumentsHashes.new.call)
+      when "nested"
```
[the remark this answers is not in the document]
deborahchua: Would you be able to update the commit links which have changed since your latest push.

app/lib/filters/nested_filter.rb
```diff
@@ -0,0 +1,31 @@
...
+      @main_facet_key = params["main_facet_key"]
+      @sub_facet_key = params["sub_facet_key"]
+      @main_facet_value = params[@main_facet_key]
+      @sub_facet_value = params[@sub_facet_key]
+
+      super
+    end
+
+    def query_hash
+      return {} unless @main_facet_value && @sub_facet_value
```
KludgeKML, line 13: Test coverage doesn't cover this return triggering - is it possible in practise?

app/lib/filters/nested_filter.rb
```diff
@@ -0,0 +1,31 @@
...
+        @main_facet_key => [@main_facet_value],
+        @sub_facet_key => [@sub_facet_value],
+      }
+    end
+
+  private
+
+    def value
+      @value ||= begin
+        return nil unless params
```
KludgeKML, line 25: Test coverage doesn't cover this return triggering - is it possible in practise?

app/models/nested_facet.rb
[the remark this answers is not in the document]
KludgeKML: I agree - drop it. It's not covered by the test suite, and if you see it working in practise without this code...

spec/controllers/finders_controller_spec.rb
KludgeKML: Is this essentially reverting a previous commit, then? If so, could you drop a link to that original commit?

spec/models/nested_facet_spec.rb
```diff
@@ -1,183 +1,121 @@
...
-      end
+        ],
+      },
+      {
+        "label" => "Allowed value 3",
+        "value" => "allowed-value-3",
+      },
+    ]
+  end
+  let(:value_hash) { {} }
```
KludgeKML, line 57: Since value_hash is never varied, should it just be hardcoded in the subject on line 4?

spec/models/nested_facet_spec.rb
```diff
@@ -117,5 +123,62 @@
...
+
+      it "returns `selected` flag for each option" do
+        expect(subject.main_facet_options).to eq(
+          [
+            {
+              "text": "All facet names",
+              "value": "",
+            },
+            {
+              "text": "Allowed value 1",
```
KludgeKML, line 156: Could you add in an allowed value that _isn't_ selected as well, because otherwise this expectation would be satisfied by line 20 of nested_facet just being:
`selected: true,`

spec/models/nested_facet_spec.rb
```diff
@@ -117,5 +123,62 @@
...
+              "text": "All sub facet names",
+              "value": "",
+            },
+            {
+              text: "Allowed value 1 - Sub facet Value 1",
+              value: "allowed-value-1-sub-facet-value-1",
+              "selected": true,
+              "data_attributes":
+                {
+                  "main_facet_label": "Allowed value 1",
```
KludgeKML, line 175: Could you add in an allowed value that isn't selected as well, because otherwise this expectation would be satisfied by line 37 of nested_facet just being:
selected: true,

KludgeKML:
@lauraghiorghisor-tw - @deborahchua and I have looked through this PR and it looks broadly okay, but there are a few notes around commit messages, coverage, and an odd method. I think we'd be happy to sign off if those areas were addressed.

spec/controllers/finders_controller_spec.rb
lauraghiorghisor-tw: Just changing the commit a bit, not reverting - test was in reply to Chris's comment on the commit here https://github.com/alphagov/finder-frontend/pull/3679/commits/0d183dc3669034cc1ff40b885e2311dff7355b63#r1973177322.

spec/models/nested_facet_spec.rb
lauraghiorghisor-tw: Added `"value" => "allowed-value-1-sub-facet-value-2"` which is now not included in the `value_hash`.

spec/models/nested_facet_spec.rb
lauraghiorghisor-tw: Added `"value": "allowed-value-2"` that is not in the `value_hash`, for which selected is now `false`.

spec/models/nested_facet_spec.rb
lauraghiorghisor-tw: I am now adding it as a `let` var in the subsequent commit where it is varied.

app/models/nested_facet.rb
```diff
@@ -1,35 +1,132 @@
...
 
+  def value_fragments
+    [
+      value_fragment(selected_main_facet_value, key),
+      value_fragment(selected_sub_facet_value, sub_facet_key),
+    ].compact
+  end
+
+  def value_fragment(value, key)
+    return nil if value.nil?
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Yes, it is, it's the flow where there is no facet selection/only main facet selection. 

If there is no query parameter corresponding to a known facet or sub facet key, then the `value_fragment` is nil for that key, then the `value_fragments` is either `[]`, or has a single item. We exit the `sentence_fragment` if there is no main facet value, because we are not really interested in representing an "orphaned" sub facet selection. But we still call all the other methods if there is only main but no sub-facet value so we need the nil return in the `value_fragment`.

From what I can tell, the `sentence_fragment` is used in [this presenter](https://github.com/alphagov/finder-frontend/blob/2fe653c7bd78987a0c112e318b6524f085ed5fc1/app/presenters/facet_tags_presenter.rb#L18) where there is a presence check. 

Will add more tests in the `describe "#sentence_fragment"`.

app/lib/filters/nested_filter.rb
lauraghiorghisor-tw: Thanks for pointing this out! 

Added additional cases in the `query_builder_spec.rb` + changed the logic so that we can still filter by any single key rather than return {} if either are missing. This is to match the behaviour where both this query https://www.gov.uk/search/news-and-communications?level_two_taxon=13fb7519-c48e-4577-90f8-aaf483fa0c03&order=updated-newest (subfacet only) and this query https://www.gov.uk/search/news-and-communications?level_one_taxon=495afdb6-47be-4df1-8b38-91c8adb1eefc&level_two_taxon=13fb7519-c48e-4577-90f8-aaf483fa0c03&order=updated-newest (both facets) give the same query. We're not 100% what the user preference is. 

The previous `return {} unless @main_facet_value && @sub_facet_value` would have returned all results (no filter applied).

Just to note here - as per the taxon behaviour (and now nested), even if the search filter is applied and the results populated, the "applied filters" do not show at the top of the search results, and the selects do not get `selected` if there is only a sub-facet key query param passed. But at least this behaviour is now consistent across the two facets.

app/lib/filters/nested_filter.rb
lauraghiorghisor-tw: Thanks for pointing this out!

We're required to define this method if we inherit from `Filter` -> we only ever use it in `Filter`'s `active?` method which is then used in the [`FilterQueryBilder`](https://github.com/alphagov/finder-frontend/blob/48e06c7ec51ebad2936b37e1a3457d8c7e19e9af/app/lib/search/filter_query_builder.rb#L11).

I refined the logic a bit to be specific to our params and added tests for it in a new `nested_filter_spec.rb` - testing the `active?` method as `value` is private, but that suffices since it is the only usage. Also covered the `query_hash` in that test.

app/lib/facets_builder.rb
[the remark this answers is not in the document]
lauraghiorghisor-tw: I have given up on updating that on every rebase - so mentioning the commit by name 😅

app/lib/filters/nested_filter.rb
lauraghiorghisor-tw: ✅ Added more test coverage in `nested_filter_spec.rb`.

app/models/document.rb
```diff
@@ -102,18 +102,22 @@ def build_date_metadata(key)
...
-                                   else
-                                     metadata_labels_from_allowed_values(facet.allowed_values, document_values, facet.key)
-                                   end
+      extract_metadata_labels(facet.allowed_values, facet.key, metadata_labels, document_hash[facet.key])
+      extract_metadata_labels(facet.sub_facets_allowed_values, facet.sub_facet_key, metadata_labels, document_hash[facet.sub_facet_key]) if facet.respond_to?(:sub_facet_key)
     end
     metadata_labels.map(&method(:build_text_metadata)).select(&method(:metadata_value_present?))
   end
 
+  def extract_metadata_labels(allowed_values, facet_key, metadata_labels, document_values)
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Did a thing, check it out.

app/models/nested_facet.rb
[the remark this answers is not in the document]
lauraghiorghisor-tw: I double checked the commit history - and it seems the `query_params` pre-dates some changes in the `applied_filters` but it was likely left behind - see https://github.com/alphagov/finder-frontend/blob/48e06c7ec51ebad2936b37e1a3457d8c7e19e9af/app/models/taxon_facet.rb#L56.

deborahchua (approved):
Thanks for making those changes @lauraghiorghisor-tw , looks good to me! 👍

KludgeKML (approved):
LGTM!

PR #3705, opened 2025-03-20
Don't use Google Vertex for feeds with keywords

sihugh:
Feeds are overridden so that they (rightly) order items newest to oldest.

This negates the relevance benefits of Google Vertex and means we're paying over the odds. Feed readers are also voracious, so are likely costing us a chunk of money for little benefit for users.

(First commit is the actual small change, second one is a rename as the meaning has changed)

- https://finder-frontend-pr-3705.herokuapp.com/search/all.atom?q=steve
- https://finder-frontend-pr-3705.herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://finder-frontend-pr-3705.herokuapp.com/drug-device-alerts

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -145,7 +145,7 @@ private
     Search::Query.new(
       content_item,
       filter_params,
-      override_sort_for_feed: is_for_feed,
+      is_for_feed: is_for_feed,
       ab_params: {},
       v2_serving_config:,
     )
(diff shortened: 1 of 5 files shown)


csutter (approved):
Non-blocking feedback: My gut feeling is that `is_feed` or `is_for_feed` would be a better name than `override_settings_for_feed`, but I'm not wedded to the idea 🤔

PR #3772, opened 2025-06-13
Let Search API v2 decide default serving config

csutter:
We currently define a "default" serving config (when neither an AB test nor a debug flag needs to override it) both in this app and in Search API v2 itself.

This makes it harder to update what the default is as we'd have to do it in two places at the same time.

Instead, this changes Finder Frontend to just not specify a serving config in situations where we don't want to override the default, and leaves it to Search API v2 to figure out.

### DIFF
--- a/spec/lib/search/query_builder_spec.rb
+++ b/spec/lib/search/query_builder_spec.rb
@@ -446,8 +446,8 @@ describe Search::QueryBuilder do
       context "if no custom serving config is given" do
         let(:v2_serving_config) { nil }
 
-        it "includes the default serving config query" do
-          expect(query).to include("serving_config" => "default_search")
+        it "does not include a serving config query" do
+          expect(query).not_to have_key("serving_config")
         end
       end
     end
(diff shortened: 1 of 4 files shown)


spec/lib/search/query_builder_spec.rb
```diff
@@ -446,8 +446,8 @@
       context "if no custom serving config is given" do
         let(:v2_serving_config) { nil }
 
-        it "includes the default serving config query" do
-          expect(query).to include("serving_config" => "default_search")
+        it "does not include a serving config query" do
+          expect(query).not_to have_key("serving_config")
```
leenagupte, line 450: Shouldn't this be `include("serving_config" => nil)` rather than the key not existing at all?

leenagupte (commented):
I have one small inline question

spec/lib/search/query_builder_spec.rb
csutter: No, as we `#compact` the params in `QueryBuilder#v2_serving_config_query` so it won't get sent through at all 🚫

leenagupte: Ah! I missed that. Thanks!

leenagupte (approved):

PR #3785, opened 2025-06-27
Use new parts link field from Search API

chaecramb:
Search API now includes a `link` field in every element of the parts array. For some attachments whose part paths do not equal `parent_path + "/" + slug`, consuming this new field avoids broken links (e.g. consultation outcomes that live under `/.../outcome/...)`.

This PR:

- Updates `SearchResultPresenter#structure_parts`
    - If `part[:link]` is present, use it directly.
    - Otherwise keep the previous behaviour of concatenating the `parent_path` and `slug` to form the path.


kevindew (commented):
Looks good to me. 

Shall we amend the comment a little to suggest the fallback is just for the transitionary period while link is populated, as we expect the fallback is only temporary?

PR #3795, opened 2025-07-15
Skip parts that arrive without a `slug`

chaecramb:
Trello: https://trello.com/c/L3TUHrsG/2564-fix-broken-search-api-links

The upcoming Search API re-index (https://github.com/alphagov/search-api/pull/3271) will introduce a new `link` field on each part that stores the full attachment path, replacing the current approach of supplying only a `slug`.
This change is required because some consultation-outcome attachments live under an extra `/outcome/` directory, so concatenating the parent URL with their `slug` produces a 404.  Supplying the whole path in `link` fixes that.

During the back-fill there will be a transition period where Search API re-presents documents with `link` populated but potentially a blank `slug`.  If finder-frontend tries to render such a part it would produce a broken URL and raise a presenter error.

This commit adds a guard that simply skips parts with a blank `slug`. It clears the path for a follow-up commit that will start using `link` and ultimately retire `slug`

This safeguard can be removed once all documents carry `link` and the legacy `slug` field has been fully phased out.

### DIFF
--- a/app/presenters/search_result_presenter.rb
+++ b/app/presenters/search_result_presenter.rb
@@ -85,6 +85,10 @@ private
 
   def structure_parts
     structured_parts = parts.map do |part|
+      # Silently skip parts that have no slug
+      # this is an interim conditional until `slug` is replaced by `link` in Search API
+      next if part[:slug].blank?
+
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM!

PR #3856, opened 2025-09-30
Use inverse header for taxon and content type search results pages

hannalaakso:
## What

Use inverse header for taxon and content type search results pages (topic pages with filter parameters).

**Review URLs (unchanged)**:
- https://finder-frontend-pr-3856.herokuapp.com/search/all
- https://finder-frontend-pr-3856.herokuapp.com/drug-device-alerts
- https://finder-frontend-pr-3856.herokuapp.com/drug-safety-update
- https://finder-frontend-pr-3856.herokuapp.com/administrative-appeals-tribunal-decisions

**Integration only**
- https://www.integration.publishing.service.gov.uk/search/guidance-and-regulation?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://www.integration.publishing.service.gov.uk/search/guidance-and-regulation?topic=6acc9db4-780e-4a46-92b4-1812e3c2c48a

## Why

To bring all our templates (including finder pages that accept a topic ID) that use the inverse header up to date with the latest gem changes, as well as reducing duplication of code in custom app styles where a shared gem component could be used.

https://gov-uk.atlassian.net/jira/software/c/projects/NAV/boards/1422?selectedIssue=NAV-18210

### Before

Screenshot 2025-10-01 at 11 49 02

### After

Screenshot 2025-10-01 at 11 48 49

## Anything else

There may be an issue with the use of the phase banner here, although it shouldn’t affect search, taxon, or topic pages.

From what I can see, the phase banner is only used on specialist finder pages, such as:
- https://www.gov.uk/tax-and-chancery-tribunal-decisions
- https://www.gov.uk/find-funding-for-land-or-farms

The phase banner includes HTML for a hyperlink (for example, “find out what beta means”). This seems to rely on the publisher correctly applying the `govuk-link` class, which doesn’t always happen. That said, it appears technically possible to do this (based on the code https://github.com/alphagov/finder-frontend/blob/main/app/views/finders/_before_content.html.erb#L7), but it could raise an accessibility issue. I’m not entirely sure what the best solution is - should the content include an accessible inverse link that’s passed to the phase banner, or should the styles be handled in the stylesheet (which would require overrides)?

## Depends on

https://github.com/alphagov/govuk_publishing_components/pull/5053 - I've added a Do Not Merge label until the gem PR has been merged.

Fixes https://gov-uk.atlassian.net/jira/software/c/projects/NAV/boards/1422?selectedIssue=NAV-18210.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -192,7 +192,7 @@ GEM
     govuk_personalisation (1.1.0)
       plek (>= 1.9.0)
       rails (>= 6, < 9)
-    govuk_publishing_components (61.4.1)
+    govuk_publishing_components (62.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown


MartinJJones (commented):
Nice work, the changes look good to me, happy to approve once the gem bump to get the latest inverse header component styles is included

MartinJJones (approved):

PR #3970, opened 2026-01-28
Handle nil fields in organisations registry

CodeSonia:
### What

Previously, the `organisations_as_hash` method assumed that every result from the `Rummager search` had both a slug and a title. This caused a `NoMethodError` when either was nil:

`undefined method 'empty?' for nil (NoMethodError)
        .reject { |result| result["slug"].empty? || result["title"].empty?`

This change replaces `empty?` with `blank?`, which safely handles `nil` values. As a result, fetching or caching organisations will no longer crash, even when some are missing slugs.

- https://[HEROKU-APP-ID].herokuapp.com/search/all
- https://[HEROKU-APP-ID].herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://[HEROKU-APP-ID].herokuapp.com/drug-device-alerts

### DIFF
--- a/app/lib/registries/organisations_registry.rb
+++ b/app/lib/registries/organisations_registry.rb
@@ -29,7 +29,7 @@ module Registries
     def organisations_as_hash
       GovukStatsd.time("registries.organisations.request_time") do
         fetch_organisations_from_rummager
-          .reject { |result| result["slug"].empty? || result["title"].empty? }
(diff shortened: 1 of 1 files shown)


sairamya93 (approved):
LGTM!

PR #4078, opened 2026-06-01
Remove phase banner from before content and call in layout for public

luke-booker:
## What
See publishing components PR for further context: https://github.com/alphagov/govuk_publishing_components/pull/5481#issue-4564134511

Removed the rendering of the phase banner from the body view architecture and directly feed it into the layout configuration.

## Why
Accompanies the foundational layout changes introduced in the publishing components gem to transition the phase banner into the header element for accessibility compliance.

Jira card: https://gov-uk.atlassian.net/browse/[NAV-18918]

## Visual Changes
Before: The phase banner rendered visually separated below the main header.

After: The phase banner successfully renders inside the header element.

| Before | After |
|--------|--------|
| Screenshot 2026-06-02 at 10 35 25 | Screenshot 2026-06-02 at 10 35 42 | 
Screenshot 2026-06-02 at 10 37 54 | Screenshot 2026-06-02 at 10 38 48 |

[NAV-18918]: https://gov-uk.atlassian.net/browse/NAV-18918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/layouts/finder_layout.html.erb
+++ b/app/views/layouts/finder_layout.html.erb
@@ -1,4 +1,13 @@
-<% title = yield :title %>
+<%
+  title = yield :title
+
+  if @content_item&.show_phase_banner?
+    phase_banner = {
+      phase: @content_item.phase,
+      message: sanitize(@content_item.phase_message)
+    }
+  end
+%>
 
 <%= content_for :head do %>
   <%= javascript_include_tag 'application', integrity: false, type: "module" %>
@@ -15,6 +24,7 @@
     title: safe_join([title, "GOV.UK"], " - "),
     emergency_banner: render("govuk_web_banners/emergency_banner"),
     global_banner: render("govuk_web_banners/global_banner"),
+    phase_banner:
   } do %>
     <div class="<%= class_names("govuk-width-container" => !@full_width) %>">
       <%= yield :before_content %>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -463,7 +463,7 @@ GEM
       opentelemetry-common (~> 0.20)
       opentelemetry-registry (~> 0.2)
       opentelemetry-semantic_conventions
-    opentelemetry-semantic_conventions (1.40.0)
+    opentelemetry-semantic_conventions (1.41.0)
       opentelemetry-api (~> 1.0)
     ostruct (0.6.3)
     parallel (1.28.0)
--- a/app/views/finders/_before_content.html.erb
+++ b/app/views/finders/_before_content.html.erb
@@ -1,13 +1,6 @@
 <% content_for :before_content do %>
   <div class="app-before-content<%= " app-before-content--inverse" if is_inverse? %>">
     <div class="<%= 'govuk-width-container' if is_inverse? %>">
-      <% if content_item.show_phase_banner? %>
-        <%= render 'govuk_publishing_components/components/phase_banner', {
-            phase: content_item.phase,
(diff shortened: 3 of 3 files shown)


app/views/layouts/finder_layout.html.erb
```diff
@@ -15,6 +15,7 @@
     title: safe_join([title, "GOV.UK"], " - "),
     emergency_banner: render("govuk_web_banners/emergency_banner"),
     global_banner: render("govuk_web_banners/global_banner"),
+    phase_banner: (defined?(content_item) && content_item&.show_phase_banner? ? { phase: content_item.phase, message: sanitize(content_item.phase_message) } : nil)
```
jon-kirwan: One small comment - there's quite a bit happening here in one line, so I wonder if we might extract it into a variable and then just pass that into the phase banner instead, e.g.

```rb
<%
  phase_banner = if content_item.show_phase_banner?
    {
      phase: content_item.phase,
      message: sanitize(content_item.phase_message)
    }
  end
%>

...

<%= render "govuk_publishing_components/components/layout_for_public", {
  emergency_banner: render("govuk_web_banners/emergency_banner"),
  global_banner: render("govuk_web_banners/global_banner"),
  phase_banner:,
```

Also, I think it’s safe to assume there will be a content_item, so you can probably drop the `defined?` check. Just keep the check on `content_item&.show_phase_banner?`.

jon-kirwan: Another approach would be to render the phase banner here, similar to the emergency and global banners on the preceding lines, and pass the rendered banner into the component.

That would make the handling consistent across these 3 banners. That said, it still wouldn’t be exactly the same, since the phase banner includes logic to check whether it should be shown, as well as options that are passed into the component.

@hannalaakso @MartinJJones - do you have any thoughts?

jon-kirwan (commented):
Thanks @luke-booker, all looking good and working well. I’ve added a few comments and copied in Hanna and Martin to get another opinion on the best approach for passing in the phase banner - either as a fully rendered component or as options for rendering it within the gem.

app/views/layouts/finder_layout.html.erb
luke-booker: Thank you for reviewing @jon-kirwan, I've made some updates.

Hopefully I'm understanding this correctly, but the phase banner expects a config hash whereas the emergency and global banners accept injected HTML. Am I right in thinking we want the emergency and global banners to be flexible content, whereas the phase banner should be rigid/standardised?

Also, unlike the other two banners, given where the phase banner now sits inside the header rather than in the body, passing it as a hash protects the DOM integrity of our navigation. There might be some risk in injecting raw HTML directly into the header? This way the gem retains strict control over the header's HTML markup, while the emergency and global banners can safely accept raw HTML injections because they live further downstream in the body. From what I can see there isn't any raw HTML injection in the header at the moment, curious to see what you think. 

Any thoughts @MartinJJones @hannalaakso?

jon-kirwan (approved):

app/views/layouts/finder_layout.html.erb
jon-kirwan: Thanks, Luke. Just noting that we discussed this and agreed it’s better to keep the phase banner setup in Finder Frontend, and pass the options to the public layout component so it can render the phase banner there.

PR #4082, opened 2026-06-03
Update Dart Sass to v1.100.0 to prepare for Design System v6 upgrade

luke-booker:
## What

- Ran bundle update sass-embedded, which successfully resolved and updated the gem to `sass-embedded (1.100.0)` in `Gemfile.lock`.

## Why

GOV.UK Frontend v6.0 will be dropping support for Ruby Sass, LibSass, and older versions of Dart Sass. To prepare the application for this upgrade, we need to ensure our Sass compiler is running Dart Sass v1.79.0 or later.

Jira card: https://gov-uk.atlassian.net/browse/NAV-18914

## Visual changes

None. This is a dependency requirement upgrade only.

## Testing carried out locally

✅  Verified stylesheets compile successfully. 
✅  All tests passing successfully.
✅  Application is building without issues locally.


MartinJJones (approved):
Nice work, the changes look good to me 👍

non-blocking: it could be worth adding a commit message linking to the changelog entry on the design system https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#use-dart-sass-v1790-or-later-to-compile-your-sass-stylesheets, but not essential

jon-kirwan:
Thanks Luke, this looks good to me.

One other thought, having looked at how this is setup in Frontend which includes `sass-embedded (1.85.1)`: rather than adding `sass-embedded` as a separate entry in the Gemfile, whether we should instead run `bundle update sass-embedded` to get the latest version?

https://github.com/alphagov/frontend/blob/main/Gemfile
https://github.com/alphagov/frontend/blob/main/Gemfile.lock#L700

luke-booker:
@MartinJJones @jon-kirwan 

Thank you both for taking a look at these. I will add the commit message linking to the changelog entry.
I've checked and there are no issues with running bundle update sass-embedded and lose that extra Gemfile entry. I'll make that change and check each app.

PR #4102, opened 2026-07-02
Change breadcrumbs on collections finders

andysellick:

## What
- change collections finders so that the name and link to the collection is moved into the breadcrumbs, instead of directly above the heading as a link
- add the name of the collection as the context in the heading component, to restore the emphasis on where the user is

## Why
The existing pattern isn't a usual thing across GOV.UK - it seems more consistent with other pages that this link to the 'parent' page is part of the breadcrumbs.

## Visual changes

Before

Screenshot 2026-07-02 at 14 59 51

After

Screenshot 2026-07-02 at 14 59 58

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -289,12 +289,6 @@ mark {
   box-sizing: border-box;
 }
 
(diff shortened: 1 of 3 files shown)


KludgeKML (approved):
LGTM

PR #4110, opened 2026-07-07
Restyle finders headers layout

### LINKED ISSUE #4103: Finders headers have inconsistent layouts
The layout of information in the header of finders is often inconsistent. Sometimes it's also a little bit strange, where measures put in place to make the layout work in some situations apply to other situations where it's less appropriate.

I think the reason for this is that in the code there's a single `govuk-grid-row` that wraps everything that could be in the header, followed by a lot of optional grid columns. The main one that is always there is the 2/3rds column containing the title/h1 of the finder, but that can be followed by a 1/3rd or a 2/3rd column, which itself can be followed by more 1/3rd columns. The layout relies upon these columns wrapping onto new lines where they exceed the width of the row, but this is not ideal. We should probably have a more formal structure where each possible row is contained within its own `govuk-grid-row`, rather than having multiple columns that can overflow into new, informal rows.

Below are some examples of the possible header layouts.

# Title only
https://www.gov.uk/government/groups

The row contains a single 2/3rds column.

Image

# Title with metadata
https://www.gov.uk/aaib-reports

The row contains a single 2/3rds column, with the title and metadata about associated organisations.

Image

# Title with metadata, plus related links
https://www.gov.uk/drug-device-alerts

Title and metadata are in a 2/3rds column, related links in a 1/3rd column. It's worth noting that the related links have a margin top and bottom of 50px, probably to prevent them from being at the top of the page along with the title, but this has consequences for other combinations later.

Image

# Title with metadata, plus related links and description
https://www.gov.uk/animal-disease-cases-england

Title and metadata are in a 2/3rds column, followed by another 2/3rds column with the description, followed by a 1/3rd column with related links. We're now at the point where there are two many elements to fit in a single row, so they start to wrap. Note the position of related links now - very pushed down due to the 50px top margin.

Image

# Other considerations

- the [_show_header.html.erb template](https://github.com/alphagov/finder-frontend/blob/main/app/views/finders/_show_header.html.erb#L61-L65) (where all of this happens) also contains another optional 1/3rd column for if there's a `content_item.logo_path`, but I've not found a live example of that
- there are some [inverted, full width headers](https://www.gov.uk/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0) where these extra columns could potentially appear, but again, I haven't found any examples (they should layout the same as the others though if any changes are made, hopefully)
- we might also have to cater for a situation where there is only a heading 1 and related links/logo (with no metadata or description) although I haven't found one yet

andysellick:

## What
- improve the layout of finders headers by splitting the multiple grid columns into distinct rows, adding some specific spacing, and adding additional logic
- ensures the title of the page is always on its own row, with the metadata / description / extra links / image all appearing in a separate row beneath

## Why
Fixes https://github.com/alphagov/finder-frontend/issues/4103

## Visual changes

https://www.gov.uk/drug-device-alerts
Note the change of alignment for the links on the right, now aligned with the metadata beneath the heading.

Before | After
------- | ------
Screenshot 2026-07-07 at 15 28 38 | Screenshot 2026-07-07 at 15 28 48

https://www.gov.uk/animal-disease-cases-england
Note the change of alignment for the links on the right, now aligned with the metadata beneath the heading.

Before | After
------- | ------
Screenshot 2026-07-07 at 15 28 22 | Screenshot 2026-07-07 at 15 28 29

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -1,8 +1,15 @@
 $app-colour-true-black: #000000;
 
-.metadata-summary {
+.header-item-row {
   margin-bottom: govuk-spacing(6);
-  @include govuk-font(19);
+
+  @include govuk-media-query($from: tablet) {
+    margin-bottom: govuk-spacing(8);
+  }
+}
+
+.header-item-element {
+  margin-bottom: govuk-spacing(6);
 }
 
 .finder-logo {
@@ -11,11 +18,11 @@ $app-colour-true-black: #000000;
 
 .finder-logo__image {
   max-width: 100%;
+  margin-bottom: govuk-spacing(6);
 }
 
 .related-links {
-  margin-top: govuk-spacing(8);
-  margin-bottom: govuk-spacing(8);
+  margin-bottom: govuk-spacing(6);
 }
 
 .related-links__item {
--- a/app/helpers/page_metadata_helper.rb
+++ b/app/helpers/page_metadata_helper.rb
@@ -1,12 +1,11 @@
 module PageMetadataHelper
-  def page_metadata(content_item, filter_params)
+  def page_metadata(content_item)
     organisation_links = content_item.organisations.map do |organisation|
       link_to(organisation["title"], organisation["web_url"])
     end
 
     {}.tap do |metadata|
       metadata[:from] = organisation_links if organisation_links.present?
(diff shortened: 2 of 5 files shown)


KludgeKML (approved):
LGTM. Visibly much improved!

AshGDS (approved):
LGTM 👍

PR #4119, opened 2026-07-15
[PNP-10075] Add maximum page value for search

KludgeKML:
- Google's search can't handle situations where we've asked it to cap results at more than 2,147,483,647 (max signed int 32). The content item provides details/default_documents_per_page of 10, and we use that to derive a max page size that is safe to ask Google for.
- If the page request is bigger than that, return a BadRequest.
- Also return a BadRequest if the page is negative (Google search will handle this by just rounding up to 1, but we shouldn't be encouraging it).

https://gov-uk.atlassian.net/browse/PNP-10075

- https://[HEROKU-APP-ID].herokuapp.com/search/all
- https://[HEROKU-APP-ID].herokuapp.com/search/news-and-communications?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
- https://[HEROKU-APP-ID].herokuapp.com/drug-device-alerts

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -77,6 +77,12 @@ private
     unless filter_params.fetch(:keywords, "").is_a?(String)
       raise ActionController::BadRequest, "Invalid 'keywords' query parameter"
     end
+
+    max_page = Search::Query::MAX_RESULTS_VALUE / content_item.default_documents_per_page
+    page = filter_params.fetch(:page, 0).to_i
+    if page.negative? || page > max_page
+      raise ActionController::BadRequest, "Invalid page value"
+    end
   end
 
   def redirect_to_destination
--- a/app/lib/search/query.rb
+++ b/app/lib/search/query.rb
@@ -3,6 +3,7 @@
 module Search
   class Query
     SITE_SEARCH_FINDER_BASE_PATH = "/search/all".freeze
+    MAX_RESULTS_VALUE = 2_147_483_647
 
     include ActiveModel::Validations
 
--- a/spec/controllers/finders_controller_spec.rb
+++ b/spec/controllers/finders_controller_spec.rb
@@ -245,6 +245,20 @@ describe FindersController, type: :controller do
         get :show, params: { slug: "lunch-finder", q: { "invalid" => "hash" } }
         expect(response.status).to eq(400)
       end
+
+      it "renders a BadRequest when the page parameter is negative" do
+        stub_content_store_has_item("/lunch-finder", lunch_finder)
+
+        get :show, params: { slug: "lunch-finder", q: "hash", page: "-1" }
+        expect(response.status).to eq(400)
+      end
+
+      it "renders a BadRequest when the page parameter is larger than the maximum page size" do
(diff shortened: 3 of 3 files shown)


emmalowe:
I'm curious about our general approach to error handling like this, because I think we should put something like this inside `search-api-v2` for requests from other places (e.g. app). Is it bad to have it in both places? Here it would also apply to requests to v1 (I think), which I'm not sure is really necessary (though with such a large number, probably doesn't do any harm).

KludgeKML:
> I'm curious about our general approach to error handling like this, because I think we should put something like this inside `search-api-v2` for requests from other places (e.g. app). Is it bad to have it in both places? Here it would also apply to requests to v1 (I think), which I'm not sure is really necessary (though with such a large number, probably doesn't do any harm).

@emmalowe Yeah, I did consider that it would affect V1 and as you say, I figured in this case no harm because it's unlikely any page that high will actually exist.

I think "defence in depth" is wise here - you should probably also have it in search-api-v2 somewhere, if only because you can't rely on Finder-Frontend being the only place that calls you. There are (or may be) some other consumers of search-api-v2, right?

emmalowe:
> I think "defence in depth" is wise here - you should probably also have it in search-api-v2 somewhere, if only because you can't rely on Finder-Frontend being the only place that calls you. There are (or may be) some other consumers of search-api-v2, right?

Right, that's exactly what I was thinking. We get calls from the GOV.UK app as well, for example.

I was more wondering about the other way round, where if there is validation in `search-api-v2` do you need it in `finder-frontend` as well? I guess errors coming back from `search-api-v2` need to be handled in `finder-frontend` anyway, and it's better not to send the request in the first place if you know it's not going to work?

KludgeKML:
> > I think "defence in depth" is wise here - you should probably also have it in search-api-v2 somewhere, if only because you can't rely on Finder-Frontend being the only place that calls you. There are (or may be) some other consumers of search-api-v2, right?
> 
> Right, that's exactly what I was thinking. We get calls from the GOV.UK app as well, for example.
> 
> I was more wondering about the other way round, where if there is validation in `search-api-v2` do you need it in `finder-frontend` as well? I guess errors coming back from `search-api-v2` need to be handled in `finder-frontend` anyway, and it's better not to send the request in the first place if you know it's not going to work?

I mean, that would certainly be okay. There's a trade-off here, I guess. If finder-frontend _and_ search-api-v2 handle this error, then we have the downside of multiple bits of code but the upside of reduced internal traffic (because the problem is stopped closer to the edge). If search-api-v2 handles it and passes back bad request we have the downside of increased traffic but the upside of having a single maintenance point.

I think given our resources, single maintenance point is the better choice here. Do you want to add the code to search-api-v2 to return a 400, and we'll handle GdsApi::HTTPBadRequest?

emmalowe:
> Do you want to add the code to search-api-v2 to return a 400, and we'll handle GdsApi::HTTPBadRequest?

Sure, I can do that. That approach makes sense to me.

https://github.com/alphagov/search-api-v2/pull/703

KludgeKML:
Closing as made obsolete by: https://github.com/alphagov/finder-frontend/pull/4122

PR #4136, opened 2026-08-06
Fix highlighted search results in high contrast

### LINKED ISSUE #2649: `<mark>` elements in search results can be invisible in Windows High Contrast Mode
### Description of issue

If you use https://www.gov.uk/search with Windows High Contrast Mode (HCM) enabled, your search results will have invisible words.

Screenshot of issue

### Affected systems

This happens in Edge, Chrome, and IE11 on Windows, when using Windows HCM with a dark theme.

This doesn't happen in Firefox.

### Cause of issue

The finder frontend styles uses `<mark>` elements to wrap the search query in search results, but styles them to have a transparent background colour:

https://github.com/alphagov/finder-frontend/blob/009fa6353e20426e099f951ab0877bdece95f344/app/assets/stylesheets/finder_frontend.scss#L30

In forced colour modes such as HCM, the browser will will ignore the CSS colour properties, and use the default colours, except for the alpha value of the background colour [^1]. This means that the mark will be forced to have a yellow background and black text, except the background will be transparent. With a black background this becomes black text on a black background.

[^1]: https://bugs.chromium.org/p/chromium/issues/detail?id=1253295

### Suggested fix

The tech-docs-gem had a similar issue and fixed it in https://github.com/alphagov/tech-docs-gem/pull/265 by using the `CanvasText` colour, which is respected in forced colour modes.

andysellick:

## What / why
Fixes the highlighted text in search results when using high contrast mode in Windows 11.

- mark elements were invisible (black text on a black background) in some browsers when Windows high contrast mode was enabled
- occurred on the words in user search queries, highlighted in search results
- solution copied from the tech docs gem: https://github.com/alphagov/tech-docs-gem/pull/265
- but tested on Windows in various browsers
- Fixes https://github.com/alphagov/finder-frontend/issues/2649

Note for testing: Windows high contrast mode is available in browserstack through the 'Settings' option in the left hand menu.

## Visual changes
Before | After
------ | -----
Screenshot 2026-08-06 at 10 27 16 | Screenshot 2026-08-06 at 10 27 25

### DIFF
--- a/app/assets/stylesheets/finder_frontend.scss
+++ b/app/assets/stylesheets/finder_frontend.scss
@@ -38,6 +38,22 @@ mark {
   background-color: inherit;
   color: inherit;
   @include govuk-typography-weight-bold;
+
+  // In forced color mode some browsers will keep the background transparent
+  // but set the text colour black, making the highlighted text unreadable.
+  // The following blocks fix this by setting the text colour to be the same as
+  // other text in a way that forced color mode will respect.
+
+  @media screen and (-ms-high-contrast: active) {
+    // IE does not support `CanvasText`,
+    // and `currentColor` does not work with Blink
+    color: currentColor; // stylelint-disable-line value-keyword-case
+  }
+
+  @media screen and (forced-colors: active) {
+    background-color: Canvas;  // needed for Firefox
+    color: CanvasText;
+  }
 }
 
 .related-links__link {


luke-booker (approved):
Looking good. Checked on Windows using Edge and Firefox 👍

PR #4142, opened 2026-08-12
Replace colour variables with govuk-functional-colour function

MartinJJones:
## What
Replace colour variables with govuk-functional-colour function

## Why
The Sass variables for accessing functional colours are deprecated, and will be removed in a future breaking release of govuk-frontend

Jira card: https://gov-uk.atlassian.net/browse/NAV-19017


MartinJJones:
@dependabot rebase

jon-kirwan (approved):

